PR #1, opened 2025-06-27
Add os lib

KyleBonnes:
Basic OS library.
All funcs tested on q 2.7 on homer (Linux)
All funcs tested on q 4.1 locally (Windows)


os/os.q
```diff
@@ -0,0 +1,99 @@
...
+
+// Delay for a specified number of seconds
+.os.sleep:{[num] num:string num; system ("sleep ", num; "timeout /t ", num, " >nul")[.os.isWindows];}
+
+// Change current working directory
+.os.cd:{[path] system "cd ", .os.toPath path;};
+
+// Change the permissions of a file/directory
+.os.chmod:{[path; mode]
+    if[.os.isWindows; 'nyi];
```
jonathonmcmurray:
```suggestion
    if[.os.isWindows; '`nyi];
```

current code will be throwing a subtly different error as q will be trying to find a variable named `nyi` and failing; it will signal `'nyi` but the difference is it will happen inside the `.os.chmod` function rather than where `.os.chmod` is called

(same for other similar cases in other functions)

os/os.q
```diff
@@ -0,0 +1,99 @@
+// Library for utility functions for interacting with the operating system
+
+// Indicator for if the underlying OS is Windows
+.os.isWindows:.z.o in `w32`w64;
```
jonathonmcmurray: let's avoid camelCase please

```suggestion
.os.iswindows:.z.o in `w32`w64;
```

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+.os.del:{[path] system ("rm ";"del ")[.os.isWindows], .os.toPath path;};
+
+// Delete a directory
+.os.delDir:{[path] system ("rm -r ";"rd /s /q ")[.os.isWindows],.os.toPath path;};
+
+// Create directory (without error if existing), make parent directories as needed
+.os.mkdir:{[path] system "mkdir \"", .os.toPath[path], "\"", $[.os.isWindows; " 2>nul"; " -p"];};
+
+// Move/rename a file/directory
+.os.mv:{[src; dest] system ("mv "; "move /y ")[.os.isWindows], " " sv .os.toPath each (src; dest);};
```
jonathonmcmurray:
```suggestion
.os.mv:{[src;dest]system("mv "; "move /y ")[.os.isWindows]," " sv .os.toPath each (src;dest);};
```
IMO we should avoid unnecessary spaces, I don't think they add any clarity or readability here (obviously same applies elsewhere as well)

os/os.q
KyleBonnes: Awwww no...
If I had the choice between importing different OS modules, I'd import the one with camelCase naming for the functions 🤪 

Is there room for debate on how this repo's going to go in terms of coding-style? I'll sacrifice my love of whitespace for camelCase 😆

jonathonmcmurray: I'd import the non-camelCase one :) all code in TorQ currently conforms to the lowercase only convention (with exception of a few global vars that are uppercased) so I think it makes sense to stick with this given a lot of the packages we're building are going to be coming from TorQ code - also eases the transition if people want to move from using TorQ libraries to the new packages if function names etc. remain the same

os/os.q
DHopkinson-DI: made suggested change

os/os.q
DHopkinson-DI: Removed camelCase

os/os.q
DHopkinson-DI: Removed redundant space

os/os.md
```diff
@@ -0,0 +1,168 @@
...
+
+Designed to work on both **Linux/Unix** and **Windows**, functions automatically detect the OS and choose the correct system command.
+
+**All functions that accept path arguments can accept string or hsym paths.**
+
+---
+
+## :desktop_computer: OS Detection
+
+### `.os.isWindows`
```
jonathonmcmurray: need to update docs to remove camelCase as well to match the code

jonathonmcmurray (commented):
Looks good, just need docs updated to match code changes

os/os.q
```diff
@@ -0,0 +1,99 @@
+// Library for utility functions for interacting with the operating system
+
+// Indicator for if the underlying OS is Windows
```
jonnypress: This looks good generally.

Can we add a "dry run" mode for anything that involves a system call? System calls always make me nervous, so for each function that involves one we should have the option to just output a string of what it is going to call, without actually calling it.

os/os.md
```diff
@@ -0,0 +1,168 @@
...
+
+## :link: Symlinks and FIFOs
+
+### `.os.createsymlink[target;name]`
+Creates a symbolic link pointing from `name` to `target`.
+
+- Uses `mklink` on Windows and `ln -s` on Unix.
+- Requires admin rights on Windows for file symlinks (unless Developer Mode is enabled).
+
+### `.os.mkfifo[path]`
```
daviddemnerDI: there's also mktemp which is useful

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+.os.iswindows:.z.o in `w32`w64;
+
+// Return string/symbol/hsym as a string path, using the separators of the OS
+.os.topath:{[path]
+    if[10h<>type path;path:string path];
+    if[.os.iswindows;path:ssr[path;"/";"\\"]];
```
daviddemnerDI: what if you specify a windows path and you're on linux?

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+    :$[":"=first path;1_;] path;
+ };
+
+// Get the absolute path of a file/directory without resolving symlinks
+.os.abspath:{[path]
+    :trim system("realpath -s ",.os.topath[path];"for %F in (\"",.os.topath[path],"\") do @echo %~fF")[.os.iswindows];;
```
daviddemnerDI: extra ; at end. and I know using list indexing will work but maybe $ is clearer (same applies to all of the below)
```suggestion
    :trim system$[.os.iswindows;"for %F in (\"",.os.topath[path],"\") do @echo %~fF";"realpath -s ",.os.topath[path]];
```

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+ };
+
+// Get the absolute path of a file/directory resolving symlinks
+.os.realpath:{[path]
+    if[.os.iswindows;'`nyi];
+    :system"realpath ",.os.topath path;
+ };
+
+// Check if a file/directory exists
+.os.exists:{[path] :not ()~key hsym$[10h=type path;`$;] path;};
```
daviddemnerDI: a bit tidier to use .os.topath maybe? that's sort of the reason for the function
```suggestion
.os.exists:{[path] :not ()~key hsym `$.os.topath[path];
```

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+
+// Check if a path exists and is a file
+.os.isfile:{[path] :.os.exists[path]&{x~key x} hsym$[10h=type path;`$;] path;};
```
daviddemnerDI: same

os/os.q
```diff
@@ -0,0 +1,99 @@
...
+
+// Check if a path exists and is a directory
+.os.isdir:{[path] :.os.exists[path]&not {x~key x} hsym$[10h=type path;`$;] path;};
```
daviddemnerDI: same

os/os.md
jmlemay: Added.

os/os.q
jmlemay: At the moment, this is a nop
```
q).os.iswindows
0b
q).os.topath"C:\\path\\to\\file"
"C:\\path\\to\\file"
```
On the one hand, I can't imagine anyone ever wanting to transform a windows path to a linux one, but I for symmetry, we could support it. Curious to get people's opinions.

os/os.q
jmlemay: I tend to agree. Since I've inherited this, it shall be so.

os/os.q
jmlemay: There's also not really a good way to distinguish between a windows path file 'C:\file', and a linux file called 'C:\file'. Again, why would anyone do this? But it becomes a bit messy to go Windows -> Linux.

jmlemay: Added some code to flip "\" to "/" (at the very least match the documentation). We will not handle a leading drive letter, however. If we want that it's a bit trickier.

os/os.md
```diff
@@ -0,0 +1,174 @@
...
+## :hammer_and_wrench: Process Control
+
+### `.os.kill[pid;sig]`
+Sends a signal to a process (e.g.,`2`,`3`,or `9`).
+
+### `.os.kill2[pid]`
+Interrupts a process (`SIGINT`,equivalent to `kill -2`). Behaves as a normal kill on Windows.
+
+### `.os.kill3[pid]`
+Quits a process (`SIGQUIT`,`kill -3`).
```
jonnypress: What does this do on windows?

os/os.md
```diff
@@ -0,0 +1,174 @@
...
+q).os.topath `:some/relative/path // on Unix...
+"some/relative/path"
+```
+
+### `.os.abspath[path]`
+Returns the **absolute path** to a file/directory **without resolving symlinks** (does not check that the path exists).
+
+```q
+.os.abspath "../file.txt"
+```
```
jonnypress: Can we get the return value examples for each of these?

os/os.q
jmlemay: Added.

os/os.md
jmlemay: On windows, all kill commands are `taskkill /PID {pid}`, except for `kill9`, which is `taskkill /f /PID {pid}`. Worth calling out in the docs.

os/os.md
jmlemay: Done.

jonathonmcmurray (approved):

PR #3, opened 2025-07-10
Add timer lib

jamiechandler99:
Tested on linux versions 3.5 and up 

Resolves #10

### DIFF
--- a/timer/timer.md
+++ b/timer/timer.md
@@ -172,7 +172,7 @@ helloFunc:{show "Hello from job 1"};
 .timer.addjob.simple[`job1; helloFunc; 5]; / Runs every 5 seconds
 
 echoFunc:{show x};
-.timer.addjob.custom[`job2;echoFunc;("Echo this!");10;2;3;0Np;1b];
+.timer.addjob.custom[`job2;echoFunc;enlist "Echo this!";10;2;3;0Np;1b];
 
 timeAligned:{show "On the quarter hour"};
 .timer.addjob.mode[`job3;timeAligned;();15;5];
--- a/timer/timer.md
+++ b/timer/timer.md
@@ -1,0 +1,194 @@
+## KDB+/Q Timer Function Library
+
+A lightweight, customizable timer management library in kdb+/q for scheduling and executing time-based functions. This module supports fixed-interval scheduling, job metadata tracking, conditional disabling, and debug logging—ideal for systems requiring timed task orchestration.
+
+---
+
+### Features
+- Register arbitrary functions with scheduling parameters
+- Support for passing execution arguments to functions
+- Five execution modes for timing alignment
+- Tracks both scheduled and actual run timestamps
+- Auto-disable based on time or run limits
+- Control job activation, status, and failure handling
+- Toggleable logging for development and troubleshooting
+
+---
+
+
+### Core Concepts
+Timer jobs are tracked in ```.timer.procs``` an in-memory table storing job metadata:
+
+| Column        | Type      | Description                                                           |
+| ------------- | --------- | --------------------------------------------------------------------- |
+| id            | symbol    | Unique function identifier                                            |
+| func          | function  | Function to be executed                                               |
+| params        | list      | Parameters passed to the function if required                         |
+| created       | timestamp | Time job was registered                                               |
+| period        | int       | scheduling interval in seconds or minutes depending on mode           |
+| mode          | short     | Scheduling mode for computing run times, see "Execution Mode" section |
+| prevstart     | timestamp | Scheduled time of previous run                                        |
+| actualstart   | timestamp | Actual start time of previous run                                     |
+| prevend       | timestamp | End time of previous run                                              |
+| nextstart     | timestamp | time of next scheduled run                                            |
+| runs          | int       | Count of successful executions                                        |
+| maxruns       | int       | Max number of runs before job is disabled                             |
+| maxtime       | timestamp | Cutoff time for job scheduling                                        |
+| status        | boolean   | Whether job is active (1b) or disabled (0b)                           |
+| disableonfail | boolean   | Flag to disable job on failure if enabled                             |  
+
+---
+
+### Execution Mode
+Timer jobs use one of four scheduling modes, defined by the mode column:
+
+| Mode     | Scheduling Logic                                             | Example                       |
+| -------- | ------------------------------------------------------------ | ----------------------------- |
+| 1        | Jobs scheduled x seconds after previous scheduled start time |                               |
+| 2        | Jobs scheduled x seconds after previous actual start time    |                               |
+| 3        | Jobs scheduled x seconds after previous end time             |                               |
+| 4        | Jobs scheduled for the xth minute of every hour              | x = 10 : 13:15,14:15,15:15... |
+| 5        | Jobs scheduled on the hour in x minute intervals             | x = 15 : 13:00,13:15,13:30... |
+
+---
+
+### Usage .timer.addjob
+Functions allow you to register new timer-based jobs with varying degrees of customization—ranging from full control over scheduling, parameters, and failure handling to simplified shortcuts for common use cases.
+
+| Parameter     | Type      | Description                                                           |
+| ------------- | --------- | --------------------------------------------------------------------- |
+| id            | symbol    | Unique function identifier, must not already exist in .timer.procs    |
+| func          | function  | Function to be executed when job runs                                 |
+| params        | list      | arguments to pass into the function at run time                       |
+| period        | int       | Time period in seconds or minutes for next run time calculations      |
+| mode          | short     | Scheulding more to determine next run time                            |
+| maxruns       | int       | Optional limit on nnumber of times the job may run. 0Ni to disables   |
+| maxtime       | timestamp | Optional timestamp after which the job becomes inactive, 0Np disables |
+| disableonfail | boolean   | If set to 1b, disables future runs if the function fails to execute   |
+
+##### .timer.addjob.custom
+Lets you fully define a scheduled job by specifying its function, parameters, timing, mode, and control flags for execution limits and failure handling.
+```
+.timer.addjob.custom[`job1;{show x};("Hello!");10;1;5;0Np;1b]
+```
+
+##### .timer.addjob.simple 
+Provides a fast way to schedule a job using sensible defaults—ideal for quick setup when you only need to specify a function, its interval, and minimal configuration. Schedule mode is defaulted to 1.
+```
+.timer.addjob.simple[`job2;{show "Hi"};30]
+```
+
+##### .timer.addjob.mode 
+Streamlined way to create a simple job with custom mode and parameters while applying default safeguards for failure handling and execution limits.
+```
+.timer.addjob.mode[`job3;{show "Mode 2"};();15;2]
+```
+
+---
+
+### Control & Monitoring
+The timer system includes a suite of control functions that let you start, stop, and inspect timer jobs with precision. These utility functions ensure operational flexibility while maintaining the integrity of the scheduling loop.
+
+##### .timer.init
+```
+.timer.init[]
+```
+- Purpose: Initializes the timer scheduler by hooking .z.ts to .timer.main[].
+- Effect: Establishes a recurring loop that checks for jobs due to run based on .timer.cycletime. If .z.ts is already defined, it safely preserves and wraps it.
+- Usage: Call once when starting your timer system or restarting after .timer.disable[].
+
+##### .timer.disable
+```
+.timer.disable[]
+```
+- Purpose: Stops the timer execution loop by restoring the original .z.ts handler.
+- Effect: Prevents .timer.main[] from being automatically triggered, effectively pausing scheduled job execution.
+- Safety: Checks if .z.ts was previously saved before attempting to revert it.
+- Usage: Useful during maintenance, debugging, or when manually controlling job flow
+
+##### .timer.enablejobs
+```
+.timer.enablejobs[`job1]
+.timer.enablejobs[`job1`job2]
+```
+- Purpose: Reactivates one or more jobs by setting status:1b in .timer.procs.
+- Usage: Accepts a single symbol or list of job IDs. Jobs will resume scheduling and execution as normal.
+
+##### .timer.disablejobs
+```
+.timer.disablejobs[`job3]
+.timer.disablejobs[`job4`job5]
+```
+- Purpose: Deactivates one or more jobs by setting status:0b, preventing them from being picked up by the scheduler.
+- Usage: Ideal for temporarily suspending jobs without deleting their metadata.
+
+##### .timer.deletejobs
+```
+.timer.deletejobs[`job3]
+.timer.deletejobs[`job4`job5]
+```
+- Purpose: Removes jobs from .timer.procs table. Can be done without disabling loop.
+- Usage: Ideal for re-defining job params wihtout interrupting other jobs.
+
+
+##### .timer.getactive
+```
+.timer.getactive[]
+```
+- Purpose: Queries .timer.procs and returns all currently active jobs (status=1b).
+- Output: Returns a table of job rows, allowing inspection of current scheduling, metadata, and run statistics.
+- Usage: Useful for dashboarding, monitoring job health, or debugging live workflows
+
+---
+
+### User Customisations
+These are global control variables and can be adjusted by the user after import before initializing with .timer.init.
+#### Debug
+```
+.timer.debug:1b
+```
+- Purpose: Enables verbose logging for job execution.
+- Type: Boolean (1b for enabled, 0b for disabled).
+- Usage: When set to 1b, job execution attempts, successes, and errors will be logged to the console with timestamps and context via .timer.msg.info and .timer.msg.err.
+- Default: 0b (disabled).
+This is useful during development or troubleshooting to inspect scheduler behavior.
+#### Cycle Time
+```
+.timer.cycletime:1000
+```
+- Purpose: Sets the interval (in milliseconds) between scheduler checks for jobs that are due to run.
+- Type: Integer.
+- Usage: Controls how frequently .z.ts triggers the main execution loop .timer.main[].
+- Default: 1000 (1 second).
+Adjust this value to increase responsiveness or reduce CPU usage, depending on the expected job timing precision.
+
+### Example 
+```
+\l timer.q / - package import process may change, just for example
+
+.timer.cycletime:500;
+
+helloFunc:{show "Hello from job 1"};
+.timer.addjob.simple[`job1; helloFunc; 5]; / Runs every 5 seconds
+
+echoFunc:{show x};
+.timer.addjob.custom[`job2;echoFunc;("Echo this!");10;2;3;0Np;1b];
+
+timeAligned:{show "On the quarter hour"};
+.timer.addjob.mode[`job3;timeAligned;();15;5];
+
+.timer.init[];
+
+/ Enable and disable jobs manually if needed
+.timer.disablejobs[`job3]
+.timer.enablejobs[`job3]
+
+/ View all active jobs
+.timer.getactive[]
+
+/ Temporarily halt job execution
+.timer.disable[]
+
+/ Restart scheduler loop again later
+.timer.init[]
+```
--- a/timer/timer.md
+++ b/timer/timer.md
@@ -1,6 +1,6 @@
 ## KDB+/Q Timer Function Library
 
-A lightweight, customizable timer management library in kdb+/q for scheduling and executing time-based functions. This module supports fixed-interval scheduling, job metadata tracking, conditional disabling, and debug logging—ideal for systems requiring timed task orchestration.
+A lightweight, customizable timer management library in kdb+/q for scheduling and executing time-based functions. This module supports fixed-interval scheduling, job metadata tracking, conditional disabling, and debug logging — ideal for systems requiring timed task orchestration.
 
 ---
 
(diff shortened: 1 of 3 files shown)


timer/timer.q
```diff
@@ -0,0 +1,115 @@
...
+    created:`timestamp$();       / - timestamp of function initiation
+    period:`int$();              / - time period to base timer scheduling on
+    mode:`short$();              / - mode to base scheduling logic on 
+    prevstart:`timestamp$();     / - scheduled start timestamp of previous run
+    actualstart:`timestamp$();   / - start timestamp of previous run
+    prevend:`timestamp$();       / - end timestamp of previous run 
+    nextstart:`timestamp$();     / - timestamp to run next time 
+    runs:`int$();                / - running count of number of runs 
+    maxruns:`int$();          / - optional flag to disable time function after number of runs
+    maxtime:`timestamp$();    / - optional flag to disable job after timestamp
```
jonathonmcmurray: comment alignment

timer/timer.q
```diff
@@ -0,0 +1,115 @@
+// Library for creating and managing stored timer functions
+
+/ Override variables to change logic
```
jonathonmcmurray: let's stick with a single comment style throughout, we'll align with Kyle's PRs i.e. `//`

```suggestion
// Override variables to change logic
```

(obviously this also applies to other comments throughout)

timer/timer.q
```diff
@@ -0,0 +1,115 @@
...
+    mode:`short$();              / - mode to base scheduling logic on 
+    prevstart:`timestamp$();     / - scheduled start timestamp of previous run
+    actualstart:`timestamp$();   / - start timestamp of previous run
+    prevend:`timestamp$();       / - end timestamp of previous run 
+    nextstart:`timestamp$();     / - timestamp to run next time 
+    runs:`int$();                / - running count of number of runs 
+    maxruns:`int$();          / - optional flag to disable time function after number of runs
+    maxtime:`timestamp$();    / - optional flag to disable job after timestamp
+    status:`boolean$();          / - when true job is enabled, false job disabled
+    disableonfail:`boolean$());  / - if job fails will disable future runs
```
jonathonmcmurray:
```suggestion
    disableonfail:`boolean$()  / - if job fails will disable future runs
    );
```

Closing paren on new line

timer/timer.q
```diff
@@ -0,0 +1,115 @@
+// Library for creating and managing stored timer functions
+
+/ Override variables to change logic
+.timer.debug:0b;
+.timer.cycletime:1000;
+
+/ In memory table maintaining time functions and scheduling of execution 
+.timer.procs:([id:`$()]          / - symbol reference to unique timer functions
```
jonathonmcmurray:
```suggestion
.timer.procs:(
    [id:`$()]          / - symbol reference to unique timer functions
```

also, I think `.timer.jobs` would make more sense as the name of this table, given all the functions refer to jobs (addjob, disablejob etc.)

timer/timer.q
```diff
@@ -0,0 +1,115 @@
...
+    `.timer.procs insert (id;func;params;.z.P;`int$period;`short$mode;0Np;0Np;0Np;0Np;0i;maxruns;maxtime;1b;disableonfail);
+    .timer.upd.start[id];
+    };
+.timer.addjob.simple:.timer.addjob.custom[;;();;1;0Ni;0Np;1b];
+.timer.addjob.mode:.timer.addjob.custom[;;;;;0Ni;0Np;1b];
+
+/ Internal utility functions
+.timer.msg.custom:{[code;msg]0N!(string .z.P)," ### ",code," ### ",msg};
+.timer.msg.info:.timer.msg.custom["INFO";];
+.timer.msg.err:.timer.msg.custom["ERROR";];
```
jonathonmcmurray: We'll need to rethink this, we'll want to use existing logging code from another package if possible but I think we probably also don't want to force people to use our logging package. This is probably OK for now but flagging that we'll need to revisit this

timer/timer.q
```diff
@@ -0,0 +1,115 @@
...
+
+/ Cycle functions
+.timer.main:{
+    torun:exec id from .timer.procs where status,nextstart<.z.P;
+    if[count torun;.timer.runandschedule each torun];
+    };
+
+.timer.init:{
+    .timer.dotz.orig:@[get;`.z.ts;{}];
+    .z.ts:$[null .timer.dotz.orig;.timer.main;{.timer.dotz.orig`;.timer.main`}];
```
jonathonmcmurray: with the error trapped `get` above, I don't think this can ever be null

```q
q)null {}
0b
```

Possibly we should do something like this instead?

```suggestion
    .z.ts:$[{}~.timer.dotz.orig;.timer.main;{.timer.dotz.orig`;.timer.main`}];

```

timer/timer.q
```diff
@@ -0,0 +1,115 @@
+// Library for creating and managing stored timer functions
+
+/ Override variables to change logic
+.timer.debug:0b;
+.timer.cycletime:1000;
```
jonathonmcmurray: need some comments here to explain what these are/mean (I know it's in docs, but not everyone reads docs)

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+| runs          | int       | Count of successful executions                                        |
+| maxruns       | int       | Max number of runs before job is disabled                             |
+| maxtime       | timestamp | Cutoff time for job scheduling                                        |
+| status        | boolean   | Whether job is active (1b) or disabled (0b)                           |
+| disableonfail | boolean   | Flag to disable job on failure if enabled                             |  
+
+---
+
+### Execution Mode
+Timer jobs use one of four scheduling modes, defined by the mode column:
```
jonathonmcmurray: there are 5 modes listed below

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+
+| Mode     | Scheduling Logic                                             | Example                       |
+| -------- | ------------------------------------------------------------ | ----------------------------- |
+| 1        | Jobs scheduled x seconds after previous scheduled start time |                               |
+| 2        | Jobs scheduled x seconds after previous actual start time    |                               |
+| 3        | Jobs scheduled x seconds after previous end time             |                               |
+| 4        | Jobs scheduled for the xth minute of every hour              | x = 10 : 13:15,14:15,15:15... |
```
jonathonmcmurray: in the example, if x=10 surely it should be 13:10,14:10,15:10... ? Unless I'm misunderstanding something

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+| 5        | Jobs scheduled on the hour in x minute intervals             | x = 15 : 13:00,13:15,13:30... |
```
jonathonmcmurray: "on the hour in x minute intervals" seems confusing to me, maybe "on the hour and in x minute intervals"?

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+---
+
+
+### Core Concepts
+Timer jobs are tracked in ```.timer.procs``` an in-memory table storing job metadata:
+
+| Column        | Type      | Description                                                           |
+| ------------- | --------- | --------------------------------------------------------------------- |
+| id            | symbol    | Unique function identifier                                            |
+| func          | function  | Function to be executed                                               |
```
jonathonmcmurray: can func also be a symbol reference to a function as well as an actual function? I think that's how it works in torq, not sure if it's changed here

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+| 1        | Jobs scheduled x seconds after previous scheduled start time |                               |
+| 2        | Jobs scheduled x seconds after previous actual start time    |                               |
+| 3        | Jobs scheduled x seconds after previous end time             |                               |
+| 4        | Jobs scheduled for the xth minute of every hour              | x = 10 : 13:15,14:15,15:15... |
+| 5        | Jobs scheduled on the hour in x minute intervals             | x = 15 : 13:00,13:15,13:30... |
+
+---
+
+### Usage .timer.addjob
+Functions allow you to register new timer-based jobs with varying degrees of customization—ranging from full control over scheduling, parameters, and failure handling to simplified shortcuts for common use cases.
```
jonathonmcmurray:
```suggestion
Functions allow you to register new timer-based jobs with varying degrees of customization — ranging from full control over scheduling, parameters, and failure handling to simplified shortcuts for common use cases.

```

timer/timer.md
```diff
@@ -0,0 +1,194 @@
+## KDB+/Q Timer Function Library
+
+A lightweight, customizable timer management library in kdb+/q for scheduling and executing time-based functions. This module supports fixed-interval scheduling, job metadata tracking, conditional disabling, and debug logging—ideal for systems requiring timed task orchestration.
```
jonathonmcmurray:
```suggestion
A lightweight, customizable timer management library in kdb+/q for scheduling and executing time-based functions. This module supports fixed-interval scheduling, job metadata tracking, conditional disabling, and debug logging — ideal for systems requiring timed task orchestration.

```

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+| -------- | ------------------------------------------------------------ | ----------------------------- |
+| 1        | Jobs scheduled x seconds after previous scheduled start time |                               |
+| 2        | Jobs scheduled x seconds after previous actual start time    |                               |
+| 3        | Jobs scheduled x seconds after previous end time             |                               |
+| 4        | Jobs scheduled for the xth minute of every hour              | x = 10 : 13:15,14:15,15:15... |
+| 5        | Jobs scheduled on the hour in x minute intervals             | x = 15 : 13:00,13:15,13:30... |
+
+---
+
+### Usage .timer.addjob
```
jonathonmcmurray:
```suggestion
### Usage 
##### .timer.addjob
```

Not 100% sure on this - this section is a bit confusing tbh, it's not immediately clear to me that `.timer.addjob` is not something I can call directly

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+| disableonfail | boolean   | If set to 1b, disables future runs if the function fails to execute   |
+
+##### .timer.addjob.custom
+Lets you fully define a scheduled job by specifying its function, parameters, timing, mode, and control flags for execution limits and failure handling.
+```
+.timer.addjob.custom[`job1;{show x};("Hello!");10;1;5;0Np;1b]
+```
+
+##### .timer.addjob.simple 
+Provides a fast way to schedule a job using sensible defaults—ideal for quick setup when you only need to specify a function, its interval, and minimal configuration. Schedule mode is defaulted to 1.
```
jonathonmcmurray:
```suggestion
Provides a fast way to schedule a job using sensible defaults — ideal for quick setup when you only need to specify a function, its interval, and minimal configuration. Schedule mode is defaulted to 1.

```

timer/timer.md
```diff
@@ -0,0 +1,194 @@
...
+.timer.cycletime:1000
+```
+- Purpose: Sets the interval (in milliseconds) between scheduler checks for jobs that are due to run.
+- Type: Integer.
+- Usage: Controls how frequently .z.ts triggers the main execution loop .timer.main[].
+- Default: 1000 (1 second).
+Adjust this value to increase responsiveness or reduce CPU usage, depending on the expected job timing precision.
+
+### Example 
+```
```
jonathonmcmurray:
```suggestion
```q
```
may not be supported everywhere, but this should enable q syntax highlighting for the example where supported

timer/timer.q
```diff
@@ -0,0 +1,115 @@
...
+    maxruns:`int$();          / - optional flag to disable time function after number of runs
+    maxtime:`timestamp$();    / - optional flag to disable job after timestamp
+    status:`boolean$();          / - when true job is enabled, false job disabled
+    disableonfail:`boolean$());  / - if job fails will disable future runs
+
+/ Job functions for adding new functions to be scheduled and executed
+.timer.addjob.custom:{[id;func;params;period;mode;maxruns;maxtime;disableonfail]
+    if[id in key .timer.procs;
+        .timer.msg.err"Cannot add id : ",(string id)," already exists in .timer.procs";
+        :()];
```
jonathonmcmurray: Might be best to signal an error here instead of logging, and allow the caller to catch it & handle as needed

```suggestion
        '"Cannot add id : ",(string id)," already exists in .timer.procs";
    ];
```

timer/timer.q
jamiechandler99: Yes agreed. For now this just make debugging messages a little more clear so worth keeping for now imo. But as we add logging and a more unified structure across all packages definitely worth revisting

timer/timer.md
jamiechandler99: Updated to include support for symbol reference of a defined function

timer/timer.md
jamiechandler99: Reformatted to add clarity

timer/timer.q
jonathonmcmurray: this is marked as resolved but I don't see any change/response?

jonathonmcmurray: think this just needs addressed then I'll be happy with this one

jamiechandler99: Thanks for catching absolutely right. Could have sworn I updated this already, but should be resolved now.

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+    if[id in key .timer.jobs;'"Cannot add id : ",(string id)," already exists in .timer.jobs"];
+    if[-11h=type func;func:get func];
+    `.timer.jobs insert (id;func;params;.z.P;`int$period;`short$mode;0Np;0Np;0Np;0Np;0i;maxruns;maxtime;1b;disableonfail);
+    .timer.upd.start[id];
+    };
+.timer.addjob.simple:.timer.addjob.custom[;;();;1;0Ni;0Np;1b];
+.timer.addjob.mode:.timer.addjob.custom[;;;;;0Ni;0Np;1b];
+
+// Internal utility functions
+.timer.msg.custom:{[code;msg]0N!(string .z.P)," ### ",code," ### ",msg};
```
jonnypress: I don't think we want 0N! here for printing. If we want to print messages we need a logging function, but not clear how this will all work w.r.t. managing dependencies with e.g. a logging library, so leave as is for now.

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+.timer.disable:{
+    if[0=system"t";.timer.msg.info".z.ts not enabled... doing nothing";:()];
+    $[{}~.timer.dotz.orig;system"t 0";.z.ts:.timer.dotz.orig];
+    };
+
+.timer.enablejobs:{[ids] ids:(),ids;.timer.jobs:update status:1b from .timer.jobs where id in ids};
+.timer.disablejobs:{[ids] ids:(),ids;.timer.jobs:update status:0b from .timer.jobs where id in ids};
+.timer.deletejobs:{[ids] ids:(),ids;.timer.jobs:delete from .timer.jobs where id in ids};
+
+.timer.getactive:{[id] select from .timer.jobs where status};
```
jonnypress: What is the param used for here?

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+        if[r`disableonfail;r[`status]:0b];
+        ];
+    r:.timer.upd.status[r];
+    .timer.jobs[id]:r;
+    if[r[`status];.timer.upd.start[id]];
+    };
+
+// Cycle functions
+.timer.main:{
+    torun:exec id from .timer.jobs where status,nextstart<.z.P;
```
jonnypress: I think all the start/ends should be done in UTC because otherwise we'll get odd behaviour during daylight savings changes

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+    if[count torun;.timer.runandschedule each torun];
+    };
+
+.timer.init:{
+    .timer.dotz.orig:@[get;`.z.ts;{}];
+    .z.ts:$[{}~.timer.dotz.orig;.timer.main;{.timer.dotz.orig`;.timer.main`}];
```
jonnypress: Need to be consistent here with how we stack .z. definitions between this and the other packages. 
Also the default parameter to .z.ts is the current time in UTC (.z.p), here you are passing in null symbol.

timer/timer.q
jamiechandler99: Left over from a legacy version. I've removed.

timer/timer.q
jamiechandler99: Switched scheduling logic for start/end times to use UTC

timer/timer.q
jamiechandler99: Switched function call to use defaults. In terms of consistency across packages, not sure of the best way to tackle this. My idea at the moment is a top level package that is utilities to be used across packages to ensure they are aligned. Something that can implement logging consistently across all packages and contain utility functions like a function to handle updating .z.ts to include any existing definition. This package could be applied within the import framework by default but not mandatory that way individual users arent forced to use it, and will be where we put any logic that would be duplicated across packages. Pros : ensures consistency across the whole project, Cons: starts to get away from the standalone package objective and add more dependencies

timer/timer.q
jamiechandler99: ack. Mentioned in other comment how I think we should handle in the future

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+// Evalution functions for executing jobs and scheduling
+.timer.runandschedule:{[id]
+    r:.timer.jobs[id];
+    if[.timer.debug;.timer.msg.info"Executing job id: ",(string id)," for start time: ",(string r`nextstart)];
+    r[`actualstart]:.z.p;
+    ret:$[count r`params;
+        .[r`func;(),r`params;{.timer.msg.err"Job execution failed with error : ",x;0b}];
+        @[r`func;`;{.timer.msg.err"Job execution failed with error : ",x;0b}]
+        ];
+    if[not -1h=type ret;ret:1b];
```
jonnypress: Logic here needs improved. What if the function being called returns a boolean value? 

I think this whole block 

ret:$[count r`params;
        .[r`func;(),r`params;{.timer.msg.err"Job execution failed with error : ",x;0b}];
        @[r`func;`;{.timer.msg.err"Job execution failed with error : ",x;0b}]
        ];
if[not -1h=type ret;ret:1b];

could become something like 

ret:.[{$[count y; x . (),y; x@`];1b};r`func`params; {.timer.msg.err"Job execution failed with error : ",x;0b}];

Need to test the edge cases on that, e.g. is a function that takes a single argument handled correctly?

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+    if[not ret;
+        if[.timer.debug;.timer.msg.err"Job execution failed for job id : ",(string id)];
+        if[r`disableonfail;r[`status]:0b];
+        ];
+    r:.timer.upd.status[r];
+    .timer.jobs[id]:r;
+    if[r[`status];.timer.upd.start[id]];
```
jonnypress: Similarly how TorQ does it, it would be nice to have a configurable function to allow re-definition of how a timer function is executed, e.g. .timer.execute

Example (and default implementation) would be to pass the timer funcation call through the 0 handle. This means that it will show up in usage logs. Other example would be to modify it to log out each timer call.

timer/timer.md
```diff
@@ -0,0 +1,196 @@
...
+| period        | int       | Time period in seconds or minutes for next run time calculations      |
+| mode          | short     | Scheulding more to determine next run time                            |
+| maxruns       | int       | Optional limit on nnumber of times the job may run. 0Ni to disables   |
+| maxtime       | timestamp | Optional timestamp after which the job becomes inactive, 0Np disables |
+| disableonfail | boolean   | If set to 1b, disables future runs if the function fails to execute   |
+
+#### .timer.addjob.custom
+Lets you fully define a scheduled job by specifying its function, parameters, timing, mode, and control flags for execution limits and failure handling.
+```q
+.timer.addjob.custom[`job1;{show x};("Hello!");10;1;5;0Np;1b]
```
jonnypress: I should be able to schedule a job with a specific start time - doesn't appear I can do that currently?

timer/timer.q
jonnypress: When I say consistency across packages, I mean in terms of the code we use to "stack" definitions. 

IMO, the best way to do it is to pass in the old definition to the new function, and invoke it, so you don't have to create e.g. .timer.dotz.orig which has no useful future value. 

example: 

$[@[{value x;1b};`.z.ts;0b];     // test if one already exits
 .z.ts:{[x;y] mynewfunc y; x@y}[.z.ts];
 .z.ts:{mynewfunc x}]

We need to be consistent with this approach across packages.

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+
+// Cycle functions
+.timer.main:{
+    torun:exec id from .timer.jobs where status,nextstart<.z.p;
+    if[count torun;.timer.runandschedule each torun];
+    };
+
+.timer.init:{
+    .timer.dotz.orig:@[get;`.z.ts;{}];
+    .z.ts:$[{}~.timer.dotz.orig;.timer.main;{.timer.dotz.orig .z.p;.timer.main .z.p}];
```
jonnypress: what happens if this init function is called more than once? Does it cope? If it doesn't cope, then we need protection against that to avoid it happening. I think this is probably an idempotent function, i.e. if we call it multiple times only the first time does anything.

timer/timer.q
```diff
@@ -0,0 +1,116 @@
...
+        if[r`disableonfail;r[`status]:0b];
+        ];
+    r:.timer.upd.status[r];
+    .timer.jobs[id]:r;
+    if[r[`status];.timer.upd.start[id]];
+    };
+
+// Cycle functions
+.timer.main:{
+    torun:exec id from .timer.jobs where status,nextstart<.z.p;
```
jonnypress: This could get quite expensive if there are a lot of functions configured, and the timer has a short period. Have a look at the TorQ code - it trackes the next time that it has to do anything and then the next timer check is just 

if[timenow < next time we have to do something]

Similarly, TorQ uses the a function to retrive the current time, rather than .z.p directly. This is useful in backtesting use cases or anything where we want to utilise a simulated timeline.

timer/timer.q
jamiechandler99: Updated .z.ts to align with above

timer/timer.q
jamiechandler99: Have simplified logic above given recommendation. Have also adjusted to implement .timer.execute with additional function execution through 0 handle for usage log tracking

timer/timer.md
jamiechandler99: Have adjusted all addjob functions to take a dictionary of configurable options and added startattime as one of them to allow the user to customize the initial job start time

timer/timer.q
jamiechandler99: Added protection around .timer.enabled status to make sure this doesnt cause issues

timer/timer.q
jamiechandler99: Updated to be similar to TorQ where unless the next minimum start time is reached, this will not waste resources checking against the entire in memory table

jamiechandler99:
Added logic to address comments, aligned with most recent version of style guide and added unit tests for package aligned with our testing framework

jonnypress (commented):
Looks good to me as first version

jonnypress (approved):
Sorry meant to hit approve in the last comment

PR #4, opened 2025-07-15
Add datadog lib

DHopkinson-DI:
Based on TorQ's datadog.q script
All funcs tested on q 4.1 on homer (Linux)
All funcs tested on q 4.1 locally (Windows 11)


datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
...
+    response
+ }
+
+// Sends events via https post to datadog api
+.dg.sendevent_webreq:{[event_title;event_text;priority;tags;alert_type]
+    url:.dg.dogstatsd_url,"events?api_key=",.dg.dogstatsd_apikey;
+    json:.j.j `title`text`priority`tags`alert_type!(event_title;event_text;priority;$[0h=type tags;","sv tags;tags];alert_type);
+    $[.z.o like "l*";
+      response:.[.Q.hp;(url;.h.ty`json;json);{'"error with https request: ",x}];
+      [response:.[.dg.nonlinux_webreq;(url;json;"event");{'"error with https request: ",x}];hdel `:event.json]];
```
jonathonmcmurray: why are we using `.Q.hp` on linux, but not elsewhere? `.Q.hp` should work on any OS I think?

Also if we're keeping `.dg.nonlinux_webreq` I think we should move the `hdel` inside it rather than having to delete it separately

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
+// -----datadog.q -----
+// Packaged used to push metrics and events from a q process to datadog.
+// default delivery method is through the datadog agent installed on the host. use ".dg.webreq 1b" to switch delivery to https
+
+
+
+// Following two functions used to push data to datadog agent on linux os
+
+//send event on linux os using datadog agent
```
jonathonmcmurray: keep comments consistent please (i.e. start with a space after the `//` & then uppercase letter)

```suggestion
// Following two functions used to push data to datadog agent on linux os

// Send event on linux os using datadog agent
```

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
+// -----datadog.q -----
+// Packaged used to push metrics and events from a q process to datadog.
+// default delivery method is through the datadog agent installed on the host. use ".dg.webreq 1b" to switch delivery to https
+
+
+
+// Following two functions used to push data to datadog agent on linux os
+
+//send event on linux os using datadog agent
+.dg.sendevent_l:{[event_title;event_text;priority; tags;alert_type]
```
jonathonmcmurray: no underscores in function/variable names please

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
...
+    $[.z.o like "w*";[.dg.sendevent:.dg.sendevent_ps;.dg.sendmetric:.dg.sendmetric_ps];
+      $[.z.o like "l*";[.dg.sendevent:.dg.sendevent_l;.dg.sendmetric:.dg.sendmetric_l];
+       '"Currently only linux and windows operating systems are supported to send metrics and events. Please use "".dg.setwebreq 1b"" to attempt a web request"]];
+ }
+
+// Override function to determine if web request is used or datadog agent
+.dg.setwebreq:{[bool]
+    $[bool;[.dg.sendevent:.dg.sendevent_webreq; .dg.sendmetric:.dg.sendmetric_webreq;.dg.webrequest::1b];
+      [.dg.setosfunctions[];.dg.webrequest::0b]];
+ }
```
jonathonmcmurray: this is quite messy IMO. I think `bool` is very undescriptive, and the block syntax inside a `$` conditional makes things harder to read

Maybe something like

```suggestion
.dg.setwebreq:{[useweb]
    // set functions to web request versions
    if[useweb;
      .dg.sendevent:.dg.sendevent_webreq;
      .dg.sendmetric:.dg.sendmetric_webreq;
      .dg.webrequest::1b;
      :();
    ];
    // otherwise, use OS based functions
    .dg.setosfunctions[];
    .dg.webrequest::0b;
 }
```

Arguably I'd integrate `.dg.setosfunctions` here as well instead of having two separate functions

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
...
+
+ };
+
+// Utility functions used to manage send functions
+
+// sets sendmetric and sendevents ddagent functions depending on os
+.dg.setosfunctions:{
+    $[.z.o like "w*";[.dg.sendevent:.dg.sendevent_ps;.dg.sendmetric:.dg.sendmetric_ps];
+      $[.z.o like "l*";[.dg.sendevent:.dg.sendevent_l;.dg.sendmetric:.dg.sendmetric_l];
+       '"Currently only linux and windows operating systems are supported to send metrics and events. Please use "".dg.setwebreq 1b"" to attempt a web request"]];
```
jonathonmcmurray: Don't technically need the nested `$` as you can have multiple conditions in a single `$`. However as mentioned elsewhere I think the block syntax inside a conditional is messy so maybe

It might make sense to setup some dictionaries & use those instead e.g.

```q
.dg.win:`sendevent`sendmetric!(.dg.sendevent_ps;.dg.sendmetric_ps)
.dg.lin:`sendevent`sendmetric!(.dg.sendevent_l;.dg.sendmetric_l)
.dg.web:`sendevent`sendmetric!(.dg.sendevent_webreq;.dg.sendmetric_webreq)
```

and then when you go to set the functions you can just do e.g.

```q
.dg,:.dg[x]
```

where `x` is one of `win`, `lin` or `web`.

You could also directly define the functions in the dictionaries like

```q
.dg.win.sendevent:{...}
.dg.win.sendmetric:{...}
```

etc. (i.e. instead of defining these as .dg.sendevent_ps etc.)

datadog/datadog.q
jonathonmcmurray: Also, I don't think the `.dg.webrequest` variable is used anywhere, so probably not needed?

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
...
+    .dg.dogstatsd_apikey:@[value;`.dg.dogstatsd_apikey;getenv[`DOGSTATSD_APIKEY]];                                                            // define dogstatsd_apikey
+    .dg.dogstatsd_url:@[value;`.dg.dogstatsd_url;":https://api.datadoghq.eu/api/v1/"];                                                        // define dogstatsd_url
+    .dg.dogstatsd_metriclog:([] time:`timestamp$();host:`$();message:();metric_name:();metric_value:`float$();https:`boolean$();status:());   // define table to capture requests
+    .dg.dogstatsd_eventlog:([] time:`timestamp$();host:`$();message:();event_title:();event_text:();https:`boolean$();status:());             // define table to capture requests
+    .dg.webrequest:@[value;`.dg.webrequest;0b];                                                                                               // default to disabled - datadog agent used
+    .dg.setwebreq[.dg.webrequest]                                                                                                             // sets delivery method
```
jonathonmcmurray: probably makes sense to pass this as a param to `.dg.init` instead of having to set it as a config variable before calling `.dg.init`?

datadog/datadog.q
```diff
@@ -0,0 +1,114 @@
...
+    cmd: raze "bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#",$[0h=type tags;","sv tags;tags],"' > /dev/udp/127.0.0.1/",string .dg.dogstatsd_port,"\"";
+    response:system cmd;
+    .dg.dogstatsd_metriclog,:(.z.p;.z.o;cmd;metric_name;`float$metric_value;0b;response);
+ };
+
+// Following three functions are used to push metrics and events to datadog through udp and powershell on windows os
+
+// shell command to push data
+.dg.pushtodogagent:{[message]
+    cmd: "powershell -Command \"";
```
jonathonmcmurray: is powershell guaranteed to be available on any Windows install? e.g. on Windows Server installs etc.? (not familiar enough to know if it's universally available/permissioned etc.)

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+
+- Send custom metrics and events to datadog platform.
+- Allows posts to be pushed via datadog agent or https
+- Log all posts and delivery status to in memory tables.
+
+---
+
+## :gear: Configuration
+
+Config variables used to connect to datadog and change the mode of delivery can be set **before loading** the script:
```
jonathonmcmurray: I don't think these should have to be set before loading the script, but rather before calling the init function?

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+.dg.dogstatsd_port   : getenv[`DOGSTATSD_PORT]                                          // Values the datadog agent port or sets to default
+.dg.dogstatsd_apikey : getenv[`DOGSTATSD_APIKEY]                                        // Sets datadog api key
+.dg.dogstatsd_url    : ":https://api.datadoghq.eu/api/v1/"                              // defines datadog api endpoint url
+.dg.dogstatsd_metriclog:([] time:`timestamp$();host:`$();message:();                    // define table to log metrics and delivery status
+    metric_name:();metric_value:`float$();https:`boolean$();status:());              
+.dg.dogstatsd_eventlog:([] time:`timestamp$();host:`$();message:();                     // define table to log events and delivery status
+    event_title:();event_text:();https:`boolean$();status:());                
+.dg.webrequest:@[value;`.dg.webrequest;0b];                                             // set webrequest var to (default: 0b) 
+.dg.setwebreq[.dg.webrequest]                                                           // override with https functions
+```
```
jonathonmcmurray: I think some or all of this is in the init function? Shouldn't need to be done explicitly so I don't think it belongs in docs, seems confusing

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+
+
+### :rocket: Push to Datadog Agent Functions
+
+Primary functions used to push data to datadog. These are the only functions required to send data as they are overridden depending on os/http.
+
+| Function         | Params                                                                                         | Description                                    |
+|------------------|------------------------------------------------------------------------------------------------|------------------------------------------------|
+| `.dg.sendmetric` | (metric_name: string; metric_value: float; tags:string)                                        | Default metric delivery function (default: {}) |
+| `.dg.sendevent`  | (event_title: string; event_text: string; priority: string; tags: string; alert_type: string ) | Default event delivery function (default: {})  |
```
jonathonmcmurray: I think it's confusing to say the default is `{}` - I get what you mean, but this will be overwritten during initialisation, so I don't think it's really meaningful and just serves to create confusion

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+
+
+### :rocket: Overriding Functions Pushing https
+
+Functions used to send https posts to datadog api endpoint.
+
+| Function                | Params                                                                                        | Description                                                             |
+|-------------------------|-----------------------------------------------------------------------------------------------|-------------------------------------------------------------------------|
+| `.dg.sendmetric_webreq` | (metric_name: string; metric_value: float; tags:string)                                       | Pushes metric data to datadog api endpoint via https if webrequest = 1b |
+| `.dg.sendevent_webreq`  | (event_title: string; event_text: string; priority: string; tags:string; alert_type: string ) | Pushes event data to datadog api endpoint via https if webrequest = 1b  |
```
jonathonmcmurray: Arguably we don't need to document these functions - no one should ever be calling these, they should only be calling `.dg.sendevent` and `.dg.sendmetric`, so those are the only ones we should document IMO

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+
+### :hammer_and_wrench: Utilities
+
+Functions used to determine appropriate send functions and set flags.
+
+| Function             | Params          | Description                                           |
+|----------------------|-----------------|-------------------------------------------------------|
+| `.dg.setosfunctions` | (::)            | Determines appropriate override by os                 |
+| `.dg.setwebreq`      | (bool: boolean) | Overrides functions to use https rather datadog agent |
+
```
jonathonmcmurray: Similarly these are internal functions that don't need to be in the docs

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
...
+| `.dg.setwebreq`      | (bool: boolean) | Overrides functions to use https rather datadog agent |
+ 
+
+
+---
+
+
+## :label: Log Tables Schema
+
+Metric Log is stored in `.dg.dogstatsd_metriclog` with the following columns:
```
jonathonmcmurray: An explanation of what "Metric Log" means (and "Event Log" below) would be good here - it might seem obvious from schema but from perspective of someone looking at this for first time, best to be explicit.

datadog/datadog.md
```diff
@@ -0,0 +1,169 @@
+# `datadog.q` – Metric and event publishing to datadog for kdb+
```
jonathonmcmurray:
```suggestion
# `datadog.q` – Metric and event publishing to Datadog for kdb+
```

Capitalise Datadog anywhere it appears in docs

datadog/datadog.q
DHopkinson-DI: I believe powershell 5.1 is installed by default on windows 10 and above with earlier versions installed on previous os. I also believe its installed by default on all modern windows servers dating back to 2016, with early versions (4.1) installed on windows server 2012. I could look at integrating with cmd instead?

datadog/datadog.q
DHopkinson-DI: Testing `.Q.hp`  on both windows and macOS systems gives the error `'conn. OS reports: protocol not available.`  or a similar error . So I built `.dg.nonlinux_webreq` in response to that. I will move the hdel inside the function

datadog/datadog.q
DHopkinson-DI: I agree!

datadog/datadog.q
jonathonmcmurray: That just means that TLS certs aren't set up https://code.kx.com/q/kb/ssl/#tls-client-mode

We have these set up on homer, but on other Linux machines you'd likely get a similar error by default, and you can prevent the error on Windows/mac by setting up certs (or setting `export SSL_VERIFY_SERVER=NO` env var, but setting up certs is better)

datadog/datadog.q
jonathonmcmurray: Fair enough - probably worth documenting somewhere that it relies on powershell on Windows, just in case of any situtations where it could be absent

datadog/datadog.q
DHopkinson-DI: Do you think it would be better to just apply `.Q.hp`, remove `.dg.nonlinux_webreq` and make a note to either set up TLS certs or set the env var?

datadog/datadog.q
DHopkinson-DI: Yeah, wasn't sure the direction to go in, makes sense. Thanks, have made the changes

datadog/datadog.q
jonathonmcmurray: Yes I think that would be best

datadog/datadog.q
```diff
@@ -0,0 +1,88 @@
...
+ };
+
+// Utility functions used to manage send functions
+
+// Determine if web request is used or datadog agent, then assign appropriate functions
+.dg.setfunctions:{[useweb]
+    if[null method:$[useweb;`web;("lw"!`lin`win) first string .z.o];
+       '"Currently only linux and windows operating systems are supported to send metrics and events. Please use "".dg.setfunctions 1b"" to attempt a web request"];
+    .dg.sendmetric::.dg[method] `sendmetric;
+    .dg.sendevent::.dg[method] `sendevent;
```
jonathonmcmurray:
```suggestion
    .dg,:.dg[method];
```
think this should work the same here

jonathonmcmurray (dismissed):

di/datadog/datadog.q
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+/ utility functions used to manage send functions
+
+setfunctions:{[useweb]
+  / determine if web request is used or datadog agent, then assign appropriate functions
+  if[null method:$[useweb;`web;("lw"!`lin`win) first string .z.o];
+    '"Currently only linux and windows operating systems are supported to send metrics and events. Please use ""setfunctions 1b"" to attempt a web request"];
+  .z.m.sendmetric:value .Q.dd[.Q.dd[.z.M;method];`sendmetric];
+  .z.m.sendevent:value .Q.dd[.Q.dd[.z.M;method];`sendevent]
```
jonathonmcmurray:
```suggestion
  .z.m.sendmetric:value ` sv .z.M,method,`sendmetric;
  .z.m.sendevent:value ` sv .z.M,method,`sendevent
```

should do the same & IMO a bit cleaner

j-muldoon: accepted suggestion

di/datadog/datadog.q
```diff
@@ -5,22 +5,22 @@
...
 eventlog:([]time:`timestamp$();host:`$();message:();title:();text:();https:`boolean$();status:());
-opsys:.z.o; / pre-define opsertaing system to help with testing
+opsys:.z.o; / pre-define operating system to help with testing
 
 / following two functions used to push data to datadog agent on linux os
 
 lin.sendevent:{[eventtitle;eventtext;priority;tags;alerttype]
   / send event on linux os using datadog agent
-  cmd:raze"eventtitle=",eventtitle,"; eventtext=","\"",eventtext,"\"","; priority=","\"",priority,"\"","; tags=","\"#",$[0h=type tags;","sv tags;tags],"\"",";alerttype=",alerttype,"; ","echo \"_e{${#eventtitle},${#eventtext}}:$eventtitle|$eventtext|p:$priority|#$tags|t:$alerttype\" |nc -4u -w0 127.0.0.1 ",string .z.m.agentport;
+  cmd:raze"eventtitle=",eventtitle,"; eventtext=","\"",eventtext,"\"","; priority=","\"",priority,"\"","; tags=","\"#",$[0h=type tags;","sv tags;tags],"\"",";alerttype=",alerttype,"; ","echo \"_e{${#eventtitle},${#eventtext}}:$eventtitle|$eventtext|p:$priority|#$tags|t:$alerttype\" |nc -4u -w0 127.0.0.1 ",string agentport;
```
eliotrobinson: There has to be a better way to write this line... the other cmd lines arent as bad but i think this could do with a tidy up
```
  / send event on linux os using datadog agent
  cmd:raze
    ("eventtitle=";      eventtitle;
     "; eventtext=\"";   eventtext;
     "\"; priority=\"";  priority;
     "\"; tags=\"#";     $[0h=type tags;","sv tags;tags];
     "\";alerttype=";    alerttype;
     "; echo \"_e{${#eventtitle},${#eventtext}}:$eventtitle|$eventtext|p:$priority|#$tags|t:$alerttype\" |nc -4u -w0 127.0.0.1 ",string agentport
    );
```

di/datadog/datadog.q
```diff
@@ -5,22 +5,22 @@
...
   response:system cmd;
-  .z.m.eventlog,:(.z.p;.z.o;cmd;eventtitle;eventtext;0b;response);
+  eventlog,:(.z.p;opsys;cmd;eventtitle;eventtext;0b;response);
   };
 
 lin.sendmetric:{[metricname;metricvalue;tags]
   / send metric on linux os using datadog agent
-  cmd:raze"bash -c \"echo  -n '",metricname,":",(string metricvalue),"|g|#",$[0h=type tags;","sv tags;tags],"' > /dev/udp/127.0.0.1/",string .z.m.agentport,"\"";
+  cmd:raze"bash -c \"echo  -n '",metricname,":",(string metricvalue),"|g|#",$[0h=type tags;","sv tags;tags],"' > /dev/udp/127.0.0.1/",string agentport,"\"";
```
eliotrobinson:
```
cmd:"bash -c \"echo  -n '",metricname,":",string[metricvalue],"|g|#",$[0h=type tags;","sv tags;tags],"' > /dev/udp/127.0.0.1/",string[agentport],"\"";

```

di/datadog/datadog.q
```diff
@@ -29,7 +29,7 @@ pushtodogagent:{[message]
   / shell command to push data
   cmd:"powershell -Command \"";
   cmd,:" $udpClient = New-Object System.Net.Sockets.UdpClient;";
-  cmd,:" $udpClient.Connect('127.0.0.1','",raze string .z.m.agentport, "');";
+  cmd,:" $udpClient.Connect('127.0.0.1','",raze string agentport, "');";
```
eliotrobinson:
```
  cmd,:" $udpClient.Connect('127.0.0.1','",string[agentport],"');";
```

di/datadog/datadog.q
```diff
@@ -0,0 +1,89 @@
...
+  .z.m.sendmetric:value ` sv .z.M,method,`sendmetric;
+  .z.m.sendevent:value ` sv .z.M,method,`sendevent;
+  };
+
+init:{[useweb]
+  / initialisation function
+  .z.m.agentport:@[value;.z.M.agentport;"I"$getenv`DOGSTATSD_PORT];       / define datadog agent port
+  .z.m.apikey:@[value;.z.M.apikey;getenv`DOGSTATSD_APIKEY];               / define datadog api key
+  .z.m.baseurl:@[value;.z.M.baseurl;":https://api.datadoghq.eu/api/v1/"]; / define base api url
+  .z.m.setfunctions useweb;                                               / sets delivery method
```
eliotrobinson:
```
setfunctions useweb
```

di/datadog/datadog.q
jonathonmcmurray: if I'm reading this line right, it first sets a bunch of env vars and then uses those env vars within an `echo` - seems a bit redundant? Perhaps we should think about generating the final string directly from q?

Kx's recently released `printf` module may or may not be useful here as well? https://github.com/KxSystems/printf

```q
q)([printf]):use`kx.printf
q)eventtitle:"abc";eventtest:"abc def";priority:"low";tags:("tag1";"tag2");alerttype:"warning";
q)printf ("echo \"_e{%s,%s}:%s|%s|p:%s|#%s|t:%s\"";eventtitle;eventtest;eventtitle;eventtest;priority;$[0h=type tags;","sv tags;tags];alerttype)
"echo \"_e{abc,abc def}:abc|abc def|p:low|#tag1,tag2|t:warning\""
```

Would need to make it clear that the `printf` module was a dependency if we went with this approach.

benlord77: Would you recommend at the start of the script adding a line that loads the printf module?

jonathonmcmurray: yep

eliotrobinson: i would put it in the init.q function. @jonathonmcmurray did we ever figure anything out for dependency management? we could put a hacky way in here, something like (but def not limited to..)
```
if[not`printf in key .z.m;@[use;`di.printf;'"printf module not found, please install"]]
```

j-muldoon: With the printf setup, the cmd string looks something like:

```
"echo \""_e{Test_Event,This is a test}:Test_Event|This is a test|p:normal|#test|t:info\"" |nc -4u -w0 127.0.0.1 5010"
```

Is this what we are looking for? What is _e doing, as it appears to be expanded in versions before printf was added?

jonathonmcmurray: Not sure tbh, I was assuming that the `_e` was just part of the message that datadog expects for events - what do you mean it was "expanded in versions before printf was added"?

The main thing we need to test is if the module successfully publishes to datadog - we have a datadog agent installed on homer and I think I can give you access to the datadog web interface

j-muldoon: Hi, just checking past commits, _e wasn't expanded, I think I just got confused with the extra commands defining env variables before the echo. I'll have a look at using datadog to test now.

j-muldoon: not sure how to test if messages to agent are sent correctly yet, but we can confirm the format for metrics and events looks correct:

https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics.

di/datadog/datadog.q
```diff
@@ -0,0 +1,90 @@
...
+  .z.m.sendmetric:value ` sv .z.M,method,`sendmetric;
+  .z.m.sendevent:value ` sv .z.M,method,`sendevent;
+  };
+
+init:{[useweb]
+  / initialisation function
+  ([.z.m.printf]):use`kx.printf;
+  .z.m.agentport:@[value;.z.M.agentport;"I"$getenv`DOGSTATSD_PORT];       / define datadog agent port
+  .z.m.apikey:@[value;.z.M.apikey;getenv`DOGSTATSD_APIKEY];               / define datadog api key
+  .z.m.baseurl:@[value;.z.M.baseurl;":https://api.datadoghq.eu/api/v1/"]; / define base api url
```
j-muldoon: Should the base api url be retrieved from an environment variable as well?

di/datadog/datadog.q
jonathonmcmurray: Ah I missed that the first references to eventtitle and eventtext were counts (`#eventtitle`), I guess that needs a tweak - I think within the string those two should go from `%s` to `%d` and then in the following args they should be come e.g. `count eventtitle`

di/datadog/datadog.q
jonathonmcmurray: yes I think that makes sense, with a default in the module if env var is not set

we also need to update how these config varaiables get set e.g. see the usage module changes here https://github.com/DataIntellectTech/kdbx-modules/pull/57/files - we should have an init function that accepts config variables as a dictionary, and sets sensible defaults when nothing passed (in this case we should have preference of 1. values passed to init > 2. env vars > 3. sane defaults e.g. this URL, 8125 as default port number for dogstatd etc.)

di/datadog/datadog.q
j-muldoon: addressed this

di/datadog/datadog.q
j-muldoon: Would something like this work? Just checking before I dive in, as it would require large edits in tests and markdown.

```q
init:{[configs]

  / local function
  envcheck: {$[count x; x; y]};

  / Default values
  .z.m.agentport:envcheck["J"$getenv`DOGSTATSD_PORT;8125]; / define datadog agent port
  .z.m.apikey:getenv`DOGSTATSD_APIKEY; / define datadog api key - default value is empty string, so no need to check
  .z.m.baseurl:envcheck[getenv `DOGSTATSD_URL;":https://api.datadoghq.eu/api/v1/"]; / define base api url
  .z.m.useweb:envcheck["B"$getenv`DOGSTATSD_WEB;0b]; / define web use boolean

  / Values from config dictionary take priority
  if[not configs~(::);
    vars:`agentport`apikey`baseurl`useweb inter key configs;
    (.Q.dd[.z.M] each key[vars#configs]) set' value[vars#configs]
    ];

  / initialisation function
  if[not`printf in key .z.m;([.z.m.printf]):@[use;`kx.printf;{'"printf module not found, please install"}]]
  setfunctions useweb;                                               / sets delivery method
  };
```

di/datadog/datadog.q
```diff
@@ -11,7 +11,7 @@ opsys:.z.o; / pre-define operating system to help with testing
 
 lin.sendevent:{[eventtitle;eventtext;priority;tags;alerttype]
   / send event on linux os using datadog agent
-  cmd:printf ("echo \"_e{%s,%s}:%s|%s|p:%s|#%s|t:%s\" |nc -4u -w0 127.0.0.1 %s";eventtitle;eventtext;eventtitle;eventtext;priority;$[0h=type tags;","sv tags;tags];alerttype;agentport);
+  cmd:printf ("echo \"_e{%d,%d}:%s|%s|p:%s|#%s|t:%s\" |nc -4u -w0 127.0.0.1 %s";count eventtitle;count eventtext;eventtitle;eventtext;priority;$[0h=type tags;","sv tags;tags];alerttype;agentport);
```
j-muldoon: it looks like (from [here](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=events)) individual tags can be in two formats: #<TAG_KEY_1>:<TAG_VALUE_1> or <TAG_2>. Should this be enforced throughout?

di/datadog/datadog.q
j-muldoon: Additionally, how do we deal with variable paths to dependencies? e.g. if a user has placed printf in a directory that isn't `kx/`. Will we need to extend `use` to search an extra level deeper?

di/datadog/datadog.q
```diff
@@ -0,0 +1,91 @@
...
+/ following two functions used to push data to datadog agent on linux os
+
+lin.sendevent:{[eventtitle;eventtext;priority;tags;alerttype]
+  / send event on linux os using datadog agent
+  cmd:printf ("echo \"_e{%d,%d}:%s|%s|p:%s|#%s|t:%s\" |nc -4u -w0 127.0.0.1 %s";count eventtitle;count eventtext;eventtitle;eventtext;priority;$[0h=type tags;","sv tags;tags];alerttype;agentport);
+  response:system cmd;
+  eventlog,:(.z.p;opsys;cmd;eventtitle;eventtext;0b;response);
+  };
+
+lin.sendmetric:{[metricname;metricvalue;tags]
```
j-muldoon: Working on a way to pass in variable number of arguments to sendmetric - as in the datadog docs, only the first two arguments are required. The below idea seems to work  (apart from with the list of tags, but that can be worked on.). You can pass in an empty string "" to skip an optional argument.

```q
smlambda:{[arglist]

    if[("" in 2#arglist) or 2> count arglist; '"Requires first two args - metricname and metricvalue"];

    /- Take optional argument of "" to skip entry (handled in the qSQL at the end)
    /- Extra work required with the taglist (or maybe we can just require it to have items in one of two formats?)
    vars:`metricname`metricvalue`metrictype`samplerate`taglist;
    leaders:("";":";"|";"|@";"|#:,|");
    n:min (count arglist;count vars);
    t:([] vars:(n#vars); vals: (n#arglist); leaders:(n#leaders));
    raze exec joined from update joined:raze each (leaders,'vals) from t where not 0=count each vals
 };

sendmetric:('[smlambda;enlist]);
```
Or a more concise version that also keeps track of metricname and metricvalue for the log:
```q
smlambda:{[arglist]
    /- TODO: Add enforcing of tag formats
    if[("" in 2#arglist) or 2> count arglist; '"Requires first two args - metricname and metricvalue"];
    metricname:arglist 0;
    metricvalue:arglist 1;
    leaders:("";":";"|";"|@";"|#:,|");
    n:min (count arglist;count leaders);
    /- Take optional argument of "" to skip entry
    raze ((n#leaders),' n#arglist) where not 0=count each n#arglist
 }
```
Some examples:

```q
sendmetric["test"]
'Requires first two args - metricname and metricvalue
  [0]  sendmetric["test"]
```
```q
sendmetric["";"value"]
'Requires first two args - metricname and metricvalue
  [0]  sendmetric["";"value"]
       ^
```
```q
sendmetric["name";"value";"type";""]
"name:value|type"
```
```q
sendmetric["name";"value";"";"samplerate"]
"name:value|@samplerate"
```
```q
sendmetric["name";"value";"type";"samplerate"]
"name:value|type|@samplerate"
```

j-muldoon: The following is a breaking change (which can be transformed to be non breaking by projecting empty strings into the optional type and samplerate arguments). The order here is used because it is consistent with the order of arguments to the datadog message format. Examples of the value of `cmd` are shown afterwards. Note that nulls in first two arguments still needs to be properly handled. On that note, does anyone have any tips with falsey vs truthy in kdb? 

This is a proof of concept that can be extended to provide the correct datadog message format wherever required. It is different from current approach as optional arguments are now exactly that - optional.

```q
lin.smlambda:{[arglist]
  if[("" in st:2#arglist) or 2> count arglist; '"Requires first two args - metricname and metricvalue"];
  / Convert metric value to string
  arglist[1]:raze string arglist[1];
  / List of separators for datadog metric format: 
  / <METRIC_NAME>:<VALUE>|<TYPE>|@<SAMPLE_RATE>|#<TAG_KEY_1>:<TAG_VALUE_1>,<TAG_2>
  leaders:("";":";"|";"|@";"|#");
  if[(count leaders)<count arglist; '"rank"];
  / Comma separate list of tags (if index doesn't exist, indexing returns "")
  if[not 0 = count arglist[4]; arglist:@[arglist;4;{[tags] $[0h=type tags;"," sv tags;tags]}]];
  
  n:min (count arglist;count leaders);
  / Take optional argument of "" to skip entry
  ddmsg: raze ((n#leaders),' n#arglist) where not 0=count each n#arglist;
  / Define name and value variables
  metricname:arglist 0;
  metricvalue:arglist 1;

  cmd:printf("bash -c \"echo  -n '%s' > /dev/udp/127.0.0.1/%s\"";ddmsg;string agentport);
  response:system cmd;
  metriclog,:(.z.p;opsys;cmd;metricname;"F"$metricvalue;0b;response)
 };

/ Allow variable number of arguments to be passed
lin.sendmetric:('[smlambda;enlist]);
```

Examples of value of `cmd`:

```q
sendmetric["name";"value";"type";"samplerate";"tag2"]
"bash -c \"echo  -n 'name:value|type|@samplerate|#tag2' > /dev/udp/127.0.0.1/8125\""

sendmetric["name";"value";"type";"samplerate";("tagkey:tagvalue";"tag2")]
"bash -c \"echo  -n 'name:value|type|@samplerate|#tagkey:tagvalue,tag2' > /dev/udp/127.0.0.1/8125..

sendmetric["name";"value";"type";"samplerate"]
"bash -c \"echo  -n 'name:value|type|@samplerate' > /dev/udp/127.0.0.1/8125\""

sendmetric["name";"value";"";"samplerate"]
"bash -c \"echo  -n 'name:value|@samplerate' > /dev/udp/127.0.0.1/8125\""

sendmetric["name";"value";"";"";("tagkey:tagvalue";"tag2")]
"bash -c \"echo  -n 'name:value|#tagkey:tagvalue,tag2' > /dev/udp/127.0.0.1/8125\""

sendmetric["name";"";"type";"";("tagkey:tagvalue";"tag2")]
'Requires first two args - metricname and metricvalue
  [0]  sendmetric["name";"";"type";"";("tagkey:tagvalue";"tag2")]
       ^

sendmetric["name"]
'Requires first two args - metricname and metricvalue
  [0]  sendmetric["name"]
       ^

sendmetric["name";0N;"type";"";("tagkey:tagvalue";"tag2")]
"bash -c \"echo  -n 'name|type|#tagkey:tagvalue,tag2' > /dev/udp/127.0.0.1/8125\""
```

jonathonmcmurray: I think it's better/cleaner to just pass a dictionary to the function to allow for optional params i.e. have a couple of keys that _must_ be in dict (can use pattern-matching syntax in function argument declaration to enforce this) and the rest of keys are optional

di/datadog/datadog.q
jonathonmcmurray: I think that can be left to the user to either pass as `:` separated string or individual string, should work the same either way I think? Either way they are passing string/list of strings

di/datadog/datadog.q
jonathonmcmurray: yeah I think something like that looks reasonable - don't need an env var for DOGSTATD_WEB, just default to 0b unless that is passed to init function.

don't worry about searching other paths for the `printf` module - can just output an error if it's not installed in the expected location (i.e. `use` fails to find it)

di/datadog/datadog.q
j-muldoon: Ok, sounds good. Do you think the idea to be sending datadog messages with optional arguments is something I should work on through the whole file as a task? Or are there other priorities?

jonathonmcmurray: yeah I think this would be a good idea

j-muldoon: Hi, I have added a lot of checks and conversions here, and the input is a dictionary. I first attempted just normal pattern matching with `([metricname;metricvalue]):dict` which raises a `'match` error if those keys dont exist in the dictionary. This worked fine, and pattern matching on the optional arguments with 

```q
(metrictype;samplerate;taglist): dict[`metrictype`samplerate`taglist]
```
meant the variables were saved as an empty (with type corresponding to whatever the type of the first value in the dictionary was). This is fine, as no error is raised, and as long as there is no comparison being done on the variable empties that might come out this way. However, in any case, I couldn't avoid manually checking the keys of the input dictionary, both for error checks, and checks in the main action of the function.  (see where I use `pars: key dict`). This feels like something pattern matching can be useful for but I couldn't think of a good way to implement.

Filter function is used to sanitise the input. What sort of standard are we looking for for sanitisation. Also, should errors be thrown, or is there a preferred approach for these kdb-x modules?

The good news is, the filter function can be chucked straight into other metric send functions, and the approach can be easily converted to work for sending events as well.


```q
metfilter:{[dict]
  
  requiredpars:`metricname`metricvalue;
  optionalpars: `metrictype`samplerate`tags;
  validpars: requiredpars,optionalpars;

  / Checks
  if[not 99h=type dict; '"input must be a dictionary"];
  pars: key dict;
  if[(count validpars)<count dict; '"rank"];
  if[not (count pars)=count distinct pars; '"keys must be unique"];
  if[any not pars in validpars; '"valid argument names: ", csv sv string each validpars];
  if[not all requiredpars in pars; '"required arguments: ", csv sv string each requiredpars];
  / Conversions
  dict[`metricvalue]: string dict[`metricvalue];
  if[`samplerate in pars;dict[`samplerate]:string dict[`samplerate]];
  if[`tags in pars;dict[`tags]:{$[0h=type x;"," sv x;x]} dict[`tags]];

  / Standardise order to fit datadog format
  k!dict[k:validpars where validpars in pars]
 }
sendmetric:{[dict:metfilter]

  (pars;args): (key;value)@\:dict;
  leaders:([metricname:"";metricvalue:":";metrictype:"|";samplerate:"|@";tags:"|#"]);
  / Match error here if `metricvalue`metricname keys not in dict
  ([metricname;metricvalue]):dict;

  ddmsg: raze (leaders[pars]),'(args);
  cmd:printf("bash -c \"echo  -n '%s' > /dev/udp/127.0.0.1/%s\"";ddmsg;string agentport);
  response:system cmd;
  metriclog,:(.z.p;opsys;cmd;metricname;"F"$metricvalue;0b;response)
 }
```

jonathonmcmurray: I think it'll be easier to comment and feedback on this if you push the change, so we can comment on individual lines etc. broadly speaking I think it looks like it's along the right lines

di/datadog/datadog.q
```diff
@@ -0,0 +1,135 @@
...
+lin.sendevent:{[eventtitle;eventtext;priority;tags;alerttype]
+  / send event on linux os using datadog agent
+  cmd:printf ("echo \"_e{%d,%d}:%s|%s|p:%s|#%s|t:%s\" |nc -4u -w0 127.0.0.1 %s";count eventtitle;count eventtext;eventtitle;eventtext;priority;$[0h=type tags;","sv tags;tags];alerttype;string[agentport]);
+  response:system cmd;
+  eventlog,:(.z.p;opsys;cmd;eventtitle;eventtext;0b;response);
+  };
+
+lin.sendmetric:{[dict:metfilter]
+  / extract dictionary values
+  (pars;args): (key;value)@\:dict;
```
jonathonmcmurray: can simplify this further

```suggestion
lin.sendmetric:{[dict:metfilter]
  / extract dictionary values
  (pars!args):dict;
```

or even


```suggestion
lin.sendmetric:{[(pars!args):metfilter]
```

di/datadog/datadog.q
```diff
@@ -0,0 +1,135 @@
...
+  leaders:([metricname:"";metricvalue:":";metrictype:"|";samplerate:"|@";tags:"|#"]);
+  / match error here if `metricvalue`metricname keys not in dict
+  ([metricname;metricvalue]):dict;
```
jonathonmcmurray: isn't this already checked in `metfilter`? don't think we need to check this here

di/datadog/datadog.q
```diff
@@ -0,0 +1,135 @@
...
+  if[not (count pars)=count distinct pars; '"keys must be unique"];
+  if[any not pars in validpars; '"valid argument names: ", csv sv string each validpars];
+  if[not all requiredpars in pars; '"required arguments: ", csv sv string each requiredpars];
+  / Conversions
+  dict[`metricvalue]: string dict[`metricvalue];
+  if[`samplerate in pars;dict[`samplerate]:string dict[`samplerate]];
+  if[`tags in pars;dict[`tags]:{$[0h=type x;"," sv x;x]} dict[`tags]];
+
+  / Standardise order to fit datadog format
+  k!dict[k:validpars where validpars in pars]
```
jonathonmcmurray:
```suggestion
  / Standardise order to fit datadog format
  (validpars inter key k)#k
```

di/datadog/datadog.q
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+lin.sendmetric:{[dict:metfilter]
+  / extract dictionary values
+  (pars;args): (key;value)@\:dict;
+  leaders:([metricname:"";metricvalue:":";metrictype:"|";samplerate:"|@";tags:"|#"]);
+  / match error here if `metricvalue`metricname keys not in dict
+  ([metricname;metricvalue]):dict;
+
+  ddmsg: raze (leaders[pars]),'(args);
```
jonathonmcmurray:
```suggestion
  ddmsg:raze (leaders[pars]),'(args);
```

di/datadog/init.q
```diff
@@ -0,0 +1,20 @@
...
+/ exportable getter functions for event and metric table
+getmetriclog:{.z.m.metriclog};
+
+geteventlog:{.z.m.eventlog};
+
+/ exportable functions to access the relevant send event/metric functions
+/ the releveant metric and event functions are assinged when init is run
+exportsendmetric:{[dict].z.m.sendmetric[dict]};
+
+exportsendevent:{[eventtitle;eventtext;priority;tags;alerttype].z.m.sendevent[eventtitle;eventtext;priority;tags;alerttype]};
```
jonathonmcmurray: why are these necessary? why not just export `sendmetric` and `sendevent` directly?

di/datadog/datadog.q
j-muldoon: It's also used to extract those values and is a safety against any tinkering with metfilter in the future. The alternative is 
```q
(metricname;metricvalue):dict[`metricname`metricvalue]
```

Is the alternative preferred?

j-muldoon: With the pars!args inside signature change, I have changed this to `(metricname;metricvalue):args 0 1`.

di/datadog/datadog.q
```diff
@@ -5,15 +5,60 @@
 / define tables to capture events and metrics
 metriclog:([]time:`timestamp$();host:`$();message:();name:();metric:`float$();https:`boolean$();status:());
 eventlog:([]time:`timestamp$();host:`$();message:();title:();text:();https:`boolean$();status:());
-opsys:.z.o; / pre-define operating system to help with testing
+/ pre-define operating system to help with testing
+opsys:.z.o;
+
+
+/eventtitle,eventtext,eventdate,hostname,priority,alerttype,tags,aggregation_key,source_type_name,related_event_id,device,category,cr_name,cr_type,integration_id,message,timestamp,change_metadata,new_value,prev_value
```
jonathonmcmurray: what is this for?

di/datadog/datadog.q
```diff
@@ -67,37 +112,52 @@ pushtodogagent:{[message]
...
   / sends events via https post to datadog api
   url:baseurl,"events?api_key=",apikey;
-  json:.j.j`title`text`priority`tags`alert_type!(eventtitle;eventtext;priority;$[0h=type tags;","sv tags;tags];alerttype);
+  if[(first baseurlversion)=`v1;
+  json:.j.j `title`text`priority`tags`alert_type`aggregation_key`date_happened`device_name`host`related_event_id`source_type_name!(eventtitle;eventtext;priority;tags;alerttype;aggregation_key;eventdate;device;hostname;related_event_id;source_type_name)];
+  if[(first baseurlversion)=`v2;(
+  if[(`$category)=`$"alert";
+  json:.j.j(enlist`data)!enlist(`attributes`type!(enlist(`aggregation_key`attributes`category`integration_id`message`tags`timestamp`title!(aggregation_key;enlist(`custom`links`priority`status!(custom;enlist(`category`title`url!(linkcategory;linktitle;linkurl));priority;status));category;integration_id;message;tags;timestamp;title);alerttype)))];
+  / if[(`$category)=`$"change";
```
jonathonmcmurray: don't leave commented out code, if this isn't needed, remove it

di/datadog/datadog.q
```diff
@@ -67,37 +112,52 @@ pushtodogagent:{[message]
...
+  (eventtitle;eventtext):args 0 1;
+  leaders:([eventtitle:printf("_e{%d,%d}:";count eventtitle;count eventtext);
+    eventtext:"|";eventdate:"|d:";hostname:"|h:";priority:"|p:";alerttype:"|t:";tags:"|#"]);
+  ddmsg:raze (leaders[pars]),'(args);
+  response:raze@[pushtodogagent;ddmsg;{'"Error pushing data to agent: ",x}];
+  eventlog,:(.z.p;.z.h;ddmsg;eventtitle;eventtext;0b;response);
   };
 
-/ the following two functions are used to push data to datadog through https post using .Q.hp
+/ the following two functions are used to push data to datadog through https post usSingSS .Q.hp
```
jonathonmcmurray: seem to be a few extra Ss in this comment

di/datadog/datadog.q
```diff
@@ -67,37 +112,52 @@ pushtodogagent:{[message]
...
   };
 
-win.sendmetric:{[metricname;metricvalue;tags]
-  / override used to send metrics through windows powershell to datadog agent
-  metric:raze metricname,":",string metricvalue,"|g|#",$[0h=type tags;","sv tags;tags];
-  response:raze@[pushtodogagent;metric;{'"Error pushing data to agent: ",x}];
-  metriclog,:(.z.p;.z.h;metric;metricname;`float$metricvalue;0b;response);
+win.sendmetric:{[(pars!args):metfilter]
+  leaders:([metricname:"";metricvalue:":";metrictype:"|";samplerate:"|@";tags:"|#"]);
+  (metricname;metricvalue):args 0 1;
```
jonathonmcmurray: this assumes that metricname & metricvalue are first two keys of dictionary passed in, do we want to enforce that?

di/datadog/datadog.q
```diff
@@ -5,15 +5,60 @@
...
-  
   requiredpars:`metricname`metricvalue;
   optionalpars: `metrictype`samplerate`tags;
-  validpars: requiredpars,optionalpars;
 
+  / Filters to check web and url version to filter required params
+  if[useweb;(requiredpars:`metric`points;optionalpars:`interval`tags`type)];
+  if[useweb;( 
+    if[(first baseurlversion)=`v1;optionalpars,:`host];
+    if[(first baseurlversion)=`v2;optionalpars,:`unit`source_typename`resource_name`resource_type`origin_metric_type`origin_product`origin_service])];
```
jonathonmcmurray:
```suggestion
  if[useweb;requiredpars:`metric`points;optionalpars:`interval`tags`type;
    if[(first baseurlversion)=`v1;optionalpars,:`host];
    if[(first baseurlversion)=`v2;optionalpars,:`unit`source_typename`resource_name`resource_type`origin_metric_type`origin_product`origin_service]
  ];
```

I think this is the same, with less unnecessary brackets etc.

di/datadog/datadog.q
```diff
@@ -112,50 +110,52 @@ pushtodogagent:{[message]
   response
   };
 
-win.sendmetric:{[(pars!args):metfilter]
+win.sendmetric:{[dict:metfilter]
```
jonathonmcmurray, line 113:
```suggestion
win.sendmetric:{[dict]
```
pretty sure this is the same when you're not unpacking to pars & args

(same for any other functions with same pattern)

jonathonmcmurray (approved):

PR #5, opened 2025-07-21
Add dataloader lib

### LINKED ISSUE #38: Package: Dataloader
### Discussed in https://github.com/DataIntellectTech/kdbx-packages/discussions/17

<div type='discussions-op-text'>

<sup>Originally posted by **jamiechandler99** July 23, 2025</sup>
Utilities for dataloading, particularly for chunking  loading of large files 

Potential start point: 
- [TorQ Write Utils Documentation](https://dataintellecttech.github.io/TorQ/utilities/#dbwriteutils.q)
- [TorQ Write Utils Code](https://github.com/DataIntellectTech/TorQ/blob/master/code/common/dbwriteutils.q)

PR [#5](https://github.com/DataIntellectTech/kdbx-packages/pull/5)</div>

DHopkinson-DI:
Same concept as TorQ's/Kx dataloader script.
Tested for kdb+ 4.1 on Linux

Fixes #38


dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+// partitiontype [optional] = the partition type - one of `date`month`year`int.  Default is `date
+// partitioncol [optional] = the name of the column to cast to the partition type to work out which partition the data should go in.  default is `time
+// chunksize [optional] = size of data chunks in bytes to read at a time.  default is 100MB
+// compression [optional] = compression parameters to use. list of 3 integers e.g. 17 2 6.
+// filepattern [optional] = specify pattern used to filter files
+// These are only set when the data is sorted on disk (in the finish function) to save on writing the data compressed, reading in and uncompressing, sorting, and writing out compressed again
+// gc [optional] = boolean flag to turn garbage collection on and off.  Default is 0b
+
+// e.g.
+// .loader.loadallfiles[`headers`types`separator`tablename`dbdir!(`sym`time`price`volume`mktflag`cond`exclude;"SPFICHB";",";`tdc;`:hdb); `:TDC/toload]
```
jonathonmcmurray: I think all this can be removed, should be covered by docs

dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+// Initialisation function
+.loader.init:{[sortparams]
+    .loader.partitions:()!();                                                                              // Maintain a dictionary of the db partitions which have been written to by the loader
+    .loader.filesread:();                                                                                  // Maintain a list of files which have been read
+    .loader.sortparams:$[not 99=type sortparams;
+                         '"Error, .loader.init requires a dictionary as the first parameter";
+                         not all `tabname`att`column`sort in key sortparams;
+                         '"Error, missing keys in dictionary. Please ensure `tabname`att`column`sort are all keys in the dictionary";
+                         .loader.sortparams:$[1=max count each value sortparams;flip (key sortparams)!(enlist each value sortparams);flip sortparams]
+                        ];
```
jonathonmcmurray: We might be able to make use of pattern matching/filter function here: https://code.kx.com/q/basics/pattern/#filter-function

i.e. define a separate function for validating sort param and then do something like `.loader.init:{[sortparams:validsortparams]`

Given packages are going to be targeting kdb-x, I think it's reasonable to assume that we can make use of newer kdb features like this

dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+
+// Initialisation function
+.loader.init:{[sortparams]
+    .loader.partitions:()!();                                                                              // Maintain a dictionary of the db partitions which have been written to by the loader
+    .loader.filesread:();                                                                                  // Maintain a list of files which have been read
+    .loader.sortparams:$[not 99=type sortparams;
+                         '"Error, .loader.init requires a dictionary as the first parameter";
+                         not all `tabname`att`column`sort in key sortparams;
+                         '"Error, missing keys in dictionary. Please ensure `tabname`att`column`sort are all keys in the dictionary";
+                         .loader.sortparams:$[1=max count each value sortparams;flip (key sortparams)!(enlist each value sortparams);flip sortparams]
```
jonathonmcmurray: don't need `.loader.sortparams:` here as that is already outside the original `$` conditional

think you can also simplify the enlisting

```suggestion
                        flip $[1=max count each value sortparams;enlist each sortparams;sortparams]
```

depending on the types involved you might even be able to just do


```suggestion
                        flip (),'sortparams
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+    reqtypes:`headers`types`tablename`dbdir`symdir`chunksize`partitioncol`partitiontype`gc!`short$(11;10;-11;-11;-11;-6;-11;-11;-1);               // Required types n
+
+    if[count w:where not (type each loadparams key reqtypes)=reqtypes;                                       // Check the types
+       '"incorrect types supplied for ",(", " sv string w)," parameter(s). Required type(s) are ",", " sv string reqtypes w];
+    if[not 10h=abs type loadparams`separator;'"separator must be a character or enlisted character"];
+    if[not 99h<type loadparams`dataprocessfunc;'"dataprocessfunc must be a function"];
+    if[not loadparams[`partitiontype] in `date`month`year`int;'"partitiontype must be one of `date`month`year`int"];
+    if[not count[loadparams`headers]=count loadparams[`types] except " ";'"headers and non-null separators must be the same length"];
+    if[c:count loadparams[`compression]; if[not (3=c) and type[loadparams[`compression]] in 6 7h; '"compression parameters must be a 3 item list of type int or long"]];
```
jonathonmcmurray: Again I think this validation logic could be put into a separate function & use the filter function feature

dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+           flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata;                  // If it hasn't been read then we have to just read it as a list of lists
+           [.loader.filesread,::loadparams[`filename];                                                      // It hasn't been seen - the first row may or may not be column headers
+           $[all (`$"," vs rawdata[0;]) in loadparams[`headers];
+              (loadparams[`types];enlist loadparams[`separator])0:rawdata;
+              flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata]]];
+    data:0!loadparams[`dataprocessfunc] . (loadparams;data);                                                // Do some optional extra processing
+    data:$[`symdir in key loadparams;                                                                       // Enumerate the table - best to do this once /.lg.o[`dataloader;"Enumerating"];
+           .Q.en[loadparams[`symdir];data];
+           .Q.en[loadparams[`dbdir];data]];
```
jonathonmcmurray:
```suggestion
    data:.Q.en[loadparams $[`symdir in key loadparams;`symdir;`dbdir];data];
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,127 @@
...
+
+
+// Functions loads data in from delimited file, applies processing function, enumerates it then writes to db
+.loader.loaddata:{[loadparams;rawdata]
+    data:$[loadparams[`filename] in .loader.filesread;                                                      // Check if we have already read some data from this file. First row may contain the header information in both cases we want to return a table with the same column names
+           flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata;                  // If it hasn't been read then we have to just read it as a list of lists
+           [.loader.filesread,::loadparams[`filename];                                                      // It hasn't been seen - the first row may or may not be column headers
+           $[all (`$"," vs rawdata[0;]) in loadparams[`headers];
+              (loadparams[`types];enlist loadparams[`separator])0:rawdata;
+              flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata]]];
```
jonathonmcmurray: avoid the block syntax please

DHopkinson-DI: Block syntax removed

dataloader/dataloader.q
DHopkinson-DI: Made suggested change

dataloader/dataloader.q
DHopkinson-DI: removed redundant comments

dataloader/dataloader.q
DHopkinson-DI: Yes, my mistake have removed the erroneous assignment. Thanks for the suggestion I have gone with `flip (),/: sortparams`

dataloader/dataloader.q
DHopkinson-DI: Made suggest change by applying filter function feature and separate validation logic

dataloader/dataloader.md
```diff
@@ -0,0 +1,125 @@
...
+| `types`           | Y        | char list | Data types to read from the file                                                                                                                                                |
+| `separator`       | Y        | char list | Delimiting character. Enlist it if first line of file is header data                                                                                                            |
+| `tablename`       | Y        | symbol    | Name of table to write data to                                                                                                                                                  |
+| `dbdir`           | Y        | symbol    | Directory to write data to                                                                                                                                                      |
+| `symdir`          | N        | symbol    | Directory to enumerate against                                                                                                                                                  |
```
jonnypress, line 79: As well as having symdir as a parameter, could we add `enumname or similar, which defaults to `sym. So basically we use .Q.ens rather than .Q.en

dataloader/dataloader.md
```diff
@@ -0,0 +1,125 @@
+# `dataloader.q` –  Loading delimited data and creating databases for kdb+
+
+
+This package is used for automated customisable dataloading and database creation and is a generalisation of http://code.kx.com/wiki/Cookbook/LoadingFromLargeFiles. 
+Load all delimeted files in a directory into memory in configurable chunk sizes then output the resulting tables to disk in kdb+ partiioned format.
```
jonnypress: partiioned - typo

I think this is underselling it a bit. It will not load it all into memory and then to disk. It will load the data chunk by chunk, so the aim is to mininimise memory usage. The memory usage for this should be related to the maximum of 

- the space required to load into memory and save one chunk of data 
- the memory required to sort the resultant table

So therefore we should be able to load large volumes of on-disk data using a relatively small memory footprint. 

A couple of examples would be good: 

1. loading very large files, but in small chunks 
2. loading data across partitions from a number of small files e.g. if we had a month worth of AAPL data in one file, and a month of MSFT data in one file, I believe the way this is structured it would handle it relatively efficiently.

dataloader/dataloader.md
jamiechandler99: Have updated to use this. Test case included

dataloader/dataloader.md
jamiechandler99: Added documentation highlighting the advantages of chunking

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+  filesread::();
+  filelist:$[`filepattern in key loadparams;
+    (key dir:hsym dir) where max like[key dir;] each loadparams[`filepattern];
+    key dir:hsym dir]; / get the contents of the directory based on optional filepattern
+  filelist:` sv' dir,'filelist;
+  {[loadparams;file] .Q.fsn[loaddata[loadparams,(enlist`filename)!enlist file];file;loadparams`chunksize]}[loadparams] each filelist;
+  .z.m.finish[loadparams];
+  };
+
+init:{[sortparams:.m.dataloader.util.sortfilter]
```
jonathonmcmurray:
```suggestion
init:{[sortparams:.z.m.util.sortfilter]
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+
+finish:{[loadparams]
+  / adds compression, sorting and attributes selected
+  if[count loadparams `compression;.z.zd:loadparams`compression]; / temporarily set compression defaults
+  {.m.dataloader.util.sorttab (x;where .z.m.partitions[;0]=x)} each distinct value .z.m.partitions[;0];
+  system"x .z.zd";
+  if[loadparams`gc; .Q.gc[]];
+  };
+
+loadallfiles:{[loadparams:.m.dataloader.util.paramfilter;dir]
```
jonathonmcmurray:
```suggestion
loadallfiles:{[loadparams:.z.m.util.paramfilter;dir]
```

dataloader/init.q
```diff
@@ -0,0 +1,6 @@
+//Load core functionality into root module namespace
+\l ::dataloader.q
+//Create secondary namespace level and load relevant script into it
```
jonathonmcmurray:
```suggestion
// load util submodule
```

dataloader/init.q
```diff
@@ -0,0 +1,6 @@
+//Load core functionality into root module namespace
```
jonathonmcmurray:
```suggestion
// load core functionality into root module namespace
```

dataloader/init.q
```diff
@@ -0,0 +1,6 @@
...
+\l ::dataloader.q
+//Create secondary namespace level and load relevant script into it
+util:use`dataloader.util
+//Return to root module namespace to simplify exposure of public functions
```
jonathonmcmurray:
```suggestion
// expose public functions
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
+/ generic dataloader library
+
+loaddata:{[loadparams;rawdata]
+  / loads data in from delimeted file, applies processing function, enumerates and writes to db
+  data:$[loadparams[`filename] in filesread; / check if some data has already been read in
+    flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata;
+    $[all (`$"," vs rawdata[0;]) in loadparams[`headers]; / it hasn't been seen, may be column headers
+      (loadparams[`types];loadparams[`separator])0:rawdata;
+      flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata]
+    ];
```
jonathonmcmurray: there seems to be some redundancy here line 6 & line 9 appear to be the same? think this could be rewritten to simplify a bit

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+loaddata:{[loadparams;rawdata]
+  / loads data in from delimeted file, applies processing function, enumerates and writes to db
+  data:$[loadparams[`filename] in filesread; / check if some data has already been read in
+    flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata;
+    $[all (`$"," vs rawdata[0;]) in loadparams[`headers]; / it hasn't been seen, may be column headers
+      (loadparams[`types];loadparams[`separator])0:rawdata;
+      flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata]
+    ];
+  if[not loadparams[`filename] in filesread;filesread,::loadparams[`filename]];
+  .c.t:(loadparams;filesread);
```
jonathonmcmurray: assuming this is a debug line that should be removed before we merge this

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+  data:0!loadparams[`dataprocessfunc] . (loadparams;data);
+  data:.Q.ens[loadparams $[`symdir in key loadparams;`symdir;`dbdir];data;loadparams[`enumname]];
+  writedatapartition[loadparams[`dbdir];;loadparams[`partitiontype];loadparams[`partitioncol];loadparams[`tablename];data] each distinct loadparams[`partitiontype]$data[loadparams`partitioncol];
```
jonathonmcmurray: this line seems a bit unwieldly. perhaps we could re-order the params to `writedatapartition` and do something like

```suggestion
  / project writedatapartition with relevant loadparams
  write:writedatapartition[data] . loadparams`dbdir`partitiontype`partitionco`tablename;
  / write data for each partition
  write each distinct loadparams[`partitiontype]$data[loadparams`partitioncol];
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+  towrite:data where partition=partitiontype$data partitioncol;
+  writepath:` sv .Q.par[dbdir;partition;tablename],`;
+  .[upsert;(writepath;towrite);{'"failed to save table: ",x}];
+  .z.m.partitions[writepath]:(tablename;partition);
+  };
+
+finish:{[loadparams]
+  / adds compression, sorting and attributes selected
+  if[count loadparams `compression;.z.zd:loadparams`compression]; / temporarily set compression defaults
+  {.m.dataloader.util.sorttab (x;where .z.m.partitions[;0]=x)} each distinct value .z.m.partitions[;0];
```
jonathonmcmurray:
```suggestion
  {.z.m.util.sorttab (x;where .z.m.partitions[;0]=x)} each distinct value .z.m.partitions[;0];
```

dataloader/dataloader.q
```diff
@@ -0,0 +1,52 @@
...
+  system"x .z.zd";
+  if[loadparams`gc; .Q.gc[]];
+  };
+
+loadallfiles:{[loadparams:.m.dataloader.util.paramfilter;dir]
+  / load all the files from a specified directory
+  .z.m.partitions::()!();
+  filesread::();
```
jonathonmcmurray: seems to be some inconsistency here - I think we should either use `.z.m` in both lines or not. If we use `.z.m`, we should use `:` instead of `::`

dataloader/util.q
```diff
@@ -0,0 +1,54 @@
+applyattr:{[dloc;colname;att]
+    .[{@[x;y;z#]};(dloc;colname;att);                                                                        //  Attempt to apply the attribute to the column and log an error if it fails
```
jonathonmcmurray: use 2 spaces for each level of indentation, as per style guide

dataloader/util.q
```diff
@@ -0,0 +1,54 @@
...
+      {[dloc;colname;att;e]
+          '"unable to apply ",string[att]," attr to the ",string[colname]," column in the this directory : ",string[dloc],". The error was : ",e;
+      }[dloc;colname;att]
+     ]
+ };
+
+// Function used to sort and apply attributes to tables on disk based on format provided at initialisation of package.
+sorttab:{[d]
+    if[1>sum exec sort from .m.dataloader.sortparams;:()];
```
jonathonmcmurray:
```suggestion
    if[1>sum exec sort from .z.m.sortparams;:()];
```

although as this is a submodule, this probably won't work - I think we need to pass `sortparams` into this function instead of referencing a global.

dataloader/util.q
```diff
@@ -0,0 +1,54 @@
...
+
+// Function checks keys are correct and value have the right types for loadallfiles argument
+paramfilter:{[loadparams]
+    if[not 99h=type loadparams; '"loadallfiles requires a dictionary parameter"];                   // Check the input
+    req:`headers`types`tablename`dbdir`separator;                                                           // Required fields
+    if[not all req in key loadparams;
+       '"loaddata requires a dictionary parameter with keys of ",(", " sv string req)," : missing ",", " sv string req except key loadparams];
+    if[not count loadparams `symdir;loadparams[`symdir]:loadparams[`dbdir]];
+    loadparams:(`dataprocessfunc`chunksize`partitioncol`partitiontype`compression`gc!({[x;y] y};`int$100*2 xexp 20;`time;`date;();0b)),loadparams; // Join the loadparams with some default values
+    reqtypes:`headers`types`tablename`dbdir`symdir`chunksize`partitioncol`partitiontype`gc!`short$(11;10;-11;-11;-11;-6;-11;-11;-1);               // Required types n
```
jonathonmcmurray:
```suggestion
    reqtypes:`headers`types`tablename`dbdir`symdir`chunksize`partitioncol`partitiontype`gc!11 10 -11 -11 -11 -6 -11 -11 -1h;               // Required types n
```

dataloader/util.q
```diff
@@ -0,0 +1,54 @@
...
+    if[c:count loadparams[`compression]; if[not (3=c) and type[loadparams[`compression]] in 6 7h; '"compression parameters must be a 3 item list of type int or long"]];
+    if[(`filepattern in key loadparams) & 10h=type loadparams[`filepattern];                              // If a filepattern was specified ensure that it's a list
+       loadparams[`filepattern]:enlist loadparams[`filepattern]];
+    loadparams
+ };
+
+// Function checks dictionary argument for init function has correct headers and types
+sortfilter:{[sortparams]
+    if[not 99h=type sortparams; '"init requires a dictionary parameter"];
+    if[not (abs type each value sortparams)~11 11 11 1h;'"Error ensure dictionary values are the correct types 11 11 11 1h or -11 -11 -11 -1h"];
```
jonathonmcmurray: I think this assumes a specific order in the `sortparams` dictionary, which I don't think we need to enforce? Any ordering of keys should be fine

dataloader/util.q
```diff
@@ -0,0 +1,54 @@
...
+    if[not `tabname`att`column`sort~key sortparams; '"Error ensure argument is a dictionary with keys `tabname`att`column`sort"];
```
jonathonmcmurray: same comment as above line

dataloader/util.q
eliotrobinson: this solution doesnt work, not sure how to do this i tried a combination of namespaces but not really working as expected, will leave this for now and take a look on monday

dataloader/dataloader.q
```diff
@@ -1,22 +1,22 @@
...
 loaddata:{[loadparams;rawdata]
-  / loads data in from delimeted file, applies processing function, enumerates and writes to db
-  data:$[loadparams[`filename] in filesread; / check if some data has already been read in
-    flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata;
-    $[all (`$"," vs rawdata[0;]) in loadparams[`headers]; / it hasn't been seen, may be column headers
-      (loadparams[`types];loadparams[`separator])0:rawdata;
-      flip loadparams[`headers]!(loadparams[`types];loadparams[`separator])0:rawdata]
+  / loads data in from delimited file, applies processing function, enumerates and writes to db
+  / NOTE: it is not trivial to check user has inputted headers correctly, assume they have
+  data:$[(`$"," vs rawdata 0)~loadparams`headers;                                               / check if first row matches headers provided
```
jonathonmcmurray: as per https://github.com/DataIntellectTech/kdbx-packages/blob/main/style.md please place comments on preceding line

dataloader/dataloader.q
```diff
@@ -26,27 +26,27 @@ writedatapartition:{[dbdir;partition;partitiontype;partitioncol;tablename;data]
...
+  {.z.m.util.sorttab(x;where .z.m.partitions[;0]=x)}each distinct value .z.m.partitions[;0];
   system"x .z.zd";
-  if[loadparams`gc; .Q.gc[]];
+  if[loadparams`gc;.Q.gc[]];
   };
 
-loadallfiles:{[loadparams:.m.dataloader.util.paramfilter;dir]
+loadallfiles:{[loadparams:.z.m.util.paramfilter;dir]
   / load all the files from a specified directory
   .z.m.partitions::()!();
```
jonathonmcmurray:
```suggestion
  .z.m.partitions:()!();
```

dataloader/dataloader.q
```diff
@@ -26,27 +26,27 @@ writedatapartition:{[dbdir;partition;partitiontype;partitioncol;tablename;data]
...
+  .z.m.finish loadparams;
   };
 
-init:{[sortparams:.m.dataloader.util.sortfilter]
+init:{[sortparams:.z.m.util.sortfilter]
   / package init function
-  .z.m.partitions:()!(); / maintain a dictionary of the db partitions written to by loader
-  .z.m.filesread:(); / maintain a list of files which have been read
+  .z.m.partitions:()!();                                                                        / maintain a dictionary of the db partitions written to by loader
+  .z.m.filesread:();                                                                            / maintain a list of files which have been read
```
jonathonmcmurray: these are set in `loadallfiles`, do they need to be set here as well? seems like `loadallfiles` will overwrite these anyway, so probably not needed here?

possibly we don't actually need this `init` function at all and could just pass the sortparams directly to `loadallfiles`?

dataloader/util.q
jonathonmcmurray: yeah I think passing it as a param instead of using a global would be better here

dataloader/dataloader.q
jonathonmcmurray: unresolving this comment as comments are still in-line

jonathonmcmurray (approved):

PR #11, opened 2025-07-22
Add timezone package

jamiechandler99:
Tested : 
 - q 3.5 and up 
 - Utility scripts supports python 3.8 and above - must have requests installed but is the only external requirement

Notes: 
Unsure best way to have this data sourced. Python to me is the most lightweight and easily automatable but does not really fit the objective of a standalone q package. That being said one way or another this package will be reliant on external data. This to me seemed good to include, although maybe for actually shipping of this package its included more as an aside. 

Other options for this would be building out the HTTP q library and see if I can replicate this logic in q alone. Will take a look at that


jonathonmcmurray (commented):
We have a tzinfo file in TorQ (I think it's just a q flat file) https://github.com/DataIntellectTech/TorQ/blob/master/config/tzinfo (based off Kx supplied data at https://github.com/KxSystems/cookbook/tree/master/timezones)

Would it make sense to include that instead of the python script etc.? I guess we might have to update it periodically if countries change observance of DST etc. but that's probably pretty minimal

jamiechandler99:
Updated to take from flat and provided steps for manual update. Have also aligned code with style.md taking this to be the golden source. Will also update other PRs to align as I work on them

timezone/timezone.q
```diff
@@ -0,0 +1,34 @@
...
+/ read and format file for internal function reference
+.timezone.config.read:{
+  tz:get hsym `$.timezone.config.file;
+  tz:delete from tz where gmtDateTime>=10170056837;
+  tz:update gmtDateTime:12h$-946684800000000000+gmtDateTime*1000000000 from tz;
+  tz:update gmtOffset:16h$gmtOffset*1000000000 from tz;
+  tz:update localDateTime:gmtDateTime+gmtOffset from tz;
+  tz:`gmtDateTime xasc tz;
+  tz:update `g#timezoneID from tz;
+  tz};
```
jonathonmcmurray:
```suggestion
  tz
  };
```

(thought I had put this in style guide but I forgot, will update)

timezone/timezone.q
```diff
@@ -0,0 +1,34 @@
...
+
+/ convert from local timestamp to gmt
+.timezone.gmttolocal:{[tz;ts]
+  if[not (tz:`$tz) in .timezone.zones;'`notValidTimezone];
+  $[0>type ts;first;(::)]@exec gmtDateTime+gmtOffset from aj[`timezoneID`gmtDateTime;([]timezoneID:(),tz;gmtDateTime:ts,());.timezone.offsets]};
```
jonathonmcmurray:
```suggestion
  $[0>type ts;first;(::)]@exec gmtDateTime+gmtOffset from aj[`timezoneID`gmtDateTime;([]timezoneID:(),tz;gmtDateTime:ts,());.timezone.offsets]
  };
```

timezone/timezone.q
```diff
@@ -0,0 +1,34 @@
...
+
+/ convert from gmt to local timestamp
+.timezone.localtogmt:{[tz;ts]
+  if[not (tz:`$tz) in .timezone.zones;'`notValidTimezone];
+  $[0>type ts;first;(::)]@exec localDateTime-gmtOffset from aj[`timezoneID`localDateTime;([]timezoneID:(),tz;localDateTime:ts,());.timezone.offsets]};
```
jonathonmcmurray:
```suggestion
  $[0>type ts;first;(::)]@exec localDateTime-gmtOffset from aj[`timezoneID`localDateTime;([]timezoneID:(),tz;localDateTime:ts,());.timezone.offsets]
  };
```

timezone/test.csv
```diff
@@ -0,0 +1,11 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+before,0,0,q,.timezone.init[],1,,Initialize package
+true,0,0,q,0<count .timezone.offsets,1,,Confirm tzinfo is read in properly
+true,0,0,q,400<count .timezone.zones,1,,Confirm distinct time zones is populated and significant number available
+true,0,0,q,2025.07.22D14:19:48.386221575=.timezone.localtogmt["America/New_York";2025.07.22D10:19:48.386221575],1,,Test gmt to local 1
+true,0,0,q,1999.03.03D12:13:48.919241092=.timezone.localtogmt["Europe/London";1999.03.03D12:13:48.919241092],1,,Test gmt to local 2
+true,0,0,q,1967.05.02D21:03:52.857237462=.timezone.gmttolocal["America/Toronto";1967.05.03D01:03:52.857237462],1,,Test local to gmt 1
+true,0,0,q,2017.09.01D04:03:52.857237462=.timezone.gmttolocal["America/Los_Angeles";2017.09.01D11:03:52.857237462],1,,Test local to gmt 2
```
jonathonmcmurray: think the comments on these lines are the wrong way round?

timezone/timezone.md
```diff
@@ -0,0 +1,57 @@
+## KDB+/Q Timezone Conversion Library
+
+This project provides utilities to manage and convert timestamps across timezones in KDB+ using reference data from TimeZoneDB.
+
+---
+
+### TimeZoneDB Data Source
+
+Timezone reference data is sourced from https://timezonedb.com/download and must be downloaded and provided to the package in order to function.
```
jonathonmcmurray: I think we should add something here to say that there is a copy of `tzinfo` included (with date of generation) and mention that these steps are only needed if updating `tzinfo`

jonathonmcmurray (dismissed):
Looks good, couple of minor points & something I need to update in style guide

jamiechandler99:
Updated to reflect updated style guide for timezone ending brace, fixed unit test comments and added additional notes in doc that tzinfo is available by default and download steps are only needed for updating

jonathonmcmurray (approved):

PR #39, opened 2025-07-24
Adding k4unit testing framework

jamiechandler99:
Simple global framework for package unit testing

### DIFF
--- a/test/k4unit.q
+++ b/test/k4unit.q
@@ -1,0 +1,141 @@
+/ k4 unit testing, loads tests from csv's, runs+logs to database
+/ csv columns: action,ms,bytes,lang,code (csv with colheaders)
+/ if your code contains commas enclose the whole code in "quotes"
+/ usage: q k4unit.q -p 5001
+/ KUT <-> KUnit Tests
+KUT:([]action:`symbol$();ms:`int$();bytes:`long$();lang:`symbol$();code:`symbol$();repeat:`int$();minver:`float$();file:`symbol$();comment:())
+/ KUltd `:dirname and/or KUltf `:filename.csv
+/ KUrt[] / run tests
+/ KUTR <-> KUnit Test Results
+/ KUrtf`:filename.csv / refresh expected <ms> and <bytes> based on observed results in KUTR
+KUTR:([]action:`symbol$();ms:`int$();bytes:`long$();lang:`symbol$();code:`symbol$();repeat:`int$();file:`symbol$();msx:`int$();bytesx:`long$();ok:`boolean$();okms:`boolean$();okbytes:`boolean$();valid:`boolean$();timestamp:`datetime$())
+/ look at KUTR in browser or q session
+/ select from KUTR where not ok // KUerr
+/ select from KUTR where not okms // KUslow
+/ select count i by ok,okms,action from KUTR
+/ select count i by ok,okms,action,file from KUTR
+/ KUstr[] / save test results 
+/ KUltr[] / reload previously saved test results
+/ action:	
+/ 	`beforeany - onetime, run before any tests
+/		`beforeeach - run code before tests in every file
+/			`before - run code before tests in this file ONLY
+/			`run - run code, check execution time against ms
+/			`true - run code, check if returns true(1b)
+/			`fail - run code, it should fail (2+`two)
+/			`after - run code after tests in this file ONLY
+/		`aftereach - run code after tests in each file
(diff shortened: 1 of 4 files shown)


test/packagetest.q
```diff
@@ -0,0 +1,12 @@
+// Generic test script to be ran for individual packages
+
+\l test/k4unit.q
+
+.test.main:{
+    if[""~p:first (.Q.opt .z.x)`package;'`noPackageDefined];
```
jmlemay:
```suggestion
  if[""~p:first (.Q.opt .z.x)`package;'`noPackageDefined];
```
I like tabwidth 4 as well, but style guide mandates 2.

test/packagetest.q
```diff
@@ -0,0 +1,12 @@
+// Generic test script to be ran for individual packages
+
+\l test/k4unit.q
+
+.test.main:{
+    if[""~p:first (.Q.opt .z.x)`package;'`noPackageDefined];
+    $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'`noTestCsv];
+    system "l ",p,"/",p,".q";
+    KUrt[];
+    show `ok xasc KUTR;
```
jmlemay: I think it could cause some confusion if we play with the order of the tests when reporting the results. I agree with the intention to make the failures obvious, but I'd be tempted to report them separately.

```suggestion
  -1"Test results:";
  show KUTR;
  $[count failures:select from KUTR where not ok;
     [-1"Test failures:";show failures];
     -1"All tests passed"];
```

test/k4unit.q
```diff
@@ -0,0 +1,141 @@
...
+/ timestamp: when test was run
+/ comment: description of the test if it's obscure.. 
+
+\c 10000 10000 
+KUstr:{.KU.SAVEFILE 0:.KU.DELIM 0:update code:string code from KUTR} / save test results
+KUltr:{`KUTR upsert("SIJSSIJSIBBBBZ";enlist .KU.DELIM)0:.KU.SAVEFILE} / reload previously saved test results 
+
+KUltf:{ / (load test file) - load tests in csv file <x> into KUT
+	before:count KUT;
+	this:update file:x,action:lower action,lang:`q^lower lang,ms:0^ms,bytes:0j^bytes,repeat:1|repeat,minver:0^minver from `action`ms`bytes`lang`code`repeat`minver`comment xcol("SIJSSIF*";enlist .KU.DELIM)0:x:hsym x;
```
jonathonmcmurray: I think there is a bug in this version of k4unit where because `code` is loaded as a symbol, it's not possible to escape quotes within the test csv. We fixed that in the version of k4unit included in TorQ by loading as a string and then converting to symbol 

https://github.com/DataIntellectTech/TorQ/blob/master/tests/k4unit.q#L59
https://github.com/DataIntellectTech/TorQ/commit/f374cf5367dd0adbda57ad889c96eca136a64f01

test/packagetest.q
```diff
@@ -0,0 +1,16 @@
+/ generic test script to be ran for individual packages
+
+\l test/k4unit.q
+
+.test.main:{
+  if[""~p:first (.Q.opt .z.x)`package;'`noPackageDefined];
+  $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'`noTestCsv];
```
jonathonmcmurray:
```suggestion
  if[""~p:first (.Q.opt .z.x)`package;'"no package defined"];
  $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'"no test csv"];
```

test/packagetest.q
```diff
@@ -0,0 +1,16 @@
...
+  system "l ",p,"/",p,".q";
+  KUrt[];
+  -1"Test results:";
+  show KUTR;
+  $[count failures:select from KUTR where not ok;
+    -1"All tests passed";
+    [-1"Test failures:";show failures]];
```
jmlemay: My fault based on my original suggestion, but these are backwards.

test/k4unit.q
jamiechandler99: Have replaced logic in line 52 with TorQ version initially reading as string and then casting to sym

test/packagetest.q
jamiechandler99: Good catch, my bad should have taken a second to check. Updated

test/packagetest.q
jamiechandler99: Updated error message on break

jmlemay (approved):

PR #40, opened 2025-07-24
Add usage lib tests

jmlemay:
Notes:

- Cherry picked a commit into this branch from here: https://github.com/DataIntellectTech/kdbx-packages/pull/39, while that one waits to get merged.
- I made very minor changes to usage.q based on a couple of small issues I found while testing. The reason the diff looks so large is because I ran `dos2unix` on it after I noticed some `^M` characters floating about.


usage/test.q
```diff
@@ -0,0 +1,109 @@
...
+.testusg.exusage:flip`id`zcmd`status`a`u`w`cmd`sz`error!"jscisi*j*"$\:(); / same schema as `.usage.usage`
+.testusg.customtracker:enlist`zcmd`args!(`;""); / keeps track of invocation
+.testusg.wildcols:`time`extime`mem!12 16 0h; / `.usage.usage` columns we can't accurately predict
+
+/ runs a handler and keeps track of it internally so we can validate it against the usage table afterwards
+/ param handler {sym} handler, e.g. `.z.ps`
+/ param args    {any} args to handler
+.testusg.runhandler:{[handler;args]
+  .testusg.id+:1;
+  if[b:handler~`.z.ws;.test.mock[`.z.ps;{}]]; / swallow websocket response
```
jonathonmcmurray, line 26: is there a typo here? or is it intended that we mock .z.ps when handler is .z.ws?

usage/test.csv
```diff
@@ -0,0 +1,121 @@
...
+run,0,0,q,.testusg.runhandler[`.z.ps;(+;2;2)],1,1,
+run,0,0,q,.testusg.runhandler[`.z.ph;"dummy"],1,1,
+run,0,0,q,.testusg.runhandler[`.z.pp;"dummy"],1,1,
+run,0,0,q,.testusg.runhandler[`.z.exit;255],1,1,
+true,0,0,q,.testusg.chkusage[],1,1,
+true,0,0,q,([]zcmd:`.z.pw`.z.po`.z.pc`.z.wo`.z.wc`.z.ws`.z.pg`.z.ps`.z.ph`.z.pp`.z.exit;args:((`test_user;"pass123");1i;2i;3i;4i;-8!"1+1";"1+1";(+;2;2);"dummy";"dummy";255))~1_.testusg.customtracker,1,1,
+run,0,0,q,.testusg.reset[],1,1,
+
+run,0,0,q,.test.mock[`.usage.logtodisk;1b],1,1,should error on invalid logging configurations
+true,0,0,q,@[.usage.init;::;::]~"logName and logDir must be set to enable on disk usage logging. .usage.logToDisk disabled",1,1,
```
jonathonmcmurray: I guess we still have some camel case in our error messages somewhere, need to remove that (it looks like the actual variables are lower case, so doubly bad as error messages actually have wrong variable names)

usage/test.q
```diff
@@ -0,0 +1,109 @@
+/ load helper scripts and run tests.
+
+/ handlers we overwrite, columns are
+/ name    {sym} handler name
+/ typ     {sym} type of handler, corresponds to a function in the `.testusg` namespace
+/ default {fn}  default kdb implementation where `(::)` means there is none
```
jonathonmcmurray: is this for kx's qdoc? we haven't done this for any of the other packages (or even for the main script of this package). I think all the params etc. in this file are pretty self-explanatory so I'd lean towards removing these (otherwise if we keep them here I think we should add them everywhere)

usage/test.q
jmlemay: It's intended. If you run .z.ws with handle 0, then the default implementation `{neg[.z.w]x}` will trigger .z.ps on this process, which at best confuses the results when trying to track calls of the handlers, and at worse fails because `x` isn't in the right format. Path of least resistance was to just temporarily override .z.ps.

usage/test.csv
jmlemay: Good catch. I must have read this message 20 times and never noticed 😅

usage/test.q
jmlemay: It was my best attempt at hybridizing kx's qdoc style (which I've been using for the last 5 years) and this repo's style guide. I'm just in the habit of over-documenting everything 🤷

jonathonmcmurray (approved):

PR #43, opened 2025-07-28
Add os lib tests

jmlemay:
Note: one TODO in the tests, which is a bit of an open question. Currently, `.os.exists link` returns false if the link points to an invalid location. This seems a bit unintuitive to me since the link itself exists; it just doesn't point anywhere. Thus, I would recommend changing the behaviour to return true, but I don't know how the collective group feels about it.


jmlemay:
Merging since I'm taking over the os lib branch. Let's move any discussions there.

PR #51, opened 2025-08-26
Addition of k4unit testing package for KDB-X

KaelynSom:
Addition of 4kunit testing package:

Usage:
q)testing:use`k4unit

q)testing.packagetest"timezone"

Preqrequisites:

- Testing CSV inside the package: (timezone/test.csv)

- Inside the test csv, the before step must load the package: 
>before,0,0,q,timezone:use`timezone,1,,Initialize package
- The commands inside the test csv must be consistent with the naming of the package in the before step (timezone in above example):
 
>true,0,0,q,2025.07.22D14:19:48.386221575=timezone.localtogmt[`$"America/New_York";2025.07.22D10:19:48.386221575],1,,Test local to gmt 1

### DIFF
--- a/k4unit/init.q
+++ b/k4unit/init.q
@@ -1,20 +1,14 @@
 / generic test script to be ran for individual packages
 
 \l k4unit/k4unit.q
-\l k4unit/mock.q
 
 packagetest:{[p]
-  // check if package is loaded 
-  //if[not (`$p) in key`.z.m;
-    //'"package is not loaded"];
 
-
-  / Check if package passed either by function variable or in command line 
+  / Load the Test CSv for the assocaited package 
+  / First line of the package will load the package using the following format 
+  / package:use`package 
   $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'"no test csv"];
-  
-  
-  // Maintained logic of test helper script (not sure if we need to remove or not)
-  if[`test.q in key hsym`$p;system"l ",p,"/test.q"]; / Load any required helper code
+
   KUrt[];
   -1"Test results:";
   show KUTR;
--- a/k4unit/init.q
+++ b/k4unit/init.q
@@ -1,6 +1,6 @@
 / generic test script to be ran for individual packages
 
-\l k4unit/k4unit.q
+\l ::k4unit.q
 
 packagetest:{[p]
   / Load the Test CSv for the assocaited package 
--- a/k4unit/init.q
+++ b/k4unit/init.q
@@ -1,0 +1,27 @@
+/ generic test script to be ran for individual packages
+
+\l k4unit/k4unit.q
+\l k4unit/mock.q
+
+packagetest:{[p]
+  // check if package is loaded 
+  //if[not (`$p) in key`.z.m;
+    //'"package is not loaded"];
+
+
+  / Check if package passed either by function variable or in command line 
+  $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'"no test csv"];
+  
+  
+  // Maintained logic of test helper script (not sure if we need to remove or not)
+  if[`test.q in key hsym`$p;system"l ",p,"/test.q"]; / Load any required helper code
+  KUrt[];
+  -1"Test results:";
+  show KUTR;
+  $[count failures:select from KUTR where not ok;
+    [-1"Test failures:";show failures];
+    -1"All tests passed"];
+  };
+  
+
+export:([packagetest:packagetest])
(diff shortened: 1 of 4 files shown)


k4unit/init.q
```diff
@@ -0,0 +1,27 @@
+/ generic test script to be ran for individual packages
+
+\l k4unit/k4unit.q
+\l k4unit/mock.q
```
jonathonmcmurray: this file doesn't seem to exist in this PR @KaelynSom do you have this locally and could push it? Is it the same as the one currently in `test/` dir on main branch of repo?

k4unit/init.q
```diff
@@ -0,0 +1,27 @@
...
+
+packagetest:{[p]
+  // check if package is loaded 
+  //if[not (`$p) in key`.z.m;
+    //'"package is not loaded"];
+
```
jonathonmcmurray: assuming this should be removed?

k4unit/init.q
```diff
@@ -0,0 +1,27 @@
...
+
+  / Check if package passed either by function variable or in command line 
+  $[not ()~key tp:hsym `$p,"/test.csv";KUltf tp;'"no test csv"];
```
jonathonmcmurray: does this assume that package is in current directory? I don't think that is the right way to do this, we should probably be using `.Q.m.mp[p]` to get the module path (Although I think this will also load the module, which may or may not be intended here)

k4unit/k4unit.q
```diff
@@ -0,0 +1,139 @@
...
+/0 - trap errors, press on regardless
+/1 - suspend if errors (except if action=`fail of course)
+DEBUG:0
+/ DELIM, csv delimiter
+DELIM:","
+/ Test Results SAVEFILE
+SAVEFILE:`:KUTR.csv
+
+\d .
+@[value;"\\l k4unit.custom.q";::];
```
jonathonmcmurray: I'm not sure this bit makes sense in module world

k4unit/k4unit.q
```diff
@@ -0,0 +1,139 @@
...
+KUslow::delete okms from select from KUTR where not okms
+KUslowf::distinct exec file from KUslow
+KUbig::delete okbytes from select from KUTR where not okbytes
+KUbigf::distinct exec file from KUbig
+KUerr::delete ok from select from KUTR where not ok
+KUerrf::distinct exec file from KUerr
+KUinvalid::delete ok,valid from select from KUTR where not valid
+KUinvalidf::distinct exec file from KUinvalid
+
+\d .KU
```
jonathonmcmurray: don't think we should be using `\d` in modules

jonathonmcmurray (commented):
does this replace `test/` in the repo? if so, PR should probably remove that one as well as adding this one

k4unit/k4unit.q
```diff
@@ -0,0 +1,139 @@
...
+	if[action=`run;
+		failed:KUfailed r:KUpexec["\\ts ";lang;code;repeat;1b];msx:`int$$[failed;0;r 0];bytesx:`long$$[failed;0;r 1];
+		ok:not failed;okms:$[ms;not msx>ms;1b];okbytes:$[bytes;not bytesx>bytes;1b];valid:not failed];
+	if[action=`true;
+		failed:KUfailed r:KUpexec["";lang;code;repeat;1b];
+		ok:$[failed;0b;r~1b];okms:okbytes:1b;valid:not failed];
+	if[action=`fail;
+		failed:KUfailed r:KUpexec["";lang;code;repeat;0b];
+		ok:failed;okms:okbytes:valid:1b];
+	`.m.k4unit.KUTR insert(action;ms;bytes;lang;code;repeat;file;msx;bytesx;ok;okms;okbytes;valid;.z.Z);
```
jonathonmcmurray:
```suggestion
	.z.M.KUTR insert(action;ms;bytes;lang;code;repeat;file;msx;bytesx;ok;okms;okbytes;valid;.z.Z);
```

k4unit/init.q
```diff
@@ -0,0 +1,24 @@
+/ generic test script to be ran for individual packages
+
+\l k4unit/k4unit.q
```
jonnypress: Is this the best way to load in the packages environment? I thought there was a package specific load option?

jonnypress (dismissed):

k4unit/init.q
jonathonmcmurray: completely overlooked this, yeah I'll fix that now

jonnypress (approved):

PR #52, opened 2025-09-09
converted timer.q to init.q

paul-mcvicker:
Converted timer.q file to init.q file, with removal of function namespaces and exporting of functions


jonnypress (approved):
LGTM

PR #54, opened 2025-09-24
ps: port TorQ pubsub → KDB-X; adapt APIs; fix bugs; add k4unit tests

submartingle:
This module started from the TorQ pubsub implementation. I used it as the base and translated/adapted the code to the KDB-X package structure,  and fixed issues found during testing (see Fixes). Several components were redesigned—for example, `subfiltered `and `suball `now differ from the originals. The result keeps original behaviour where appropriate, aligns with KDB-X conventions, and adds missing guards and unit tests (k4unit).

- Fixes
     1. When subscribing via .ps.suball or .ps.subfiltered, the current handle must be removed from both registries before   logging a new subscription. The previous logic only removed it from the target table, leaving duplicates across “all” vs “filtered”, which could cause clients to receive both full-table and filtered streams simultaneously.
     2. In code ` val:![11 99h;(selfiltered;addfiltered)][type y] . (x;y);`  I think it should be using abs type y, since users may provide a single symbol (atom) as the filter. 

- Tests 
         run  `q test/packagetest.q -package ps.`     All tests passed.

- TODO
    1. Add more edge testing cases.
    2. Confirm whether data striping is in scope for this module. I don't believe so, any thoughts welcome.

- Files
ps/ps.q – module
ps/test.csv – k4unit tests


ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+
+.ps.delhandle:{[t;h]
+     / remove handle from request-all-data table
+     if[t in key .ps.reqalldict; @[`.ps.reqalldict;t;except;h]]
+     if[not count .ps.reqalldict[t];.ps.reqalldict _:t];  
+      } 
+
+/ remove handle from request-filtered-data table
+.ps.delhandlef:{[t;h]
+     delete from `.ps.reqfilteredtbl where table=t, handle=h;}
```
jonathonmcmurray:
```suggestion
.ps.delhandlef:{[t;h]delete from `.ps.reqfilteredtbl where table=t, handle=h;}  
```

would make this a one-line function

ps/ps.q
```diff
@@ -0,0 +1,162 @@
+/ subscription table - no filters
+.ps.reqalldict:enlist[`]!();
+
+/ subscription table with filters
+.ps.reqfilteredtbl:([]table:`symbol$();handle:`int$();filts:();columns:());
+
+
+/ get all subscription handles that haven been recorded on tables
+.ps.getallhandles:{:distinct raze union[value .ps.reqalldict;exec handle from .ps.reqfilteredtbl]}
```
jonathonmcmurray:
```suggestion
.ps.getallhandles:{distinct raze union[value .ps.reqalldict;exec handle from .ps.reqfilteredtbl]}
```
would skip the explicit return for a one liner

ps/ps.q
```diff
@@ -0,0 +1,162 @@
+/ subscription table - no filters
+.ps.reqalldict:enlist[`]!();
+
+/ subscription table with filters
+.ps.reqfilteredtbl:([]table:`symbol$();handle:`int$();filts:();columns:());
+
+
```
jonathonmcmurray: excessive blank lines can be removed

ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+    m:(); x,:(); 
+    if[not all x in .ps.t; 
+         errmsg:(`$sv[csv;string  m:x except .ps.t]," not available for subscription.");
+         x@:where x in .ps.t]; 
+    $[count x; 
+     [{.ps.delhandle[x;.z.w];
+       .ps.delhandlef[x;.z.w];
+       .ps.add[x]} each x;
+      :((errmsg;(x;.ps.schemas[x]));(x;.ps.schemas[x])) [m~()]];
+        :errmsg;]
```
jonathonmcmurray: Ideally we should avoid using the block syntax within $ conditionals (it's messy & hard to follow), and we can avoid it here by just using an `if` given you're already using an explicit return. Similar applies to .ps.subfiltered.

```suggestion
    if[count x; 
     {.ps.delhandle[x;.z.w];
       .ps.delhandlef[x;.z.w];
       .ps.add[x]} each x;
      :((errmsg;(x;.ps.schemas[x]));(x;.ps.schemas[x])) [m~()];
      ];
     :errmsg;
```

ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+     if[t in key .ps.reqalldict; @[`.ps.reqalldict;t;except;h]]
+     if[not count .ps.reqalldict[t];.ps.reqalldict _:t];  
+      } 
+
+/ remove handle from request-filtered-data table
+.ps.delhandlef:{[t;h]
+     delete from `.ps.reqfilteredtbl where table=t, handle=h;}  
+
+
+.ps.suball:{
```
jonathonmcmurray: can you add explicit param(s) here please?

ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+    $[count x; 
+     [{.ps.delhandle[x;.z.w];
+       .ps.delhandlef[x;.z.w];
+       .ps.add[x]} each x;
+      :((errmsg;(x;.ps.schemas[x]));(x;.ps.schemas[x])) [m~()]];
+        :errmsg;]
+   }
+     
+
+.ps.subfiltered:{[x;y]
```
jonathonmcmurray: can we have more descriptive param names please? (applies to next couple of functions as well)

ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+/by default get all tables on top level of STP
+.ps.availtables:1b;  
+
+/user-specified list of tables for subscription
+.ps.subtables:`$();  
+
+.ps.initialized:0b;
+
+.ps.init:{
+    $[.ps.availtables | not count .ps.subtables;
```
jonathonmcmurray:
```suggestion
    $[.ps.availtables |0=count .ps.subtables;  
```

IMO `0=` is clearer in intention than `not` when dealing with a count

ps/ps.q
```diff
@@ -0,0 +1,162 @@
...
+.ps.availtables:1b;  
+
+/user-specified list of tables for subscription
+.ps.subtables:`$();  
+
+.ps.initialized:0b;
+
+.ps.init:{
+    $[.ps.availtables | not count .ps.subtables;  
+       .ps.t:tables `.; .ps.t:.ps.subtables];
```
jonathonmcmurray: $ returns a value, so can assign outside e.g.

```suggestion
    .ps.t:$[.ps.availtables |0=count .ps.subtables;tables `.;.ps.subtables];
```

jonathonmcmurray (commented):
Looks good, a few minor comments on various things there but overall looks nice. Thanks!

ps/ps.q
submartingle: understood, noted.

ps/ps.q
submartingle: updated

ps/ps.q
submartingle: updated in 9c6abd0, thanks,

ps/ps.q
submartingle: updated in 9c6abd0, thanks,

ps/ps.q
submartingle: updated in 9c6abd0, thanks,

submartingle:
> Looks good, a few minor comments on various things there but overall looks nice. Thanks!

Thanks—I've applied the comments in the latest commit; ready for another pass when convenient.

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+.z.m.reqfilteredtbl:([]table:`symbol$();handle:`int$();filts:();columns:());
+
+/ get all subscription handles that haven been recorded on tables
+getallhandles:{distinct raze union[value .z.m.reqalldict;exec handle from .z.m.reqfilteredtbl]};
+
+/ add handle to reqalldict dictionary
+add:{[t] if[not .z.w in .z.m.reqalldict[t];.z.m.reqalldict[t],:.z.w]};
+
+delhandle:{[t;h]
+    / remove handle from request-all-data table
```
jonathonmcmurray: can we convert indentation from 4 spaces to 2 spaces please for consistency with other packages & the style guide https://github.com/DataIntellectTech/kdbx-packages/blob/main/style.md

ps/init.q
```diff
@@ -0,0 +1,138 @@
+/ subscription table - no filters
+.z.m.reqalldict:enlist[`]!();
+
+/ subscription table with filters
+.z.m.reqfilteredtbl:([]table:`symbol$();handle:`int$();filts:();columns:());
```
jonathonmcmurray: don't think you need to specify `.z.m` on these


```suggestion
reqalldict:enlist[`]!();

/ subscription table with filters
reqfilteredtbl:([]table:`symbol$();handle:`int$();filts:();columns:());
```

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+add:{[t] if[not .z.w in .z.m.reqalldict[t];.z.m.reqalldict[t],:.z.w]};
+
+delhandle:{[t;h]
+    / remove handle from request-all-data table
+    if[t in key .z.m.reqalldict; @[`.m.ps.reqalldict;t;except;h]]
+    if[not count .z.m.reqalldict[t];.m.ps.reqalldict _:t];
+    };
+
+/ remove handle from request-filtered-data table
+delhandlef:{[t;h]delete from `.m.ps.reqfilteredtbl where table=t, handle=h};
```
jonathonmcmurray:
```suggestion
delhandlef:{[t;h]delete from .z.M.reqfilteredtbl where table=t, handle=h};
```

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+/ get all subscription handles that haven been recorded on tables
+getallhandles:{distinct raze union[value .z.m.reqalldict;exec handle from .z.m.reqfilteredtbl]};
+
+/ add handle to reqalldict dictionary
+add:{[t] if[not .z.w in .z.m.reqalldict[t];.z.m.reqalldict[t],:.z.w]};
+
+delhandle:{[t;h]
+    / remove handle from request-all-data table
+    if[t in key .z.m.reqalldict; @[`.m.ps.reqalldict;t;except;h]]
+    if[not count .z.m.reqalldict[t];.m.ps.reqalldict _:t];
```
jonathonmcmurray:
```suggestion
    if[t in key .z.m.reqalldict; @[.z.M.reqalldict;t;except;h]]
    if[not count .z.m.reqalldict[t];.z.m.reqalldict _:t];
```

should be using `.z.M` and `.z.m` rather than referring to namespace directly

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+endp:{(neg getallhandles[])@\:`endofperiod`};
+
+/ get table schema
+extractschema:{[table]0#value table};
+
+subscribe:{[table;filters]
+    / single entry point for subscriptions: uses default list when no table name provided; routes to suball if filters null, otherwise subfiltered
+    if[`~table;:.z.s[.z.m.t;filters]];
+    if[not`~table;
+        :$[`~filters;suball[table];subfiltered[table;filters]]]
```
jonathonmcmurray:
```suggestion
        :$[`~filters;suball;subfiltered[;filters]]table]
```
🏌️

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+
+/ broadcast to all subscribers upon end of period, client needs to define endofperiod function
+endp:{(neg getallhandles[])@\:`endofperiod`};
+
+/ get table schema
+extractschema:{[table]0#value table};
+
+subscribe:{[table;filters]
+    / single entry point for subscriptions: uses default list when no table name provided; routes to suball if filters null, otherwise subfiltered
+    if[`~table;:.z.s[.z.m.t;filters]];
```
jonathonmcmurray: would it make more sense to just do


```suggestion
    if[`~table;table:.z.m.t];
```
 
and then can also remove the second conditional as we'll never get to that point with table being null symbol

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+
+substrf:{[table;filters;columns]
+    / allow non-kdb+ process to subscribe to tables with custom conditions
+    res:subscribe[`$table;1!enlist `table`filts`columns!(`$table;filters;columns)];
+    :$[10h~type last res;'last res;res];
+    };
+
+/ set .u functions in case they get called
+.u.sub:subscribe;
+.u.pub:publish;
```
jonathonmcmurray: @jonnypress do we want to be doing this within modules? my thought is that modules shouldn't have side-effects like this in other namespaces (I know in some we're setting things in `.z` but that seems unavoidable)

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+
+/by default get all tables on top level of STP
+.z.m.availtables:1b;
+
+/user-specified list of tables for subscription
+.z.m.subtables:`$();
+
+.z.m.initialized:0b;
```
jonathonmcmurray:
```suggestion
/by default get all tables on top level of STP
availtables:1b;

/user-specified list of tables for subscription
subtables:`$();

initialized:0b;
```

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+
+/ set .u functions in case they get called
+.u.sub:subscribe;
+.u.pub:publish;
+
+/by default get all tables on top level of STP
+.z.m.availtables:1b;
+
+/user-specified list of tables for subscription
+.z.m.subtables:`$();
```
jonathonmcmurray: how does this get set by user?

ps/init.q
```diff
@@ -0,0 +1,138 @@
...
+.z.m.initialized:0b;
+
+init:{
+    .z.m.t:$[.z.m.availtables |0=count .z.m.subtables;tables[]except`reqfilteredtbl;.z.m.subtables];
+    .z.m.schemas:.z.m.t!extractschema each .z.m.t;
+    .z.m.tabcols:.z.m.t!cols each .z.m.t;
+    if[count .z.m.tabcols;@[.z.M;`initialized;:;1b]];
+    };
+
+export:.z.m
```
jonathonmcmurray: should we be exporting everything from the module? seems like some things are probably internal details that user doesn't need to use/care about. should only be exporting the functions/variables that endusers are going to access (e.g. we probably don't need to export extractschema, but we do want to export subscribe etc.)

ps/ps.md
```diff
@@ -0,0 +1,100 @@
+[ps.md](https://github.com/user-attachments/files/22662173/ps.md)
+# Publish/Subscribe Utilities
+> **Note:** Initial documentation draft (to be refined post-review).
+
+
+A lightweight kdbx-compatible package provides a list of functions to help users to subscribe to STP with custom data requirements, and publish data from the STP to downstream accordingly.
```
jonathonmcmurray: this module is standalone, so I don't think STP makes sense in this context - users of this module probably aren't using TorQ and may not be familiar with TorQ. I think it's fine to reference tickerplant (would use full name, maybe with a link to kx's tick repo, rather than abbreviation), but I think also we should highlight that the pubsub functionality will work in any given process, it doesn't _have_ to be a tickerplant

ps/test.csv
```diff
@@ -1,84 +1,114 @@
 action,ms,bytes,lang,code,repeat,minver,comment
 comment,,,,,,,Setup for server and client testing
 comment,,,,,,,Replace $QHOME (defaults to $HOME) to avoid hard-coding
-beforeany,0,0,q,system "QHOME=${QHOME:-$HOME}; q ${QHOME}/kdbx-packages/ps/ps.q -p 5010 -q &",1,,start server in background
-beforeany,0,0,q,system "sleep 1",1,,wait for server to initialize
-beforeany,0,0,q,h:hopen `::5010,1,,open handle h on client for testing
+beforeany,0,0,q,system"q ${HOME}/kdbx-packages/ps/init.q -p 5010 -q &",1,,start server in background
```
jonathonmcmurray: I think we want to be using `use` rather than explicitly loading the filepath - seems like we use `use` below, so probably just need to remove filepath from this line

ps/init.q
eliotrobinson: Maybe this is a placeholder for tests? maybe we can have a simple function that the user can call?
```
{$[x~`;0#x;x]}`
`symbol$()
{$[x~`;0#x;x]}`trade`quote
`trade`quote
```

ps/init.q
eliotrobinson: A lot of the functions ended up being available to be used publicly as defined in the ps.md descriptions, I only removed 2 from the export. Perhaps you could take a look and see if there's anymore we could remove?

pubsub/pubsub.q
```diff
@@ -11,65 +11,65 @@ getallhandles:{distinct raze union[value .z.m.reqalldict;exec handle from .z.m.r
 add:{[t] if[not .z.w in .z.m.reqalldict[t];.z.m.reqalldict[t],:.z.w]};
 
 delhandle:{[t;h]
-    / remove handle from request-all-data table
-    if[t in key .z.m.reqalldict; @[`.m.ps.reqalldict;t;except;h]]
-    if[not count .z.m.reqalldict[t];.m.ps.reqalldict _:t];
-    };
+  / remove handle from request-all-data table
+  if[t in key .z.m.reqalldict; @[.z.M.reqalldict;t;except;h]];
```
submartingle, line 15: @eliotrobinson @jonathonmcmurray ,not sure about .z.m namespace here (and the .z.M below), are they defined anywhere as I cannot find it in the kx ref? is it OK to point me to the def module? will add a brief README note and unit test. Thx.

ps/init.q
```diff
@@ -0,0 +1,136 @@
...
+initialized:0b;
+
+init:{
+  .z.m.t:$[count .z.m.subtables;.z.m.subtables;tables[]except`reqfilteredtbl];
+  .z.m.schemas:.z.m.t!extractschema each .z.m.t;
+  .z.m.tabcols:.z.m.t!cols each .z.m.t;
+  if[count .z.m.tabcols;@[.z.M;`initialized;:;1b]];
+  };
+
+export:`availtables`initialized _ .z.m
```
submartingle: Proposal: export only the four API key functions—subscribe, substr, substrf, publish—plus a few direct-call helpers (setsubtables, endp, endd, closesub, pubclear). Everything else stays internal and is invoked via those key functions.

ps/init.q
```diff
@@ -0,0 +1,136 @@
...
+endp:{(neg getallhandles[])@\:`endofperiod`};
+
+/ get table schema
+extractschema:{[table]0#value table};
+
+subscribe:{[table;filters]
+  / single entry point for subscriptions: uses default list when no table name provided; routes to suball if filters null, otherwise subfiltered
+  if[`~table;table:.z.m.t];
+  if[not`~table;
+    :$[`~filters;suball;subfiltered[;filters]]table];
```
submartingle: suggested change to remove the if conditional: 
`:$[`~filters;suball;subfiltered[;filters]]table;`

submartingle (commented):
Left 3 inline clarifications (.z.M/m clarification, minor code update for the subscribe function, API export list).

ps/init.q
jonnypress: Agreed. This shouldn't happen like this, it should be up to the module user e.g. 

pubsub:use`ps
.u.sub:pubsub.subscribe
etc.

ps/init.q
jonnypress: Yes, that seems correct to me.

jonnypress (commented):
I think this is an example where we should definitely separate the code and the init.q. So put the main body of the code in a separate file and keep the init.q clean (loads and exports). 

Also, can we make the naming conventions more clear (this is an opportunity to do better than we did before!). The module should be called something like pubsub or publishsubscribe rather than ps. I don't find the exposed function names of substr, substrf, endp, endd particularly intuitive. I'm ok with all the rest of it, let's just make the interface easy.

ps/ps.md
```diff
@@ -0,0 +1,97 @@
...
+| `ps.delhandle`     | Remove a subscriber handle from the request-all registry.                    |
+| `ps.delhandlef`    | Remove a subscriber handle from the request-filtered registry.               |
+| `ps.endd`          | Broadcast an end-of-day event to all subscribers (requires `endofday`).      |
+| `ps.endp`          | Broadcast an end-of-period event to all subscribers (requires `endofperiod`).|
+| `ps.extractschema` | Return a table’s schema.                                                     |
+| `ps.suball`        | Subscribe to all tables.                                                     |
+| `ps.subfiltered`   | Subscribe to tables with filters.                                            |
+| `ps.addfiltered`   | Subscribe to tables with custom conditions.                                  |
+| `ps.addsymsub`     | Subscribe to tables with a list of symbols (or one symbol).                  |
+| `ps.closesub`      | Remove handle upon connection close.                                         |
```
jonnypress: I would remove a bunch of these if they are internal details (not exposed)

pubsub/pubsub.q
eliotrobinson: https://docs.kx.com/public-preview/kdb-x/integrations/module-management.htm
https://docs.kx.com/public-preview/kdb-x/integrations/module-management-qsg.htm
these pages should explain the gist of `.z.m/M` as well as using `use` and `export`.

pubsub/pubsub.q
```diff
@@ -0,0 +1,129 @@
...
+    {if[count filtered:eval(?;y;z`filts;0b;z`columns);neg[z`handle](`upd;x;filtered)]}[t;x;] each d];
+  };
+
+pubclear:{[t]
+  / publish tables and clear up the contents
+  publish'[t;value each t,:()];
+  @[`.;;0#] each t;
+  };
+
+subtable:{[table;syms]
```
submartingle: just thinking for the two functions using string to subscribe for non-kdb processes, would it be better to use names "subscribestr" and "subscribestrfilter"- to avoid confusion given "subtables" is already used for the list of tables for subscription, so
     subtable->subscribestr
     subtablecustom->subscribestrfilter
let me know, I can update the code/docs if agreed.

submartingle (commented):
Left one naming suggestion: subtable→subscribestr, subtablecustom→subscribestrfilter. OK to merge as-is if you prefer.

pubsub/pubsub.md
```diff
@@ -0,0 +1,82 @@
...
+```q
+publish data with/without filters. The function takes two arguments: t and x, which are table name and data to be published respectively. when no table specified, the function will scan both reqalldict and reqfilteredtbl, when table name is found/subscribed, publish to downstream subscribers accordingly.
+```
+
+---
+
+### Utilities:
+| Function                 | Description                                                                  |
+|--------------------------|------------------------------------------------------------------------------|
+| `pubsub.setsubtables`    | Set a specified list of tables tthat are available for subscription.         |
```
jonathonmcmurray:
```suggestion
| `pubsub.setsubtables`    | Set a specified list of tables that are available for subscription.         | 
```

typo

pubsub/pubsub.q
jonathonmcmurray: these suggested names sound reasonable to me 👍

jonathonmcmurray (approved):

PR #57, opened 2025-09-30
Tidy usage package so it conforms to packaging standards

matt-proudley:
Usage package re formatted to conform to package standards


usage/init.q
```diff
@@ -0,0 +1,8 @@
+//Load core functionality into root module namespace
+\l ::usage.q
+
+//Return to root module namespace to simplify exposure of public functions
+\d .z.m
```
jonathonmcmurray: don't think this is necessary? don't think we ever left root namespace

usage/usage.q
```diff
@@ -3,7 +3,7 @@
 / handlers with usage-logging wrappers of their current definitions
 
 / table to store usage info
-.usage.usage:@[value;`.usage.usage;([]
+usage:@[value;`.z.m.usage;([]
```
jonathonmcmurray: I don't think this style of defaulting makes sense in module-world. I don't think one can update the variables within the module namespace before loading the module, so these will always just get the default values.

I think the best way to handle this is probably passing a dictionary to `.usage.init` - we can use the existing defaults for anything not passed to `.usage.init`, and we should allow calling `.usage.init[]` to use all default settings.

usage/init.q
```diff
@@ -0,0 +1,8 @@
...
+export:([init:init; getusage:getusage; 
+    readlog:readlog; flushusage:flushusage; 
+    meminfo:meminfo; formatarg:formatarg])
```
jonathonmcmurray: Minor style point, but I think we should either give every item it's own line, or put everything on one line (I'd lean towards the former)

```suggestion
export:([
    init:init;
    getusage:getusage; 
    readlog:readlog;
    flushusage:flushusage; 
    meminfo:meminfo;
    formatarg:formatarg
    ])
```

jonathonmcmurray: also, do we need to export all of these? would an enduser need to call e.g. `formatarg` or `meminfo` directly? we should only be exporting functions we expect endusers to call

usage/usage.q
```diff
@@ -177,31 +177,35 @@
...
+      (.z.m.logdir;.z.m.logname;.z.m.logtimestamp[.z.m.localtime]);
+      {.z.m.logtodisk:0b;'"Error creating log file: ",.z.m.logdir,"/usage_",.z.m.logname,"_",string[.z.m.logtimestamp[.z.m.localtime]]," | Error: " ,x}]];
   };
 
-.usage.init:{[]
-  .usage.inithandlers[];
-  .usage.initlog[];
+/ exportable function to get usage table
+getusage:{[] 
+  :.z.m.usage };
```
jonathonmcmurray:
```suggestion
getusage:{[].z.m.usage};
```

usage/test.csv
```diff
@@ -1,121 +1,158 @@
...
-run,0,0,q,.usage.usage:0#.usage.usage,1,1,
-run,0,0,q,.test.unmock[],1,1,
+true,0,0,q,([]zcmd:`ps;status:6#"bc";cmd:(.Q.s1(+;1;2);.Q.s1(`fn;`trade;([]x:1 2));"1+0")where 3#2)~`zcmd`status`cmd#usage.getusage[],1,1,
+run,0,0,q,.m.usage.usage:0#usage.getusage[],1,1, default value
+run,0,0,q,.m.usage.ignore:0b,1,1, default value
+run,0,0,q,.m.usage.ignorelist:(),1,1, default value
 
-run,0,0,q,".test.mock[`.usage.ext;{.testusg.ext,:enlist x}]",1,1,test extension
-run,0,0,q,.z.pw[`test_user;"pass123"],1,1,
+run,0,0,q,".m.usage.ext:{.test.ext,:enlist x}",1,1,test extension functionality
```
jonathonmcmurray: I don't think this is correct usage - end users shouldn't be modifing things in `.m.usage` - we should be providing a setter function for setting `ext` something like

```
setext:{[x]ext::x}
```

which would be exported and endusers could then call e.g.

```q
usage.setext {.test.ext,:enlist x}
```

to set the ext function

(should proably have better names being used here, but hopefully this illustrates the point anyway)

usage/usage.q
matt-proudley: Sorry not entirely sure what you mean here. Please could you add some detail on what the change might look like / what should the dictionary passed to usage.init[] might look like.

jonathonmcmurray: sure. so I think actually the `usage` table isn't really config so it wouldn't be one of these, but using a few of the others as examples, I think it'd be something like this:

```q
init:{[x]
  def:([logtodisk:0b;logtomemory:1b;ignore:1b;ignorelist:();...)]; / config defaults
  x:$[x~(::);def;def,x]; / if init[] called, just use defaults; if dict passed, overwrite defaults
  key[x] set' value[x];  / set passed values to config variables
  / ... do rest of init stuff ...
 }
```

so then calling init could be like e.g. with default config settings

```q
usage.init[]
```

or with some overwrites

```q
usage.init `logtodisk`logtomemory!10b
```

etc.

Does that make sense?

matt-proudley: got it, so any config values that are set as defaults (like logging to disk) that aren't exportable values can be set using args passed to usage.init if the user chooses to enable them? This as apposed to the current approach of adjusting them inside the usage.q file.

jonathonmcmurray: yeah, users should never be editing the usage.q file really

matt-proudley: have included this as a feature, might need to go a step further and add some checks on the incoming dict, otherwise user can assign any var inside the usage namespace, let me know your thoughts

usage/usage.q
```diff
@@ -177,31 +154,46 @@
...
+  .z.m.logtodisk:0b;    / whether to log to disk
+  .z.m.logtomemory:1b;  / whether to log to memory
+  .z.m.logdir:"";       / should be set before loading library to initialise on disk logging
+  .z.m.logname:"";      / log file will take the form "usage_{logname}_{date/time}.log"
+  .z.m.ignore:1b;       / whether to check the ignore list for function calls to not log
+  .z.m.ignorelist:();   / list of function to not log usage of
+  .z.m.level:3;         / log level,	0 = nothing, 1 = errors only, 2 = + open, close, queries, 3 = + log queries before execution
+  .z.m.localtime:1b;    / check time preference. Default local time
+
+  if[(not x~(::)); (.Q.dd[.z.M] each key[x]) set' value[x]];
```
jonathonmcmurray:
```suggestion
  if[not x~(::); (.Q.dd[.z.M] each key[x]) set' value[x]];
```

but we should also probably limit this to just the "allowed" config variables e.g. something like

```suggestion
  if[not x~(::);
    vars:`logtodisk`logtomemory`logdir... inter key x;
    (.Q.dd[.z.M] each key[vars#x]) set' value[vars#x]
  ];
```

usage/usage.q
```diff
@@ -17,153 +17,130 @@
...
-/ should be set before loading library to initialise on disk logging
-.usage.logname:@[value;`.usage.logname;""];
-
-/ log level
-/	0 = nothing
-/	1 = errors only
-/	2 = + open, close, queries
-/	3 = + log queries before execution
-.usage.level:@[value;`.usage.level;3];
+logtimestamp:@[value;`.z.m.logtimestamp;{{[local] $[local;.z.D;.z.d]}}];
```
jonathonmcmurray:
```suggestion
logtimestamp:{[local] $[local;.z.D;.z.d]};
```
don't think we need the defaulting stuff here

usage/usage.q
```diff
@@ -17,153 +17,130 @@
...
 
 / ID for tracking external queries
-.usage.id:@[value;`.usage.id;0];
+id:@[value;`.z.m.id;0];
```
jonathonmcmurray:
```suggestion
id:0;
```
users shouldn't be reloading module (and indeed I think it's difficult to actually do that in kdb-x anyway), so there shouldn't be a pre-existing value. so we can just set this to 0 directly

usage/usage.q
jonathonmcmurray:
```suggestion
usage:([]
```

don't need the defaulting thing, that is only relevant is `usage` table is defined in the modules namespace before loading the module, which doesn't make sense

jonathonmcmurray (commented):
couple of minor tweaks

jonathonmcmurray (approved):

PR #58, opened 2025-10-17
Converted os.q to init.q

eliotrobinson:
remove the os namespace to use the `use` keyword instead


os/init.q
```diff
@@ -0,0 +1,169 @@
...
+  };
+
+/ toggles dry system calls on (1b) or off (0b)
+setdrysyscalls:{[on]
+  .z.m.syscall:$[on;drysyscall;system]
+  };
+
+cleardrysyscalls[]
+
+export:([topath:topath;abspath:abspath;realpath:realpath;exists:exists;isfile:isfile;isdir:isdir;issymlink:issymlink;del:del;deldir:deldir;mkdir:mkdir;mv:mv;cp:cp;cpdir:cpdir;kill:kill;kill2:kill2;kill3:kill3;kill9:kill9;sleep:sleep;cd:cd;chmod:chmod;chown:chown;pwd:pwd;createsymlink:createsymlink;mkfifo:mkfifo;mktemp:mktemp;mktempdir:mktempdir;getdrysyscalls:getdrysyscalls;cleardrysyscalls:cleardrysyscalls;setdrysyscalls:setdrysyscalls])
```
jonathonmcmurray: when we're exporting a large number of functions, it might make sense to do this by exclusion instead?

```q
export:`syscall`drysyscall _ .z.m
```

or something like that should work I think

os/test.q
```diff
@@ -1,10 +1,13 @@
-.testos.sep:$[.os.iswindows;"\\";"/"];
+os:use`os;
+k4unit:use`k4unit;
+k4unit.debug 1b;
```
jonathonmcmurray: I don't think these lines should be needed here? the person running the tests will `use` k4unit in their q session & then call ``k4unit.packagetest`os``, and the test CSV will `use` the os package, so it shouldn't be necessary to `use` either inside this script? and we should leave the choice of whether or not use debug mode up to the person running the tests as well

os/test.q
```diff
@@ -60,3 +63,34 @@
...
+  };
+
+/ internal - deletes an object from the namespace it belongs to
+.testos.deletefromns:{[obj]
+  if[obj like".z.*";:system"x ",string obj]; / Special .z callbacks
+  split:` vs obj;
+  k:last obj;
+  ns:$[1=count split;`.;` sv -1_split];
+  ![ns;();0b;enlist k];
+  }
```
jonathonmcmurray: would it make more sense to put this in the k4unit package so it can be used by tests in other packages going forward?

os/test.q
eliotrobinson: was a relic leftover from testing this, removed it

jonathonmcmurray (approved):

PR #59, opened 2025-10-20
changes to exported functions in timer package and updated documentation

matt-proudley:


timer/init.q
```diff
@@ -2,7 +2,7 @@
 
 / override variables to change logic
 debug:0b; / if enabled displays messages for jobs starting and any errors in execution
-logcall:1b; / if enabled will execute timer functions throuhgh 0 handle
+logcall:1b; / if enabled will execute timer functions through 0 handle
 cycletime:100; / frequency to check for new jobs to start (in ms)
 cp:{.z.p}; / return timestamp function used to evaluate start times, can be overwritten for backtesting and simulating
```
jonathonmcmurray: I think we should pass these as dict to `init` function, similar to in the usage module

timer/init.q
```diff
@@ -97,7 +97,9 @@ enablejobs:{[ids] ids:(),ids;.m.timer.jobs:update status:1b from .m.timer.jobs w
 disablejobs:{[ids] ids:(),ids;.m.timer.jobs:update status:0b from .m.timer.jobs where id in ids};
 deletejobs:{[ids] ids:(),ids;.m.timer.jobs:delete from .m.timer.jobs where id in ids};
 
-getactive:{select from .m.timer.jobs where status};
+getactivejobs:{select from .m.timer.jobs where status};
```
jonathonmcmurray:
```suggestion
getactivejobs:{select from .z.m.jobs where status};
```

timer/init.q
```diff
@@ -138,5 +140,29 @@ init:{
   if[not system"t";system "t ",string cycletime];
   };
 
+setlogcall:{.z.m.logcall: x}
+
+setcycletime:{.z.m.cycletime: x}
+
+setcp:{.z.m.cp: x}
+
+setdebug:{.z.m.debug: x}
```
jonathonmcmurray: is it expected to change these after initialisation? if not, I think above suggestion to pass as dict to init would also mean removing these setter functions

matt-proudley: From testing so far, the cp var must be set before init is called as it is used when adding a job. The others seem to have no issues being set in init. Will keep testing.

timer/init.q
```diff
@@ -160,9 +161,6 @@ export:([
   disablejobs:disablejobs;
   enable:enable;
   disable:disable;
-  / set config variables
-  setlogcall:setlogcall;
-  setcycletime:setcycletime;
-  setcp:setcp;
-  setdebug:setdebug
+  / setter for cp variable
+  setcp:setcp
```
jonathonmcmurray, line 165: isn't this function removed above?

matt-proudley: I removed this function and realised that the variable needs to be set when adding jobs/before running init. So I added it back.

jonathonmcmurray (approved):

PR #64, opened 2025-10-28
fix(q): expose init[] to ensure state initialization

submartingle:
Follow-up to #54. `init` was not exposed, leaving required vars (e.g. `.z.m.t/schemas`) uninitialized/undefined for the main functions.

Tests did not catch this because `test.csv` accesses both public and private symbols.

This PR exposes `init` to ensure state initialization.


jonathonmcmurray (approved):
Thanks for this, approved

PR #65, opened 2025-10-29
Move dataloader to di directory

eliotrobinson:


di/dataloader/dataloader.md
jonathonmcmurray, line 1: need to update example here to ``use`di.dataloader``

jonathonmcmurray (approved):

PR #66, opened 2025-10-29
Update tests to allow for test reruns

eliotrobinson:
Currently the tests work on the first pass, but rerunning in the same process sometimes leads to fails or the code erroring. I've made some changes here to clean up the tests to allow for rerunning in the same process without errors. 3 packages were affected:

 - **pubsub**: this was erroring on subsequent runs, I determined this is because the ports being opened were being ran as a `beforeany` which had a weird interaction on second runs. changing this to just `run` seems to allow all the tests to pass again, but with the caveat that I get the two error outputs whilst running:
 ```
'2025.10.29T14:18:05.504 5010. OS reports: Address already in use
'2025.10.29T14:18:06.521 5010. OS reports: Address already in use
```
 - **timer**: this package was relatively simple to clean up, the `jobs`table just needed emptying before being ran again. This was easy enough to accomplish by just utilising `timer.deletejobs` to empty any ids.
 - **usage**: Similarly to timer, this was simple enough to fix by adding some rows to reset some variables or empty some tables

### DIFF
--- a/di/pubsub/test.csv
+++ b/di/pubsub/test.csv
@@ -1,13 +1,13 @@
 action,ms,bytes,lang,code,repeat,minver,comment
 comment,,,,,,,Setup for server and client testing
 comment,,,,,,,Replace $QHOME (defaults to $HOME) to avoid hard-coding
-beforeany,0,0,q,system"q -p 5010 -q &",1,,start server in background
-beforeany,0,0,q,system"sleep 1",1,,wait for server to initialize
-beforeany,0,0,q,h:hopen`::5010,1,,open handle h on client for testing
-beforeany,0,0,q,h"pubsub:use`di.pubsub",,initialize pubsub package
+run,0,0,q,system"q -p 5010 -q &",1,,start server in background
+run,0,0,q,system"sleep 1",1,,wait for server to initialize
+run,0,0,q,h:hopen`::5010,1,,open handle h on client for testing
+run,0,0,q,h"pubsub:use`di.pubsub",,initialize pubsub package
 true,0,0,q,h>0,1,,verify client connected to server
-beforeany,0,0,q,h"trade:([]time:\"p\"$();`g#sym:`$();price:\"f\"$();tsize:\"i\"$());",1,,define trade table on server
-beforeany,0,0,q,h"quote:([]time:\"p\"$();`g#sym:`$();bid:\"f\"$();bidsize:\"i\"$();ask:\"f\"$();asksize:\"i\"$());",1,,define quote table on server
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: di/pubsub/test.csv [hunk 1]
```diff
@@ -112,3 +112,6 @@ true,0,0,q,98h~type first res[1],1,,client receives table schema
 run,0,0,q,res:h"select from .m.di.0pubsub.reqfilteredtbl where table=`quote",1,,check filtered table for the subscription
 run,0,0,q,res:select from res where handle=clt,1,,check filtered table for the subscription
 true,0,0,q,res[`filts][0]~enlist enlist(>;`bid;50f),1,, test symbols are listed under filters
+
+run,0,0,q,neg[h](exit;0),1,,close background process
```
jonathonmcmurray: do we need a flush here to make sure this gets run?


```suggestion
run,0,0,q,neg[h](exit;0);neg[h](::),1,,close background process
```

jonathonmcmurray (dismissed):

jonathonmcmurray (approved):

di/pubsub/test.csv
```diff
@@ -1,13 +1,13 @@
...
+run,0,0,q,system"q -p 5010 -q &",1,,start server in background
+run,0,0,q,system"sleep 1",1,,wait for server to initialize
+run,0,0,q,h:hopen`::5010,1,,open handle h on client for testing
+run,0,0,q,h"pubsub:use`di.pubsub",,initialize pubsub package
 true,0,0,q,h>0,1,,verify client connected to server
-beforeany,0,0,q,h"trade:([]time:\"p\"$();`g#sym:`$();price:\"f\"$();tsize:\"i\"$());",1,,define trade table on server
-beforeany,0,0,q,h"quote:([]time:\"p\"$();`g#sym:`$();bid:\"f\"$();bidsize:\"i\"$();ask:\"f\"$();asksize:\"i\"$());",1,,define quote table on server
+run,0,0,q,h"trade:([]time:\"p\"$();`g#sym:`$();price:\"f\"$();tsize:\"i\"$());",1,,define trade table on server
+run,0,0,q,h"quote:([]time:\"p\"$();`g#sym:`$();bid:\"f\"$();bidsize:\"i\"$();ask:\"f\"$();asksize:\"i\"$());",1,,define quote table on server
 run,0,0,q,h".m.di.0pubsub.init[]",,
```
submartingle, line 11: @eliotrobinson @jonathonmcmurray   
I’ve been working on a new module (di.analytics) and trying to reuse the testing pattern from di.pubsub, just one question-- what is the reason to have 0 prefix function names, e.g. .m.di.0pubsub ?
I only dug into this while wiring up the analytics tests, which is why I’m asking now instead of when the original pubsub PR was merged.   Can you explain the convention here please. Thanks.

PR #71, opened 2025-10-31
Make improvements on back of KX suggestions

eliotrobinson:
A couple of changes here:
 - Changed the method used to export modules, basically clear up a lot of unnecessary characters
 - Changed how the util submodule is loaded for dataloader
 - removed *most* of the `.z.m/M` references, this was a bit tricky and there are some which simply breaks the module if I remove them (looking at you usage...)
 - Remove whitespace when I notice it

All tests pass for each module, and let me know if there are any more `.z.m/M` references I've missed.

### DIFF
--- a/di/usage/usage.q
+++ b/di/usage/usage.q
@@ -24,26 +24,26 @@ logtimestamp:{[local] $[local;.z.D;.z.d]};
 id:0;
 
 / increment ID and return new value
-nextid:{[] :.z.m.id+:1;};
+nextid:{[]:.z.m.id+:1};
 
 / return local time or UTC
-currenttime:{[local] $[local;.z.P;.z.p]};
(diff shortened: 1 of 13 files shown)


di/usage/usage.q
```diff
@@ -24,26 +24,26 @@ logtimestamp:{[local] $[local;.z.D;.z.d]};
 id:0;
 
 / increment ID and return new value
-nextid:{[] :.z.m.id+:1;};
+nextid:{[]:.z.m.id+:1};
```
daviddemnerDI, line 27: not sure which one Kx would prefer, .z.m or global amend like `nextid:{[]id+::1};`

jonathonmcmurray (approved):

PR #73, opened 2025-11-21
Add async module

davidjedw:
Utility functions to send deferred sync and async postback IPC messages with error trapping at various points.
Also includes function variants to send these via async broadcast.

### DIFF
--- a/di/async/async.md
+++ b/di/async/async.md
@@ -18,24 +18,6 @@ If either of these are carried out via asynchronous broadcast, the request will 
 
 ---
 
-### Package Initialization
-
-Loading the module will automatically initialise using the included async module.
-
-```q
-q)async:use`async
-```
-
-If you wish to use an alternative async file, you can call the init function with
-the path to your file
-
-```q
-q)async:use`async
-q)async.init "path/to/async"
-```
-
----
-
 ### Package Use
 
 Note, in each of the examples below handles is a list of two handles to different server processes
@@ -50,12 +32,12 @@ q)async.deferred[handles;"2+2"]
 4 4
 ```
 
-##### async.broadcast_deferred
+##### async.broadcastdeferred
 As above, except the query will be sent via asynchronous broadcast. 
 Note, that if there is an issue with any of the handles, the query won't be sent down any handle
 ```q
-// async.broadcast_deferred[handles;query]
-q)async.broadcast_deferred[handles;"2+2"]
+// async.broadcastdeferred[handles;query]
+q)async.broadcastdeferred[handles;"2+2"]
 1 1
 4 4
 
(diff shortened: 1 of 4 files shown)


di/async/async.q
```diff
@@ -0,0 +1,38 @@
...
+  res:{$[y;@[x;(::);(0b;"error: comm fail: handle closed while waiting for result")];(0b;"error: comm fail: failed to send query")]}'[abs handles;sent];
+  / return results
+  (res[;0];res[;1])}
+
+
+postback:{[handles;query;postback] 
+  / for sending asynchronous postback message to a list of handles where the message is wrapped in the function postback
+  send[0b;;({[q;p] (p;@[value;q;{"error: server fail: ",x}])};query;postback)] each handles:neg abs handles,()}
+
+broadcastdeferred:{[handles;query]
```
jonathonmcmurray: is there ever a reason _not_ to use broadcast? i.e. should we just get rid of the non-broadcast versions and rename the broadcast ones?

davidjedw: Only scenario I can think of is if there's an issue with any of the handles - with broadcast, one invalid handle causes none of the messages to be sent, whereas for the non-broadcast versions only the message to the invalid handle will fail. Bit of a rare case though, and ideally some connection management code would prevent that situation in the first place. So to tidy things up, it's probably not worth keeping the non-broadcast versions in that case

di/async/async.q
```diff
@@ -0,0 +1,25 @@
+/ library for sending async messages from a client process
+
+send:{[w;h;q]
+  / build the query to send
+  tosend:$[w; ({[q] @[neg .z.w;@[{[q] (1b;value q)};q;{(0b;"error: server fail:",x)}];()]};q);
+              ({[q] @[neg .z.w;@[{[q] value q};q;{"error: server fail:",x}];()]};q)];
+  / error trapping sending the query down the handle followed by an async flush
+  .[{x@y; x(::);1b};(h;tosend);0b]}
+
+broadcastdeferred:{[handles;query]
```
jonathonmcmurray: rename this to just `deferred` seeing as we don't have a non-broadcast version any more

di/async/init.q
```diff
@@ -0,0 +1,3 @@
+\l ::async.q
+
+export:([deferred;postback;broadcastdeferred;broadcastpostback])
```
jonathonmcmurray: this needs updated

di/async/test.csv
jonathonmcmurray, line 1: tests need updated, they still reference both broadcast & non-broadcast versions

di/async/async.md
jonathonmcmurray, line 1: docs need updated, still reference both broadcast & non-broadcast versions

di/async/async.md
```diff
@@ -0,0 +1,69 @@
+## KDB+/Q Asynchronous Communication Library
+
+This library provides kdb+/q functions for sending either deferred synchronous or asynchronous postback requests from a client  process over a handle or list of handles, with error trapping at various points. Either type of request can be sent via conventional kdb+/q IPC or asynchronous broadcast.
```
jonathonmcmurray:
```suggestion
This library provides kdb+/q functions for sending either deferred synchronous or asynchronous postback requests from a client process over a handle or list of handles, with error trapping at various points. Either type of request can be sent via conventional kdb+/q IPC or asynchronous broadcast.
```

jonathonmcmurray: (also last sentence needs updated)

jonathonmcmurray (changes requested):
few additional things needed here

di/async/async.q
```diff
@@ -0,0 +1,25 @@
+/ library for sending async messages from a client process
+
+send:{[w;h;q]
```
davidjedw: `send` function can be removed also as only relates to non-broadcast versions

jonathonmcmurray (approved):
LGTM

PR #75, opened 2025-12-19
Add mockDataGenerator Module

NikhilBiradar2000:
Need a module to allow us to build up reasonably realistic mock data sets.
We have two samples to work from

[feed.q in TorQ FSP](https://github.com/DataIntellectTech/TorQ-Finance-Starter-Pack/blob/master/code/tick/feed.q), which is the code for the dummy data feedhandler.

fakedb.q from our training content
feed.q has a more realistic way to generate a price path than fakedb.q, though fakedb.q allows generation of additional datasets and works in batch, which is what we need (rather than tick-by-tick).

See https://github.com/DataIntellectTech/kdbx-modules/discussions/74  for more

### DIFF
--- a/di/mockDataGen/init.q
+++ b/di/mockDataGen/init.q
@@ -1,0 +1,3 @@
+\l ::mockDataGen.q
+
+export:([initschema;mockDataOne;mockData;mockHdb;mockDataR;clearTables])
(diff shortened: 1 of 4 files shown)


D1I8 (approved):
Looks good to me.

PR #76, opened 2025-12-22
Adding di.cache module

boyce-pj:
Commiting code from Adam Beck and Lewis Cooley

### DIFF
--- a/di/cache/cache.q
+++ b/di/cache/cache.q
@@ -9,9 +9,16 @@ maxsize:10;
 / the maximum size of any individual result set in MB
 maxindividual:50;
 
+/ function to change default max individual value
+setmaxindiv:{.z.m.maxindividual:x}
+
+/ function to change default max size value
+setmaxsize:{.z.m.maxsize:x}
+
 / make sure the maxindividual isn't bigger than maxsize
 maxindividual:maxsize&maxindividual;
 
+/ mb conversion factor
 MB:2 xexp 20;
 
 / a table to store the cache values in memory
@@ -31,22 +38,26 @@ getid:{:id+::1};
 
 / add to cache
 add:{[function;id;status]
-  / Don't trap the error here - if it throws an error, we want it to be propagated out
-    res:value function;
-  $[(maxindividual*MB)>size:-22!res;
-    / check if we need more space to store this item
-    [now:cp[];
-    if[0>requiredsize:(maxsize*MB) - size+sum exec size from cache; evict[neg requiredsize;now]];
-    / Insert to the cache table
-    .z.M.cache upsert (id;now;now;size);
-    / and insert to the function and results dictionary
-    funcs[id]:enlist function;
-    results[id]:enlist res;
-    / Update the performance
-    trackperf[id;status;now]];
-    / Otherwise just log it as an addfail - the result set is too big
-    trackperf[id;`fail;cp[]]];
-  / Return the result
+  / don't trap the error here - if it throws an error, we want it to be propagated out
+  res:value function;
+  if[(maxindividual*MB)<=size:-22!res;
+    / log it as an addfail - the result set is too big
+    trackperf[id;`fail;cp[]];
+    :res;
+  ];
+  / check if we need more space to store this item
+  now:cp[];
+  if[0>requiredsize:(maxsize*MB) - size+sum exec size from cache;
+    evict[neg requiredsize;now];
+  ];
+  / insert to the cache table
+  .z.M.cache upsert (id;now;now;size);
+  / and insert to the function and results dictionary
+  funcs[id]:enlist function;
+  results[id]:enlist res;
+  / update the performance
+  trackperf[id;status;now];;
+  / return the result
   res};
 
 // Drop some ids from the cache
@@ -56,9 +67,9 @@ drop:{[ids]
   results:: ids _ results;
  };
 
-// evict some items from the cache - need to clear enough space for the new item
-// evict the least recently accessed items which make up the total size
-// feel free to write a more intelligent cache eviction policy !
+/ evict some items from the cache - need to clear enough space for the new item
+/ evict the least recently accessed items which make up the total size
+/ feel free to write a more intelligent cache eviction policy !
 evict:{[reqsize;currenttime]
   r:select 
     from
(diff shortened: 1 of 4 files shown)


di/cache/init.q
```diff
@@ -0,0 +1,9 @@
+/ Load core functionality into root module namespace
+\l ::cache.q
+
+export:([
+        getperf:getperf;
+        sdd:add;
+        drop:drop;
+        execute:execute
+        ])
```
jonathonmcmurray:
```suggestion
export:([getperf;add;drop;execute])
```

assuming `sdd` was a typo

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
+/ Library to provide a mechanism for storing function results in a cache and returning them from the cache if they are available and non stale.
+
+/ return timestamp function
+cp:{.z.p};
```
jonathonmcmurray: there doesn't appear to be a way to change this for end-users, so might make more sense to just use `.z.p` directly instaed of having this function?

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+    / Update the performance
+    trackperf[id;status;now]];
+    / Otherwise just log it as an addfail - the result set is too big
+    trackperf[id;`fail;cp[]]];
+  / Return the result
+  res};
+
+// Drop some ids from the cache
+drop:{[ids]
+        ids,:();
```
jonathonmcmurray: inconsistent indentation between different functions in this file, align to style guide please

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+perf:([]time:`timestamp$();id:`long$();status:`symbol$());
+
+id:0j;
+getid:{:id+::1};
+
+/ add to cache
+add:{[function;id;status]
+  / Don't trap the error here - if it throws an error, we want it to be propagated out
+    res:value function;
+  $[(maxindividual*MB)>size:-22!res;
```
jonathonmcmurray: inconsistent indentation

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+cp:.z.p;
+
+/ the maximum size of the cache in MB
+maxsize:10;
+
+/ the maximum size of any individual result set in MB
+maxindividual:50;
+
+/ make sure the maxindividual isn't bigger than maxsize
+maxindividual:maxsize&maxindividual;
```
jonathonmcmurray, line 16: we should have a function for users to change these config settings

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+
+MB:2 xexp 20;
```
jonathonmcmurray, line 19: add a comment to explain this

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+    funcs[id]:enlist function;
+    results[id]:enlist res;
+    / Update the performance
+    trackperf[id;status;now]];
+    / Otherwise just log it as an addfail - the result set is too big
+    trackperf[id;`fail;cp[]]];
+  / Return the result
+  res};
+
+// Drop some ids from the cache
```
jonathonmcmurray: comments from here onwards use `//` instead of `/`

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+
+/ add to cache
+add:{[function;id;status]
+  / Don't trap the error here - if it throws an error, we want it to be propagated out
+    res:value function;
+  $[(maxindividual*MB)>size:-22!res;
+    / check if we need more space to store this item
+    [now:cp[];
+    if[0>requiredsize:(maxsize*MB) - size+sum exec size from cache; evict[neg requiredsize;now]];
+    / Insert to the cache table
```
jonathonmcmurray: comments should start with lowercase as per style guide

di/cache/cache.q
```diff
@@ -0,0 +1,94 @@
...
+    .z.M.cache upsert (id;now;now;size);
+    / and insert to the function and results dictionary
+    funcs[id]:enlist function;
+    results[id]:enlist res;
+    / Update the performance
+    trackperf[id;status;now]];
+    / Otherwise just log it as an addfail - the result set is too big
+    trackperf[id;`fail;cp[]]];
```
jonathonmcmurray: I think it would be worth refactoring this to avoid the code block inside $ https://github.com/DataIntellectTech/kdbx-modules/blob/main/style.md#conditionals-and-execution-control

e.g. something like this, reversing the condition & returning early if the result is too large

```suggestion
  if[(maxindividual*MB)<=size:-22!res;
     / log it as an addfail - the result set is too big
     trackperf[id;`fail;cp[]];
     :res;
   ];
  / check if we need more space to store this item
  now:cp[];
  if[0>requiredsize:(maxsize*MB) - size+sum exec size from cache;
    evict[neg requiredsize;now];
  ];
  / Insert to the cache table
  .z.M.cache upsert (id;now;now;size);
  / and insert to the function and results dictionary
  funcs[id]:enlist function;
  results[id]:enlist res;
  / Update the performance
  trackperf[id;status;now]];
```

di/cache/init.q
```diff
@@ -0,0 +1,4 @@
+/ Load core functionality into root module namespace
+\l ::cache.q
+
+export:([getperf;add;drop;execute])
```
jonathonmcmurray: I don't think `add` or `drop` should be exposed to end users, I don't think we'd expect them to be using those functions directly

di/cache/cache.q
```diff
@@ -0,0 +1,105 @@
...
+maxsize:10;
+
+/ the maximum size of any individual result set in MB
+maxindividual:50;
+
+/ function to change default max individual value
+setmaxindiv:{.z.m.maxindividual:x}
+
+/ function to change default max size value
+setmaxsize:{.z.m.maxsize:x}
```
jonathonmcmurray: we need to do similar to line 19 in both of these functions to make sure maxindividual is never larger than maxsize

di/cache/cache.q
jonathonmcmurray: rest of file has been fixed but this line still has `//`

di/cache/test.csv
jonathonmcmurray, line 1: we should probably have tests for setmaxinidiv and setmaxsize, including validating that the max sizes are respected by `execute`

di/cache/cache.q
benlord77: If we are setting maxsize and maxindividual here and then line 19 runs after, would that not do this already?

jonathonmcmurray: during first load, yes, but the point of these functions is that user can set the variables _after_ inital load, and line 19 won't run again in those instances

benlord77: Ah I see, will update that now.

di/cache/cache.q
jonathonmcmurray: I don't think this is right - this will always be the time when the module was loaded, not the current time as intended. I think we can get rid of this variable and just use `.z.p` directly where this is used

di/cache/cache.q
```diff
@@ -40,27 +40,27 @@ getid:{:id+::1};
...
+     :res;
+   ];
   / check if we need more space to store this item
-  now:cp[];
+  now:cp;
   if[0>requiredsize:(maxsize*MB) - size+sum exec size from cache;
     evict[neg requiredsize;now];
   ];
-  / insert to the cache table
+  / Insert to the cache table
```
jonathonmcmurray:
```suggestion
  / insert to the cache table
```

lower case for comments as per style guide

di/cache/cache.q
```diff
@@ -40,27 +40,27 @@ getid:{:id+::1};
...
   .z.M.cache upsert (id;now;now;size);
   / and insert to the function and results dictionary
   funcs[id]:enlist function;
   results[id]:enlist res;
-  / update the performance
-  trackperf[id;status;now];;
+  / Update the performance
```
jonathonmcmurray:
```suggestion
  / update the performance
```

di/cache/test.csv
```diff
@@ -1,6 +1,14 @@
 action,ms,bytes,lang,code,repeat,minver,comment
 before,0,0,q,cac:use`di.cache,1,1,load package into session
 
+/ Test 0: max variable config checks
+run,0,0,q,maxindividual:10,1,, 
+run,0,0,q,maxsize:10,1,,
+run,0,0,q,cac.setmaxindiv[200],1,, run setmaxindiv
+run,0,0,q,cac.setmaxsize[100],1,, run setmaxsize 
+true,0,0,q,100~ maxindividual,1,, Check maxindividual configed correctly by function
+true,0,0,q,100~ maxsize,1,, Check maxindividual configed correctly by function
```
jonathonmcmurray, line 10: this is OK, but we should make sure that when we actually call execute, if result is too large it won't get cached, and that things get evicted from cache as necessary etc.

benlord77: I see the difference there, will update now.

jonathonmcmurray (approved):
LGTM

PR #82, opened 2026-01-05
Adding compression functions to new compression module

boyce-pj:
Dependent on KX log module

### DIFF
--- a/di/compression/compression.md
+++ b/di/compression/compression.md
@@ -40,11 +40,13 @@ loadcsv or showcomp functions accept `:/path/to/csv argument to specify driver c
 
 ## :label: Example Config
 
+```
 table,minage,column,calgo,cblocksize,clevel
 default,10,default, 2, 17,6
 quotes, 10,time, 2, 17, 5
 quotes,10,src,2,17,4
 depth, 10,default, 1, 17, 8
+```
 
 - tables in the db but not in the config tab are automatically compressed using default params
 - tabs with cols specified will have other columns compressed with default (if default specified for cols of tab, all cols are comp in that tab)
--- a/di/compression/compression.md
+++ b/di/compression/compression.md
@@ -71,7 +71,7 @@ depth, 10,default, 1, 17, 8
 
 ```q
 // Include compression module in a process
-.cmp: use `di.compression
+cmp:use`di.compression
 
 // View dictionary of functions
 .cmp
--- a/di/compression/compression.md
+++ b/di/compression/compression.md
@@ -1,0 +1,89 @@
+# Compression
+
+`compression.q` applies compression to any kdb+ database, handles all partition types including date, month, year, int, and can deal with top level splayed tables. It will also decompress files as required.
+
+> **Note:** Please use with caution
+
+---
+
+## :sparkles: Features
+
+- Apply compression to any kdb+ database
+- Compression uses -19! operator
+- Configured using specified csv driver file
+- Compression algorithm (none, kdb+ IPC, gzip), compression blocksize, and compression level (for gzip) can be configured
+- Flexibility to compress different tables/columns with different compression parameters
+- Preview available before compression to display files to be compressed and how
+- Summary statistics for each file returned after compression/decompression complete
+
+---
+
+## memo: Dependencies
+
+- KX log module
+
+---
+
+## :gear: compressioncsv Schema
+
+Blank compressioncsv is created when module is loaded. 
+loadcsv or showcomp functions accept `:/path/to/csv argument to specify driver csv location and will load the config to compressioncsv
+
+| Column     | Type    | Description                                            |
+|------------|---------|--------------------------------------------------------|
+| table      | `symbol`| table names for specific compression                   |
+| minage     | `int`   | minimum age of file in days before compression applied |
+| column     | `symbol`| column name for column specific compression            |
+| calgo      | `int`   | compression algorithim - 0, 1, or 2 accepted           |
+| cblocksize | `int`   | compression blocksize - between 12 and 19              |
+| clevel     | `int`   | compression level - 0 to 9 - for gzip (2) only         |
+
+## :label: Example Config
+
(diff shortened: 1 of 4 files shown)


di/compression/compression.q
```diff
@@ -0,0 +1,220 @@
...
+        .log.error[err:path,": Compression config has incorrect compression level at row(s): ", (" " sv string checklevel), ". Should be between 0 and 9."];'err];
+    if[.z.o like "w*"; if[count rowwin:where ((csvtab[`cblocksize] < 16) & csvtab[`calgo] > 0);
+        .log.error[err:path,": Compression config has incorrect compression blocksize for windows at row: ", (" " sv string rowwin), ". Must be more than or equal to 16."];'err]];
+    if[(any nulls: any null (csvtab[`column];csvtab[`table];csvtab[`minage];csvtab[`clevel]))>0;
+        .log.error[err:path,": Compression config has empty cells in column(s): ", (" " sv string `column`table`minage`clevel where nulls)];'err];
+  }
+
+loadcsv:{[inputcsv]
+    .cmp.compressioncsv::@[{.log.info["Opening ", x];("SISIII"; enlist ",") 0:"S"$x}; (string inputcsv); {.log.error["failed to open ", (x)," : ",y];'y}[string inputcsv]];
+    .[checkcsv;(.cmp.compressioncsv;string inputcsv);{.log.error["failed to load csv due to error: ",x];delete compressioncsv from `.cmp}];
```
boyce-pj: Global assignment here wasn't playing nice so I needed to explicitly define a global variable for this to be used elsewhere. A similar thing happened with .cmp.statstab
I'm not sure why ```compressioncsv:: ...``` was not able to define the global within the context of the module namespace.
For example if I load this module using:
```.cmp:use`di.compression```
all the functions define in the .cmp namespace but I needed a fully qualified global assignment within the functions themselves otherwise these data seemed to just drop...

File: di/compression/init.q [hunk 2]
```diff
@@ -0,0 +1,25 @@
+/ Load core functionality into root module namespace
+\l ::compression.q
+
+/ Load KX log module
+.logger:use`kx.log
+.log:.logger.createLog[]
+.log.addfmt[`custom;"$l: $p PID[$i] $m\n"]; / Simplified logging - DELETE
```
jonathonmcmurray: comment implies this should be deleted?

di/compression/compression.md
jonathonmcmurray, line 1: need documentation added here

di/compression/compression.q
```diff
@@ -0,0 +1,220 @@
...
+default,10,default,2,17,6
+
+The gzip algo (2) is not necessarily included on windows and unix systems.
+See: code.kx.com/wiki/Cookbook/FileCompression for more details
+
+For WINDOWS users:
+
+The minimum block size for compression on windows is 16.
+
+\
```
jonathonmcmurray: this should all be in documentation, not comment in script

di/compression/compression.q
```diff
@@ -0,0 +1,220 @@
...
+
+/inputcsv:@[value;`inputcsv;first .proc.getconfigfile["compressionconfig.csv"]];
+
+/if[-11h=type inputcsv;inputcsv:string inputcsv];
```
jonathonmcmurray: if these aren't needed, remove rather than leaving them here commented

di/compression/compression.q
jonathonmcmurray, line 1: this file should be aligned to the style guide https://github.com/DataIntellectTech/kdbx-modules/blob/main/style.md

di/compression/init.q
```diff
@@ -0,0 +1,25 @@
...
+        statstabupdate:statstabupdate;
+        singlethreadcompress:singlethreadcompress;
+        multithreadcompress:multithreadcompress;
+        compressmaxage:compressmaxage;
+        docompression:docompression;
+        summarystats:summarystats;
+        compress:compress;
+        cleancompressed:cleancompressed;
+        hashfilecheck:hashfilecheck
+        ])
```
jonathonmcmurray: don't need to assign each of these, can just do `export:([checkcsv;loadcsv;hdbstructure...])`. Also, do we need to be exporting all of these? Pretty sure a lot of these are internal functions that we wouldn't expect end users to call, in which case they shouldn't be exported.

jonathonmcmurray (changes requested):
make sure we align to the style & consistency guides https://github.com/DataIntellectTech/kdbx-modules/blob/main/consistency.md

di/compression/compression.md
```diff
@@ -0,0 +1,91 @@
...
+|`docompression`     | Compress all files in hdb according to compressioncsv config   |
+|`getstatstab`       | get function to retrieve summary stats post compression        |
+
+---
+
+## :test_tube: Example
+
+```q
+// Include compression module in a process
+.cmp: use `di.compression
```
jonathonmcmurray: I think we should make the example

```suggestion
cmp:use`di.compression
```

I think this is more aligned with how we have examples written for other modules (and also removing unnecessary spaces)

di/compression/init.q
```diff
@@ -0,0 +1,8 @@
+/ Load core functionality into root module namespace
+\l ::compression.q
+
+/ Load KX log module - needed for .log.info and .log.error
+.logger:use`kx.log
+.log:.logger.createLog[]
```
jonathonmcmurray: I think we should do these things locally within module namespace rather than globally (e.g. in case user already has a `.log` or `.logger` namespace defined, we don't want to overwrite them when they load this module)


```suggestion
/ Load KX log module - needed for .log.info and .log.error
logger:use`kx.log
log:logger.createLog[]
```

I think this should work the same (will obviously need to update any lines where we do logging to use `log` instead of `.log`) - can you check that works the same?

jonathonmcmurray (commented):
couple of minor points, otherwise think this looks good

jonathonmcmurray (approved):

PR #86, opened 2026-03-11
Add simtick

PhilSing24:
This PR updates the existing di.simtick module in preparation for the upcoming di.simcalendar (multi-day simulation) and di.simbasket (multi-instrument simulation) modules, which will depend on simtick.

Changes:
- Added sym column to trade and quote tables, required by upstream modules
- Recalibrated presets to NVDA (default), volatile, and jumpy scenarios
- Updated unit test
- Updated README with corrected parameter defaults and preset descriptions


PhilSing24:
di.simtick — updates

Price rounding: trades and quotes rounded to nearest cent
Seed sentinel changed from 0 to 0N (null long = no seed)
Fixed RNG dispatch: all normal draws now go through .z.m.rng.normal
Fixed spread timestamps: pre-trade quotes now use correct timestamps
Added validation for vol and startprice (must be positive)
Unit tests: 46 → 50
Notebook: added quote/trade consistency plot

PR #90, opened 2026-04-16
New basic logging module using kx.log as reference

DI-Dexter:
# Log

`log.q` is the default logging implementation for `di.*` modules. It writes formatted lines to stdout and satisfies the log dependency contract expected by modules such as `di.email`.

## Usage

```q
log:use`di.log
log.trace[`mymodule;"entering function"]
log.debug[`mymodule;"value is 42"]
log.info[`mymodule;"starting up"]
log.warn[`mymodule;"disk usage above 80%"]
log.error[`mymodule;"connection failed"]
log.fatal[`mymodule;"unrecoverable error, shutting down"]
```

Output format:

```
2026-04-09T12:00:00.000000000 [TRACE] [mymodule] entering function
2026-04-09T12:00:00.001000000 [DEBUG] [mymodule] value is 42
2026-04-09T12:00:00.002000000 [INFO] [mymodule] starting up
2026-04-09T12:00:00.003000000 [WARN] [mymodule] disk usage above 80%
2026-04-09T12:00:00.004000000 [ERROR] [mymodule] connection failed
2026-04-09T12:00:00.005000000 [FATAL] [mymodule] unrecoverable error, shutting down
```

## Injecting into other modules

All `di.*` modules that accept a log dependency expect a dictionary with keys `` `info`warn`error ``, each a function with signature `{[ctx;msg]}`.

```q
log:use`di.log

email:use`di.email
email.init[emailconfig;`log`send!(logdep;::)]
```

You can extend the injected dictionary with `trace`, `debug`, and `fatal` for modules that support them:

```q
logdep:`trace`debug`info`warn`error`fatal!(log.trace;log.debug;log.info;log.warn;log.error;log.fatal)
```

## createLog

`createLog` is a factory that returns an independent logger instance with level filtering, multiple output sinks, and configurable format templates. Each call to `createLog` produces a separate instance with its own state.

```q
log:use`di.log
mylog:log.createLog[]

mylog.setlvl `warn                        / suppress trace, debug, info
mylog.info "this is suppressed"           / returns () silently
mylog.warn "this appears"

mylog.setfmt `syslog                      / switch to syslog format
mylog.addfmt[`compact;"$l $m"]           / add a custom format
mylog.setfmt `compact

mylog.add[2i;`error`fatal]               / add stderr sink for error and fatal
mylog.remove[1i;`trace]                  / remove stdout from trace level
```

### Sinks

A sink is a handle (integer file descriptor or function) passed to `add`. If a function is provided it is called with the formatted line string. Built-in handles follow standard q conventions: `1i` is stdout, `2i` is stderr.

```q
buf:();
capture:{[msg] buf,:enlist msg};         / function sink - captures output
mylog.add[capture;`info`warn`error]
mylog.info "captured"
buf                                      / ("2026-...captured\n")
mylog.remove[capture;`info]
```

### Format templates

Three built-in formats are available:

| Name | Template | Example output |
|---|---|---|
| `basic` (default) | `$p $l PID[$i] HOST[$h] $m` | `2026-04-09T12:00:00.000000000 INFO PID[1234] HOST[myhost] message` |
| `syslog` | `<$s> $m` | `<6> message` |
| `raw` | `$m` | `message` |

Template variables: `$p` timestamp, `$l` level, `$i` PID, `$h` hostname, `$m` message, `$s` syslog severity number.

### `trace`
Parameters: `[ctx; msg]`

Write a trace-level message to stdout.

- `ctx` — symbol context tag (e.g. `` `mymodule ``)
- `msg` — string message

### `debug`
Parameters: `[ctx; msg]`

Write a debug-level message to stdout.

### `info`
Parameters: `[ctx; msg]`

Write an info-level message to stdout.

### `warn`
Parameters: `[ctx; msg]`

Write a warning-level message to stdout.

### `error`
Parameters: `[ctx; msg]`

Write an error-level message to stdout.

### `fatal`
Parameters: `[ctx; msg]`

Write a fatal-level message to stdout.

### `createLog`
Parameters: none

Returns an independent logger instance as a dictionary of functions. Each call returns a new instance with isolated state.

Returned keys: `` `trace`debug`info`warn`error`fatal`add`remove`setfmt`getfmt`addfmt`setlvl`getlvl ``

| Function | Parameters | Description |
|---|---|---|
| `trace`..`fatal` | `[msg]` | Write a message at the given level (filtered by active level) |
| `setlvl` | `[lvl]` | Set minimum level; one of `` `trace`debug`info`warn`error`fatal `` |
| `getlvl` | `[_]` | Return current minimum level |
| `setfmt` | `[name]` | Switch to a named format template |
| `getfmt` | `[_]` | Return current format name |
| `addfmt` | `[name;template]` | Register a new named format template |
| `add` | `[handle;lvls]` | Add a sink for one or more levels; returns the handle |
| `remove` | `[handle;lvl]` | Remove a sink from a level |

## Log dependency contract

The log dependency contract used across `di.*` modules requires a dictionary:

```q
`info`warn`error!({[ctx;msg] ...};{[ctx;msg] ...};{[ctx;msg] ...})
```

`di.log` satisfies this contract. You can also supply any custom implementation with the same signatures.


di/log/log.q
```diff
@@ -0,0 +1,190 @@
...
+/ format pattern handlers for createLog instances; each takes {[level;msg]}
+/ level is an uppercase string (e.g. "INFO"), msg is the formatted message string
+pattern:"plihms~"!(
+  {[x;y] string .z.p};
+  {[x;y] x};
+  {[x;y] string .z.i};
+  {[x;y] string .z.h};
+  {[x;y] y};
+  {[x;y] string .z.M.syslogLvl`$lower x};
+  {[x;y] "$"});
```
DI-Software-Engineering, line 27: `.z.M.syslogLvl` references the module namespace dictionary's `syslogLvl` key, but `syslogLvl` is a plain name defined at the top level of the file (line 12). At the time the `pattern` dict is evaluated, `.z.M` may not yet be set (it is typically assigned by the module loader after the file is loaded), so this lookup will fail with a rank/type error or silently return null. Access the file-level name directly as `syslogLvl` instead of `.z.M.syslogLvl`.

di/log/log.q
```diff
@@ -0,0 +1,190 @@
...
+i:0;
+inst:()!();
+
+/ factory helper: returns a 1-arg {[msg]} log function for the given level and instance
+/ each entry in sink is a (handle;sender) pair; sender is called with the formatted text
+makelevel:{[id;gv;lvl]
+  {[id;gv;lvl;msg]
+    if[(lvls?lvl)<lvls?gv`lvl;:()];
+    txt:fmtapply[gv`prep;upper string lvl;fmtmsg msg],nl;
+    {[txt;pair]
```
DI-Software-Engineering, line 89: The error handler in `makelevel` swallows all write errors silently: `@[last pair;txt;{x}]` catches any signal and returns the error string unnoticed. A failed write to a closed handle or a function sink that signals will be silently discarded with no indication of failure. Use a handler that at minimum writes to stderr, e.g. `{-2 "log sink error: ",x}`.

di/log/log.q
```diff
@@ -0,0 +1,190 @@
...
+    [fmt:first msg;
+     args:1_msg;
+     fmt:ssr[fmt;"%%";"\000"];
+     parts:"%" vs fmt;
+     subs:"sr"!({$[10h=abs type x;x;-11h=type x;string x;'`type]};.Q.s1);
+     result:first[parts],raze {[subs;args;parts;i]
+       part:parts[1+i];
+       code:first part;
+       rest:1_part;
+       if[not code in key subs;'`$"unsupported format char: ",enlist code];
```
DI-Software-Engineering, line 65: `fmtmsg` signals `` `type `` for a list message whose first element is not a 10h/string argument type (e.g. a symbol format string like `` (`fmt;arg) `` where `fmt` is a symbol), but only after partially processing. More critically, the `args` indexing `args[i]` will silently return null when `i` is out of range (more `%s`/`%r` specifiers in the format string than supplied arguments), producing malformed log lines with no error. There is no guard that `count args` matches the number of format specifiers.

di/log/log.q
```diff
@@ -0,0 +1,190 @@
...
+/ simple dependency-contract-compatible log functions; each has signature {[ctx;msg]}
+trace:{[ctx;msg] logline["TRACE";ctx;msg];};
+debug:{[ctx;msg] logline["DEBUG";ctx;msg];};
+info:{[ctx;msg] logline["INFO";ctx;msg];};
+warn:{[ctx;msg] logline["WARN";ctx;msg];};
+error:{[ctx;msg] logline["ERROR";ctx;msg];};
+fatal:{[ctx;msg] logline["FATAL";ctx;msg];};
+
+/ pre-wrapped log dependency dict ready to pass straight into any di.* module's init
+/ e.g. email.init[mylog.defaultdict]
```
DI-Software-Engineering: `add` accepts either a bare handle or a 2-element `(handle;fn)` pair, but the `remove` function (line 151) only matches on `first p` (the handle component). If the user registers the same handle twice with different sender functions, `remove` will remove both entries for that handle rather than just the one for the specified sender. The intent of storing `(handle;sender)` pairs is to allow the same handle with different senders, but `remove` conflates them. The fix is to match on the full pair: `not h~p` when `h` is a pair, or keep the existing handle-only match only when `h` is a scalar.

di/log/log.q
```diff
@@ -0,0 +1,190 @@
...
+  {[id;gv;lvl;msg]
+    if[(lvls?lvl)<lvls?gv`lvl;:()];
+    txt:fmtapply[gv`prep;upper string lvl;fmtmsg msg],nl;
+    {[txt;pair]
+      @[last pair;txt;{x}]
+    }[txt;] each (gv`sink)[lvl];
+  }[id;gv;lvl;]
+  };
+
+/ update field k with value v in instance id's state
```
DI-Software-Engineering, line 95: `updInst` writes back via `.z.m.inst:state`, but in `add` (line 112) the same assignment is done inline as `.z.m.inst:state` — bypassing `updInst`. This is inconsistent but not itself a bug. The real issue is that neither path is protected against concurrent modification: if two log instances call `add`/`remove` or `setlvl` concurrently (e.g. from a timer and a message handler in the same process), the read-modify-write on `.z.m.inst` is a race condition. This is a structural issue with the shared mutable dictionary design.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **4 warning(s)** | **0 suggestion(s)**

di/log/log.q
JamesMassey97: This is valid and will be addressed in commit. The function will be updated to ensure that the correct number of args are provided to %s/%r specifiers.

di/log/log.q
JamesMassey97: Dropping .z.M as suggested.

di/log/log.q
JamesMassey97: addressing to now print to stderr

di/log/log.q
JamesMassey97: this is a valid issue, fixing so there isn't a race.

di/log/log.q
JamesMassey97: valid, updated to take the two inputs to remove individual functions from the handle.

di/log/log.q
```diff
@@ -0,0 +1,187 @@
...
+debug:{[ctx;msg] logline["DEBUG";ctx;msg];};
+info:{[ctx;msg] logline["INFO";ctx;msg];};
+warn:{[ctx;msg] logline["WARN";ctx;msg];};
+error:{[ctx;msg] logline["ERROR";ctx;msg];};
+fatal:{[ctx;msg] logline["FATAL";ctx;msg];};
+
+/ pre-wrapped log dependency dict ready to pass straight into any di.* module's init
+/ includes all six levels (the contract only requires info/warn/error; the extras are free
+/ since createlog already computes them) e.g. email.init[mylog.logdict]
+logdict:(enlist`log)!enlist(lvls!createlog[]lvls);
```
DI-Software-Engineering, line 193: `logdict` is evaluated once at load time by calling `createlog[]`, which increments the shared instance counter `.z.m.i` and writes into `.z.m.inst`. Every subsequent `use`/reload of the module will create a new instance on top of any pre-existing state, but the `logdict` value itself is frozen at load time — the captured `gv`/`wv` closures inside the returned functions reference the instance id that existed at load time. This is expected for a module-level singleton, but combined with the missing newline at end of file (the last line has no trailing newline), loading this file a second time within the same process will silently accumulate orphaned entries in `inst` and increment `i`, which is a resource leak with no cleanup path. More critically, if the file is re-executed (e.g. via `\l`), `logdict` is rebound to a brand-new instance while any callers that captured the previous `logdict` still hold stale closures pointing at the old instance id — they continue to work but now hold a permanent reference into `inst` that can never be freed. Provide an explicit cleanup/reset of `inst` and `i` at the top of the file, or document that re-loading is unsupported.

di/log/log.q
```diff
@@ -0,0 +1,187 @@
...
+    }[txt;] each (gv`sink)[lvl];
+  }[id;gv;lvl;;]
+  };
+
+/ update field k with value v in instance id's state
+/ single indexed assignment on the shared global; avoids round-tripping through a local copy
+updinst:{[id;k;v]
+  .z.m.inst[id;k]:v;
+  };
+
```
DI-Software-Engineering, line 100: The `add` function performs a multi-level indexed assignment `inst[id;`sink;lvl],:enlist(handle;sender)` where `lvl` comes from iterating `(),sinklvls`. If `sinklvls` contains a symbol not present in the `sink` dict (i.e. not in `lvls`), this assignment will silently create a new key in `inst[id;`sink]` rather than erroring, producing an inconsistent sink dict whose extra keys are never iterated by `makelevel`. Validate that each element of `sinklvls` is a member of `lvls` before performing the assignment, or the user will see messages silently dropped with no error.

di/log/log.q
```diff
@@ -0,0 +1,187 @@
...
+  / add a sink (handle or (handle;fn) pair) for one or more log levels
+  / each sink entry is stored as (handle;sender) so mixed types never collide in a dict
+  add:{[id;gv;h;sinklvls]
+    handle:$[0h=type h;first h;h];
+    sender:$[0h=type h;last h;h];
+    {[id;handle;sender;lvl]
+      .z.m.inst[id;`sink;lvl],:enlist(handle;sender);
+    }[id;handle;sender;] each (),sinklvls;
+    handle
```
DI-Software-Engineering, line 157: `remove` only accepts a single `lvl` symbol, but `add` accepts a list of levels (`(),sinklvls`). A caller who added a sink across multiple levels with one `add` call must call `remove` once per level; calling `remove[h; `info`warn]` (a two-element list) will pass the whole list as the `lvl` index into the nested-assignment, silently producing wrong results or a type error rather than removing the sink from each level. Either accept a list in `remove` (mirroring `add`) or document and enforce the single-symbol constraint.

DI-Software-Engineering:
## DIReview Summary

di/log/log.q
JamesMassey97: the expectation is to create a log, or use the default and feed it init, not to reload using \l during the loaded module. Even if a user wanted to initialize multiple modules in a process, you would still be able to feed in the same dictionary.

di/log/log.q
JamesMassey97: add now will validate against known levels and throw an error if it is not.

di/log/log.q
JamesMassey97: Updated to mirror add.

JamesMassey97 (commented):
addressed automated comments and added commits. 

The main change compared to the initial commit on this is the logdict so that is once the log module is imported. It will run the createlog function and give the user a default dict to use. There is also default logs outside of this to align with existing example in consitency.md but I am not sure if this is necessary.

jprgrant (approved):

PR #95, opened 2026-06-03
Feature dbwrite

Ruairi-wq2:
DBwrite module code, including test.csv suite. Init.q and dbwrite.q code to define logic and export it. dbwrite.md as a README.

### DIFF
--- a/di/dbwrite/dbwrite.q
+++ b/di/dbwrite/dbwrite.q
@@ -1,0 +1,100 @@
+/ sort params table - populated by loadconfig
+params:([] tabname:`symbol$(); att:`symbol$(); column:`symbol$(); sort:`boolean$());
+
+/ save-down manipulation registry: tabname -> unary function
+savedownmanipulation:()!();
+
+/ default sort params - used when params is empty and no config file is set
+defaultparams:([] tabname:enlist`default; att:enlist`; column:enlist`time; sort:enlist 1b);
+
+/ load and validate sort.csv into .z.M.params
+/ file: hsym path; null falls back to .z.M.defaultfile
+loadconfig:{[file]
+  file:hsym file;
+  if[null file;file:.z.M.defaultfile];
+  if[null file;
+    dp:.z.M.defaultparams;
+    .z.m.loginfo[`dbwrite;"no sort config file set; using defaultparams"];
+    @[.z.M;`params;:;dp];
+    :[]];
+  p:@[
+    {.z.m.loginfo[`dbwrite;"retrieving sort settings from ",string x];("SSSB";enlist",")0:x};
+    file;
+    {[f;e]'"failed to open ",string[f],": ",e}[file]
+  ];
+  if[not all spcb:(spc:cols p) in `tabname`att`column`sort;
+    '"unrecognised columns (",(", " sv string spc where not spcb),") in ",string file];
+  if[not all atb:(at:distinct p`att) in ``p`s`g`u;
+    '"unrecognised attribute(s): ",", " sv string at where not atb];
+  @[.z.M;`params;:;p];
+  };
+
+/ apply a single kdb+ attribute to an on-disk column; logs and swallows errors
+applyattr:{[dloc;colname;att]
+  .z.m.loginfo[`dbwrite;"applying ",string[att]," attr to ",string[colname]," in ",string dloc];
+  .[{@[x;y;z#]};(dloc;colname;att);
+    {[dloc;colname;att;e]
+      .z.m.logerr[`dbwrite;"unable to apply ",string[att]," attr to ",string[colname]," in ",string[dloc],": ",e]
+    }[dloc;colname;att]
+  ];
+  };
+
+/ sort an on-disk table partition and apply attributes per sort.csv config
+/ d: tabname | (tabname;dir) | (tabname;list of dirs)
+sort:{[d]
+  if[0=count .z.M.params;.z.M.loadconfig .z.M.defaultfile];
+  .z.m.loginfo[`dbwrite;"sorting ",(st:string t:first d)," table"];
+  sp:$[count tabsp:select from .z.M.params where tabname=t;
+      [.z.m.loginfo[`dbwrite;"sort params found for: ",st];tabsp];
+    count defsp:select from .z.M.params where tabname=`default;
+      [.z.m.logwarn[`dbwrite;"no sort params for: ",st,"; using defaults"];defsp];
+    [.z.m.logwarn[`dbwrite;"no sort params for: ",st,"; skipping sort"];:()]];
+  {[sp;dloc]
+    if[count sortcols:exec column from sp where sort,not null column;
+      .z.m.loginfo[`dbwrite;"sorting ",string[dloc]," by: ",", " sv string sortcols];
+      .[xasc;(sortcols;dloc);
+        {[sc;dl;e]
+          .z.m.logerr[`dbwrite;"failed to sort ",string[dl]," by ",(", " sv string sc),": ",e]
+        }[sortcols;dloc]]];
+    if[count attrcols:select column,att from sp where not null att;
+      .z.M.applyattr[dloc;;]'[attrcols`column;attrcols`att]];
+  }[sp] each distinct (),last d;
+  .z.m.loginfo[`dbwrite;"finished sorting ",st," table"];
+  };
+
+/ apply registered pre-write manipulation to table x of type t
+/ returns modified table; on error logs and returns original unmodified table
+manipulate:{[t;x]
+  $[t in key .z.M.savedownmanipulation;
+    @[.z.M.savedownmanipulation[t];x;
+      {[x;e].z.m.logerr[`dbwrite;"save-down manipulation failed: ",e];x}[x]];
+    x]
+  };
+
+/ post-EOD hook - called after all tables written and sorted
+/ d: hdb directory (hsym), p: partition value (date)
+/ stub: override at the call site to add custom post-replay logic
+postreplay:{[d;p]};
+
+/ format current process memory stats as a loggable string
+memstats:{"mem stats: ",{"; "sv "=" sv'flip(string key x;(string value x),\:" MB")}`long$.Q.w[]%1048576};
+
+/ run .Q.gc[] and log before/after memory stats
+gc:{
+  .z.m.loginfo[`dbwrite;"starting garbage collect. ",.z.M.memstats[]];
+  r:.Q.gc[];
+  .z.m.loginfo[`dbwrite;"garbage collection returned ",(string `long$r%1048576),"MB. ",.z.M.memstats[]]
+  };
+
+init:{[config;deps]
+  / deps: `log!(logdict)
+  /   `log: `info`warn`error!(infofunc;warnfunc;errfunc) - required
+  / example: dbwrite.init[enlist[`log]!enlist logdep]
+  logdict:$[99h=type deps;$[(`log in key deps) and not (::)~deps`log;deps`log;()!()];()!()];
+  if[not count logdict;
+    '"di.dbwrite: log dependency is required; pass `info`warn`error functions - see di.log or refer to confluence documentation";
+  ];
+  .z.m.loginfo:logdict`info;
+  .z.m.logwarn:logdict`warn;
+  .z.m.logerr:logdict`error;
+  };
--- a/di/dbwrite/dbwrite.q
+++ b/di/dbwrite/dbwrite.q
@@ -1,8 +1,6 @@
 / sort params table - default row sorts all tables by time ascending
 params:([] tabname:enlist`default; att:enlist`; column:enlist`time; sort:enlist 1b);
 
-/ save-down manipulation registry: tabname -> unary function
-savedownmanipulation:()!();
 
 / load and validate sort.csv into .z.M.params
 / file: hsym path; null warns and resets params to default row
@@ -69,19 +67,27 @@ sort:{[d]
     ]]
   };
 
-/ apply registered pre-write manipulation to table x of type t
-/ returns modified table; on error logs and returns original unmodified table
-manipulate:{[t;x]
-  $[t in key .z.M.savedownmanipulation;
-    @[.z.M.savedownmanipulation[t];x;
-      {[x;e].z.m.logerr[`dbwrite;"save-down manipulation failed: ",e];x}[x]];
-    x]
+
+/ write table to a date-partitioned hdb: enumerate syms, apply p# to sym if present, write, then sort
+/ dir: hdb root (hsym); part: partition value (date/month/int); tabname: symbol; data: in-memory table
+savedown:{[dir;part;tabname;data]
+  .z.m.loginfo[`dbwrite;"saving ",string[tabname]," partition ",string[part]," to ",string dir];
+  path:.Q.par[dir;part;tabname];
+  data:.Q.en[dir;data];
+  path set $[`sym in cols data;@[data;`sym;{`p#x}];data];
+  sort[(tabname;path)];
+  .z.m.loginfo[`dbwrite;"finished saving ",string tabname];
   };
 
-/ post-EOD hook - called after all tables written and sorted
-/ d: hdb directory (hsym), p: partition value (date)
-/ stub: override at the call site to add custom post-replay logic
-postreplay:{[d;p]};
+/ upsert data into an existing on-disk partition and re-sort
+/ dir: hdb root (hsym); part: partition value; tabname: symbol; data: in-memory table
+upsert:{[dir;part;tabname;data]
+  .z.m.loginfo[`dbwrite;"upserting ",string[tabname]," partition ",string[part]," in ",string dir];
+  path:.Q.par[dir;part;tabname];
+  .[path;();,;.Q.en[dir;data]];
+  sort[(tabname;path)];
+  .z.m.loginfo[`dbwrite;"finished upserting ",string tabname];
+  };
 
 / format current process memory stats as a loggable string
 memstats:{[]"mem stats: ",{"; "sv "=" sv'flip(string key x;(string value x),\:" MB")}`long$.Q.w[]%1048576};
@@ -94,10 +100,8 @@ gc:{
   };
 
 init:{[config;deps]
-  / config: dict with optional keys
-  /   `savedownmanipulation: tabname!function dict of pre-write manipulation functions
-  / deps: `log!(logdict)
-  /   `log: `info`warn`error!(infofunc;warnfunc;errfunc) - required
+  / config: unused, pass (::)
+  / deps: `log!(logdict) - `info`warn`error!(infofunc;warnfunc;errfunc) - required
   logdict:$[99h=type deps;$[(`log in key deps) and not (::)~deps`log;deps`log;()!()];()!()];
   if[not count logdict;
     '"di.dbwrite: log dependency is required; pass `info`warn`error functions - see di.log or refer to confluence documentation";
--- a/di/dbwrite/dbwrite.q
+++ b/di/dbwrite/dbwrite.q
@@ -72,21 +72,23 @@ sort:{[d]
 / dir: hdb root (hsym); part: partition value (date/month/int); tabname: symbol; data: in-memory table
 savedown:{[dir;part;tabname;data]
   .z.m.loginfo[`dbwrite;"saving ",string[tabname]," partition ",string[part]," to ",string dir];
-  path:.Q.par[dir;part;tabname];
+  path:` sv (.Q.par[dir;part;tabname];`);
   data:.Q.en[dir;data];
   path set $[`sym in cols data;@[data;`sym;{`p#x}];data];
   sort[(tabname;path)];
   .z.m.loginfo[`dbwrite;"finished saving ",string tabname];
   };
 
-/ upsert data into an existing on-disk partition and re-sort
+/ append data to an existing on-disk partition; enumerate syms but do not sort
+/ call sort separately when the partition is complete
 / dir: hdb root (hsym); part: partition value; tabname: symbol; data: in-memory table
-upsert:{[dir;part;tabname;data]
-  .z.m.loginfo[`dbwrite;"upserting ",string[tabname]," partition ",string[part]," in ",string dir];
-  path:.Q.par[dir;part;tabname];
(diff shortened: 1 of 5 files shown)


Ruairi-wq2:
```
q)k4unit:use`di.k4unit
q)k4unit.moduletest[`di.dbwrite]
2026.06.03T09:37:00.789 start
2026.06.03T09:37:00.789 :/home/ruairi_conlon/TorQ/di/kdbx-modules/di/dbwrite/test.csv 60 test(s)
2026.06.03T09:37:00.803 end
Test results:
....
true   0  0     q    `AAPL`IBM~exec sym from get `:dbwrite_def_tp/                                          1      :/home/ruairi_conlon/TorQ/di/kdbx-modules/di/dbwrite/test.csv 0   0       1  1    1       1     2026.06.03T09:37:00.803
true   0  0     q    4=logcount                                                                             1      :/home/ruairi_conlon/TorQ/di/kdbx-modules/di/dbwrite/test.csv 0   0       1  1    1       1     2026.06.03T09:37:00.803
All tests passed
```

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,108 @@
...
+memstats:{[]"mem stats: ",{"; "sv "=" sv'flip(string key x;(string value x),\:" MB")}`long$.Q.w[]%1048576};
+
+/ run .Q.gc[] and log before/after memory stats
+gc:{
+  .z.m.loginfo[`dbwrite;"starting garbage collect. ",.z.M.memstats[]];
+  r:.Q.gc[];
+  .z.m.loginfo[`dbwrite;"garbage collection returned ",(string `long$r%1048576),"MB. ",.z.M.memstats[]]
+  };
+
+init:{[config;deps]
```
alowrydi: config argument not actually called within this function?

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,108 @@
...
+  $[t in key .z.M.savedownmanipulation;
+    @[.z.M.savedownmanipulation[t];x;
+      {[x;e].z.m.logerr[`dbwrite;"save-down manipulation failed: ",e];x}[x]];
+    x]
+  };
+
+/ post-EOD hook - called after all tables written and sorted
+/ d: hdb directory (hsym), p: partition value (date)
+/ stub: override at the call site to add custom post-replay logic
+postreplay:{[d;p]};
```
jonnypress: I don’t think this should be in here

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,108 @@
...
+        if[count attrcols:select column,att from sp where not null att;
+          .z.M.applyattr[dloc;;]'[attrcols`column;attrcols`att]];
+      }[sp] each distinct (),last d;
+      .z.m.loginfo[`dbwrite;"finished sorting ",st," table"]
+    ]]
+  };
+
+/ apply registered pre-write manipulation to table x of type t
+/ returns modified table; on error logs and returns original unmodified table
+manipulate:{[t;x]
```
jonnypress: I don’t think this should be in here

jonnypress:
You need to think about this a bit more and apply some judgement. When we are stripping code out of TorQ we are trying to créate a set of standalone, isolated modules. What is the purpose of this module? It is to provide a set of utilities for writing tables, including sorting them on disk. Postreplay and savedownmanipulation likely don’t have relevance in here (the are artifacts of other parts of TorQ). Can you also coordinate with Alex because there’s overlap with the sort lib, I think the sort lib should live in here.

Ruairi-wq2:
As it stands, the sort function operates on one table & partition combo. A list of partitions can be passed.

/ end-of-day, one partition
dbwrite.sort[`trade; .Q.par[`:hdb; .z.d; `trade]]

Considering adding the following use case functions. Not sure if they would get used/be useful. 

- sortpart[dir;part] — sort all tables in one partition (e.g., today's date), no need to list tables or build paths manually
- sortdb[dir] — sort everything (useful for rebuild, potentially slow)
- sortdb[dir;partitions] — sort a specific set of partitions across all tables

Also probably a style choice, do we want the modules to be lightweight and functional or bulkier with a bit more usecase coverage??

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,225 @@
...
+/ internal - log and rethrow a csv read failure
+readerr:{[file;e]
+  / build the message once, surface it under the dbwrite context, then rethrow it to the caller
+  m:"failed to read ",string[file],": ",e;
+  .z.m.log[`error]["dbwrite: ",m];
+  'm;
+  };
+
+/ internal - validate the header and data rows, then parse csv lines into a config table
+parsecsv:{[lines]
```
jprgrant, line 73: I think AI is overthinking things here - it has decided 0: csv reading isn't fit for purpose and decided to rebuild it from first principles. But I think reinventing basic q functionality sets a bad precedent if we give claude free reign, we'll end up with a lot of bloated code where AI is solving problems that don't actually exist in practice. I think this function should be much shorter - just process the lines using 0:, and leave the validation to checkconfig

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,225 @@
...
+  if[count bad:where (count each rows)<>count hdr;
+    '"di.dbwrite: csv data row(s) ",(", " sv string 1+bad)," must have ",(string count hdr)," fields"];
+  c:hdr!$[count rows; flip rows; (count hdr)#enlist ()];
+  if[not all (c`sort) in enlist each "01";
+    '"di.dbwrite: the sort column must contain only 0 or 1"];
+  :([] tabname:`$c`tabname; att:`$c`att; column:`$c`column; sort:"B"$c`sort);
+  };
+
+/ internal - validate a sort-config table, signalling a clear error if it is malformed
+checkconfig:{[t]
```
jprgrant, line 81: The errors here - should they also be printed through .z.m.log[error] , like the readerr one above? If not why not- I think we should be consistent unless there is a reason I'm missing. If the log library returns the string as well as printing it (I'm not sure it does - do check), then this can be a one liner, log and throw the result:
'.z.m.log[`error] "di.dbwrite: ..."
If the log function doesn't return anything then it would have to be:
.z.m.log[`error] errstr:"di.dbwrite: ..."; ' errstr;

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,225 @@
...
+
+/ internal - log a sort message then return the resolved rows
+logreturn:{[lvl;msg;rows]
+  / keeps each branch body in getsortparams to a single statement
+  .z.m.log[lvl]["dbwrite: ",msg];
+  :rows;
+  };
+
+/ internal - resolve which config rows apply to a table
+getsortparams:{[config;tab;st]
```
jprgrant: I don't like 'st' being a parameter to this function, given it is just the 'tab' name duplicated as a string. Just create st as a local variable in here again if required. There is no performance consideration here - to be honest I would likely just use string[tab] everywhere st is used, but that is a stylistic choice. However function parameters should only be added when they are strictly necessary, this one isn't since it is derivable from another param

jprgrant (changes requested):

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+  / t must be a table with columns tabname,att,column,sort
+  checkconfig t;
+  .z.m.sortconfig:t;
+  };
+
+getconfig:{[]
+  / return the current sort config stored in .z.m.sortconfig; (::) if not yet set
+  :.z.m.sortconfig;
+  };
+
```
DI-Software-Engineering, line 56: `readfile` returns raw lines from `read0`, but is wrapped with `@[readfile; file; readerr[file]]` in `readcsv`. The three-arg `@` trap form is `@[f;x;h]` — it calls `f[x]`, so this becomes `readfile[file]`, which is correct. However, `readerr` is defined as `{[file;e] ...}` (a binary), and the trap handler receives only the error string as its single argument. The projected form `readerr[file]` creates a unary `{e -> ...}[file]` projection that receives the error string as `e` — this is correct. No bug here. (Self-cancelling — omit.)

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+    .z.m.logerr[`parsecsv;err:"di.dbwrite: csv data row(s) ",(", " sv string 1+bad)," must have ",(string count hdr)," fields"];
+    'err;
+  ];
+  c:hdr!$[count rows; flip rows; (count hdr)#enlist ()];
+  if[not all (c`sort) in enlist each "01";
+    .z.m.logerr[`parsecsv;err:"di.dbwrite: the sort column must contain only 0 or 1"];
+    'err;
+  ];
+  :([] tabname:`$c`tabname; att:`$c`att; column:`$c`column; sort:"B"$c`sort);
+  };
```
DI-Software-Engineering, line 78: `parsecsv` checks that `(asc distinct hdr)~\`att\`column\`sort\`tabname` to validate header columns, but this check passes for a CSV that has exactly those four column names *in any order including duplicates*, since `asc distinct` would still match if e.g. `tabname` appeared twice and one other column were missing (e.g. `tabname,att,tabname,sort` → distinct gives 3 symbols, asc of which won't match the 4-symbol target — actually it would not match, so duplicates are caught). However, the check does not catch a header containing the four required columns *plus* additional columns, because `asc distinct hdr` of a 5-element header will never match the 4-element target. Wait — a 5-column CSV is rejected here, which is the intended behaviour per the docs. The real gap is: `count hdr` is used to validate data rows, but `hdr` is derived from the first line with `\`$"," vs first lines`. If the CSV file uses `\r\n` line endings (Windows), `first lines` from `read0` will include a trailing `\r` in the last header token, making `hdr` contain e.g. `` `sort\r `` instead of `` `sort ``, and the header validation will fail with a confusing error rather than silently misparse. This is a robustness issue on Windows-originated CSVs.

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+  ];
+  c:cols t;
+  badcols:c where not c in `tabname`att`column`sort;
+  if[count badcols;
+    .z.m.logerr[`checkconfig;err:"di.dbwrite: unrecognised config column(s): ",", " sv string badcols];
+    'err;
+  ];
+  missingcols:(`tabname`att`column`sort) where not (`tabname`att`column`sort) in c;
+  if[count missingcols;
+    .z.m.logerr[`checkconfig;err:"di.dbwrite: missing required config column(s): ",", " sv string missingcols];
```
DI-Software-Engineering, line 95: When the CSV has data rows, `flip rows` produces a list of string vectors (one per column), keyed by `hdr`. But when `rows` is empty (header-only CSV), the branch `(count hdr)#enlist ()` is taken instead, producing a list of empty generic lists. The subsequent cast `\`$c\`tabname` etc. works on the empty case. However, `"B"$c\`sort` where `c\`sort` is `()` (generic empty) will produce a boolean empty fine. The real issue is in the non-empty branch: `c:hdr!$[count rows; flip rows; ...]` — here `flip rows` where `rows` is a list of char vectors (from `"," vs/:`) gives a list of string vectors. Then `\`$c\`tabname` converts each string to a symbol. But `"B"$c\`sort` casts each sort string (e.g. `"1"`) to boolean using the `"B"` upper-case *parse* cast, not `"b"`. Per q facts, `"B"$` parses strings. `"B"$"1"` parses the string `"1"` — but `"1"` is not a standard boolean literal; q boolean parsing expects `"0b"` or `"1b"` format, not bare `"0"`/`"1"`. This means `"B"$"1"` will produce `0N` (null) rather than `1b`, silently corrupting the sort column. The validation check `if[not all (c\`sort) in enlist each "01";` catches strings not in `{"0","1"}` before the cast, but the cast itself (`"B"$"1"`) does not produce `1b` — it should be `"b"$"01" ? c\`sort` or map `"01"!01b` to convert. Fix: replace `"B"$c\`sort` with `"01"!01b` applied via `("01"!01b) c\`sort`.

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+  sp:getsortparams[config;tabname;st];
+  if[not count sp; :()];
+  sortdir[sp] each distinct (),dirs;
+  .z.m.loginfo[`sort;"finished sorting the ",st," table"];
+  };
+
+/ internal - resolve which config rows apply to a table
+getsortparams:{[config;tab;st]
+  / tab is the table-name symbol (NOT a table); named to avoid clashing with the tabname column
+  / a table uses its own rows; unlisted tables fall back to the default row, else are skipped
```
DI-Software-Engineering, line 139: `checkconfig` checks `if[98h<>type t; ...]` to require a table, but a keyed table has type `99h` and would be rejected with the message "config must be a table". This is intentional and the error message is clear, so it is fine as-is. However, `checkconfig` also accesses `t\`sort` as `t\`sort` (column access on a table returns the column vector), then checks `if[not 1h=type t\`sort; ...]`. A boolean column in a plain table has type `1h`, so this is correct. No bug.

di/dbwrite/dbwrite.q
DI-Software-Engineering, line 95: `"B"$c\`sort` will not correctly parse bare `"0"`/`"1"` strings to booleans. The upper-case `"B"$` is the *parse* form, which expects boolean literals like `"0b"`/`"1b"` — not bare digit strings. `"B"$"1"` yields null (`0Nb`), not `1b`. The validation gate above only checks that the raw strings are in `"01"` before this cast, so the cast will always produce nulls for valid input. Fix: use `"01"!01b` to map the validated strings, e.g. `("01"!01b) each c\`sort`.

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+  / sort by the columns flagged sort=1b, then hand every row to applyattr (it skips non-attributes)
+  sortcols:exec column from sp where sort, not null column;
+  if[count sortcols; sortcolumns[dloc;sortcols]];
+  applyattr[dloc;;]'[sp`column;sp`att];
+  };
+
+/ internal - log an attribute application failure without rethrowing
+attrerr:{[dl;cn;at;e]
+  / logs failure and continues so other columns and partitions still get processed
+  .z.m.logerr[`attrerr;"unable to apply ",string[at]," attr to ",string[cn]," in ",string[dl],": ",e];
```
DI-Software-Engineering, line 178: `sortdir` calls `applyattr[dloc;;]'[sp\`column;sp\`att]` for every row, including rows where `sort=1b` and `att` is the empty symbol. `applyattr` correctly guards `if[not att in \`p\`s\`g\`u; :()]` and returns early for non-attributes, so this is safe. However, `sortdir` also reads `sortcols:exec column from sp where sort, not null column` — the `not null column` guard means a row with a null column and `sort=1b` is silently excluded from the sort key, which may surprise callers who set `sort=1b` but leave `column` null. `checkconfig` already rejects null columns, so in practice this guard is unreachable after validation, but it is dead/redundant rather than a bug.

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,254 @@
...
+  };
+
+savedown:{[dir;part;tabname;data]
+  / write an in-memory table to a date-partitioned hdb partition, then sort it per .z.m.sortconfig
+  / dir: hdb root (hsym); part: partition value (date/month/int); tabname: symbol; data: in-memory table
+  / enumerates syms against the hdb sym file; sorting and attributes are driven by .z.m.sortconfig
+  .z.m.loginfo[`savedown;"saving ",string[tabname]," partition ",string[part]," to ",string dir];
+  path:` sv (.Q.par[dir;part;tabname];`);
+  path set .Q.en[dir;data];
+  sort[tabname;path];
```
DI-Software-Engineering, line 200: `savedown` writes the partition with `path set .Q.en[dir;data]` (no `p#` applied in-memory before write), then calls `sort[tabname;path]` to sort and apply attributes on disk. This is a deliberate design change from the old code. However, `path` is constructed as `` ` sv (.Q.par[dir;part;tabname];`) `` — this appends a trailing backtick, producing a path with a trailing `/`. The `sort` function receives this path and passes it to `xasc` via `.[xasc;(sortcols;dloc);...]`. `xasc` on a splayed table requires the directory path (with trailing `/`). This is consistent and correct. No bug.

di/dbwrite/dbwrite.q
DI-Software-Engineering, line 95: In `parsecsv`, the column-reorder test `(asc distinct hdr)~\`att\`column\`sort\`tabname` is used instead of checking `(asc hdr)~asc\`tabname\`att\`column\`sort` — using `distinct` means a header with a duplicated column name (e.g. `tabname,att,column,tabname`) would have the duplicate removed before comparison, and if that still leaves 4 unique names matching the required set, it would pass validation despite the duplicate. The data rows would then have 4 fields but the `hdr` (post-distinct) would be 4 symbols, so the row-length check `(count each rows)<>count hdr` uses the original `hdr` (4 items), not the distinct one — wait, `hdr` is set once as `` \`$"," vs first lines `` and never modified. The `distinct` is only used inside the check expression. So a 4-field duplicate header like `tabname,att,column,tabname` would have `asc distinct` = `` \`att\`column\`tabname `` (3 items), which won't match the 4-item target — so duplicates are caught. No bug.

DI-Software-Engineering, line 95: Duplicate finding — removing (already reported at the primary location above). Self-cancelling.

DI-Software-Engineering:
## DIReview Summary

**3 critical** | **6 warning(s)** | **0 suggestion(s)**

di/dbwrite/dbwrite.q
ascottDI: Agree with this, 
I have moved certain validation checks from parsecsv into the checkconfig function and also updated the file reading to use 0:

di/dbwrite/dbwrite.q
ascottDI: When updating the logging to align with the newer standards we have set this issue had already been addressed. 

Change already made so no changes required off the back of this.

di/dbwrite/dbwrite.q
ascottDI: Interesting point, makes a lot of sense. Have updated the getsortparams function following this.

di/dbwrite/dbwrite.q
```diff
@@ -0,0 +1,236 @@
...
+  'm;
+  };
+
+/ internal - parse csv lines into a config table
+parsecsv:{[lines]
+  / load via 0: - the header row (enlist delim) names the columns, so column order does not
+  / matter; cast sort to boolean if present. structural validation is left to checkconfig.
+  t:((count "," vs first lines)#"S";enlist",") 0: lines;
+  :$[`sort in cols t; update sort:"B"$string sort from t; t];
+  };
```
DI-Software-Engineering, line 78: When `lines` contains only a header row (zero data rows), `first lines` is the header string, but `"B"$string sort` in the subsequent `update` will be applied to a column of type S (symbols cast from the empty `sort` column). For the zero-data-row case this is likely harmless, but the real problem is `(count "," vs first lines)#"S"` — using a uniform `"S"` type vector for all columns means the `sort` column is loaded as symbols (e.g. `` `1 ``, `` `0 ``), not as booleans or even strings. `"B"$string sort` then converts `` `1``→`"1"`→`1b` and `` `0``→`"0"`→`0b`, which works for clean data, but `` `x``→`"x"`→`0b` silently coerces instead of being rejected — contradicting the stated intent. More critically, any `att` value that is a valid symbol (`` ` ``, `` `p ``, etc.) will be loaded correctly, but `tabname`/`column` values that happen to look like numbers (e.g. `1`, `0`) will be interned as symbols without issue, so actual data corruption for those is not a concern. The real defect is silent coercion of invalid `sort` values: a non-`0`/`1` sort entry becomes `0b` without error. The doc says this is intentional, but the `checkconfig` validator must then reject `0b`-coerced rows from a value like `"x"` — if it does not check the original string, these pass silently. Fix by loading `sort` as `"C"` (character/string) and validating before casting, or keep the all-`"S"` load but add an explicit pre-cast validity check in `checkconfig` for the original symbol values.

DI-Software-Engineering:
## DIReview Summary

di/dbwrite/dbwrite.q
ascottDI: Self cancelled.

di/dbwrite/dbwrite.q
ascottDI: With new updates following Jamie's comments this is no longer an issue. (and even if it werent this is the "problem that doesnt exist in practice" he was talking about)

di/dbwrite/dbwrite.q
ascottDI: refers to code that has already been removed/changed in the rewrite following jamie grants comments so no changes needed

di/dbwrite/dbwrite.q
ascottDI: false positive, ignored, therefore, no changes

di/dbwrite/dbwrite.q
ascottDI: Quoted code doesnt exist anymore following Jamie grants feedback.

Additionally this exact situation is tested in tests and it passes

di/dbwrite/dbwrite.q
ascottDI: Correctly identified as not a bug.

No changes

di/dbwrite/dbwrite.q
ascottDI: correctly identified as not a bug

no changes.

di/dbwrite/dbwrite.q
ascottDI: Talking about already deleted / fixed code following Jamie Grants feedback. 

Regardless, correctly identified as no bug anyway.

ascottDI: Bot withdrew it.

No change

di/dbwrite/dbwrite.q
ascottDI: Intentional and documented. We chose to use 0: and checkconfig rather than manual parsing. This is the same thing Jamie was talking about earlier "AI is solving problems that don't actually exist in practice". 

No changes needed.

jprgrant (approved):

PR #96, opened 2026-06-03
Rename/di.tplogutils

Olly99999:
Renamed di.tplogutils to di.tplog

Fixed testemptylog[] test

image


alowrydi:
It looks like you have made changes on the same branch as the memstats rename and they now both exist in this pull request. Worth separating so each has its own separate branch and PR.

alowrydi:
Otherwise, the tplog renaming looks complete and consistent across the four files.
 
Could you add a bit more detail on the testemptylog fix? From the PR description it's not clear what was wrong, where the fix was made, or what changed. The whole file shows as added rather than a line-by-line diff so it's difficult to identify the specific change from the PR alone.

Olly99999:
image

Original tests had an error in testemptylog[] test. 

I changed line 259 in test.q from testfile set 0#0x00; to testfile set () so it correctly creates an empty logfile and now the test runs properly

KyleBonnes (changes requested):
Similar to the memstats PR -- need to actually remove the existing "old name" files and dirs rather than just adding the new ones

Olly99999:
Removed old tplogutils

di/tplog/tplog.md
```diff
@@ -185,9 +185,9 @@ test_repair_recovers_messages[]
 test_repair_garbage_at_end[]
 ```
 
-> **Note:** If the tests refer to `tplogsutil` but you loaded the module as `tplogutils`, either:
-> - load the module into a `tplogsutil` variable as well, or
-> - update the test references to `tplogutils`.
+> **Note:** If the tests refer to `tplog` but you loaded the module as `tplog`, either:
```
KyleBonnes: This doesn't make any sense anymore -- see that the original note is talking about "if you loaded the module with a different name"
Adjust this to convery the same type of meaning as before

KyleBonnes (changes requested):

di/tplog/tplog.md
Olly99999: Changed it to a general explanation with same meaning

di/tplog/tplog.md
```diff
@@ -185,9 +185,9 @@ test_repair_recovers_messages[]
 test_repair_garbage_at_end[]
 ```
 
-> **Note:** If the tests refer to `tplogsutil` but you loaded the module as `tplogutils`, either:
-> - load the module into a `tplogsutil` variable as well, or
-> - update the test references to `tplogutils`.
+> **Note:** If the tests reference the module under a different name than the one used when loading it, either:
+> - load the module under the expected name as well, or
+> - update the test references to use the loaded module name.
```
KyleBonnes, line 190: Perfect 👍

KyleBonnes (approved):

PR #101, opened 2026-06-15
Feature: di.html

matthewv192:
image

- Ported the old TorQ html code into the new kdb-x module format - dropped the `\d .html` namespace and the global `::` assigns, loads via `use` now with a proper export list.
- Logging is injected in now and is required (init will complain if you don't pass it), same idea for the .z handlers - you can inject them or let it wire them up itself. No more hard-coded `.dotz.set` calls.
- Tidied up init so it takes one config dict and won't double-wire the handlers if you call it twice.
- General cleanup to match the style guide - one statement per line, lowercase names (renamed `readpagereplaceHP` to `readpagereplacehostport`), etc. Broke up some of the denser one-liners too.
- Fixed up `jstsiso8601` so it's vectorised and handles nulls, and got rid of the deprecated `"z"` datetime cast. Added docs and k4unit tests, all passing.

### DIFF
--- a/di/html/html.md
+++ b/di/html/html.md
@@ -1,0 +1,165 @@
+# di.html
+
+WebSocket pub/sub and HTML page serving module, extracted from TorQ.
+
+## Overview
+
+This module does two things:
+
+1. **Pub/sub over WebSockets** — allows browser clients to subscribe to kdb+ tables and receive live updates as data is published.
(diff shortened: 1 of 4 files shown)


Ruairi-wq2:
Closed this as I have duplicated it from main branch, as opposed to from forked branch.

PR #102, opened 2026-06-15
Feature Sort

ascottDI:
# di.sort — TorQ Modularisation PR

## Summary

Extracts the `.sort` namespace from TorQ's `dbwriteutils.q` into a standalone kdb-x module: `di.sort`. The module sorts on-disk kdb+ tables and applies column attributes, driven by a `sort.csv` configuration file. It satisfies the di.* module contract: dependency injection via `init`, a clean exported API, and no hard dependencies beyond an injected logger.

---

## Background

TorQ's `dbwriteutils.q` contains a `.sort` namespace used to sort HDB partitions and apply kdb+ attributes (`p`, `s`, `g`, `u`) to columns after data is written to disk. This is a common post-write step in kdb+ database maintenance — sorting by a symbol column and applying the parted attribute, for example, is standard practice for efficient data access.

This PR is part of the broader TorQ → kdb-x modularisation effort. The goal is to make each functional area independently loadable, testable, and injectable, removing the dependency on TorQ's global process framework.

---

### New files

| File | Description |
|---|---|
| `di/sort/sort.q` | Core implementation — `init`, `getsortcsv`, `sorttab`, `getparams` and internal helpers |
| `di/sort/init.q` | Module entry point — loads `sort.q` and declares the export list |
| `di/sort/test.csv` | k4unit test suite (63 tests) |
| `di/sort/sort.md` | Module README |

| Aspect | TorQ `.sort` | `di.sort` |
|---|---|---|
| Logging | Hard-coded `.lg.o` / `.lg.e` calls | Injected `log` dependency (`{[ctx;msg]}` functions) |
| Config path | `.proc.getconfigfile["sort.csv"]` global lookup | `savedir` key in `init` deps; defaults to `` `:sort.csv `` |
| Global state | Direct `@[\`.sort;\`params;:;params]` amend | `.z.m.params` mutable module state |
| Error handling | Errors propagate unchecked in some paths | `@[f;args;handler]` traps on `xasc` and `applyattr`; per-partition failures are logged and do not halt the run |
| Partial log dict | Not applicable | `init` validates all three keys (`info`, `warn`, `error`) are present and errors with a descriptive message if not |

---

## Exported API

```q
srt:use`di.sort

srt.init[deps]              / wire dependencies; required before all other calls
srt.getsortcsv[file]        / load and validate sort config from CSV
srt.sorttab[d]              / sort on-disk partitions for one table
srt.getparams[]             / return currently loaded sort config table
```

---

## sort.csv format

```
tabname,att,column,sort
trade,p,sym,1
trade,g,vol,0
default,p,sym,1
```

- `sort=1` — column is used as a sort key (`xasc`)
- `sort=0` — attribute applied without sorting
- `tabname=default` — fallback row for tables not listed explicitly

---

## Test coverage

Tests are in `test.csv` and run via `k4unit`:

```q
k4unit.moduletest`di.sort
```

### Areas covered

| Area | Tests |
|---|---|
| `init` — dependency validation | Rejects `::`, null log, empty dict, missing `log` key, non-dict log value, partial log dict (missing keys) |
| `init` — error message prefix | `"di.sort"` prefix on all thrown errors |
| `init` — valid configs | Accepts valid log dep with and without `savedir` |
| `getparams` — initial state | Empty table with correct schema before any `getsortcsv` call |
| `sorttab` — auto-load | Triggers `getsortcsv` from `savedir` when `params` is empty |
| `getsortcsv` — validation | Rejects bad column names, unknown `att` values, missing required columns, non-existent file |
| `getsortcsv` — empty CSV | Header-only file loads without error; `params` remains empty |
| `getsortcsv` — successful load | Correct row count, column schema, tabname/att/sort values |
| `sorttab` — no-params path | Returns `()` when no matching row and no default; emits warn log |
| `sorttab` — empty partition list | Handles `()` partition list without error |
| Log verification — getsortcsv | Correct info messages and `getsortcsv` context symbol on success; error message and context on file failure |
| Log verification — sorttab | Logs sort start/completion, table-specific lookup, column list, default fallback, warn on no-params |
| Log verification — applyattr | Info message with `applyattr` context symbol on attribute application |
| Log verification — error paths | Error messages logged under correct context symbols for both `xasc` and `applyattr` failures |

**Total: 63 tests, all passing.**

### Test results screenshot
image
image

## Integration test

Verified against real on-disk HDB data with two trade partitions and one quote partition:

- Unsorted `sym` column sorted ascending; `p` (parted) attribute applied correctly
- `vol` column received `g` (grouped) attribute without being used as a sort key
- `price` column (no sort config row) reordered in step with `sym` sort; no attribute applied
- `default` row applied correctly to the `quote` table
- Table with no matching params and no default row: warn logged, returns `()` without error

---

## Notes

- `di.sort` has no hard module dependencies. Only the injected `log` dict is required.
- `di.log` satisfies the log contract out of the box: `logdep:\`info\`warn\`error!(log.info;log.warn;log.error)`.
- A single `xasc` failure on one partition is logged as an error and the run continues — consistent with TorQ's original intent that a bad partition should not halt a full HDB sort.
- The `default` row in `sort.csv` is looked up by `tabname=\`default`. Any table name matching a row exactly takes precedence over `default`.

### DIFF
--- a/di/sort/sort.q
+++ b/di/sort/sort.q
@@ -18,6 +18,9 @@ init:{[deps]
     '"di.sort: log dependency is required; pass `info`warn`error functions - ",
      "see di.log for a default implementation or refer to confluence documentation";
   ];
+  if[not all (`info`warn`error) in key logdict;
+    '"di.sort: log dict must have `info`warn`error keys; got: ",(", " sv string key logdict);
+  ];
   .z.m.loginfo:logdict`info;
   .z.m.logwarn:logdict`warn;
   .z.m.logerr:logdict`error;
--- a/di/sort/sort.q
+++ b/di/sort/sort.q
@@ -1,121 +1,133 @@
 / library for sorting and applying attributes to on-disk kdb+ tables
 
-/ sort configuration table - populated by getsortcsv, read via getparams
-params:([] tabname:`symbol$(); att:`symbol$(); column:`symbol$(); sort:`boolean$());
-
-/ valid attribute symbols accepted in sort.csv att column
+/ attributes that may legitimately be applied on disk (empty leaves a column unattributed)
 validatts:``p`s`g`u;
 
 init:{[deps]
-  / wire injectable log dependency (required) and optional config from deps dict
-  / optional keys: `savedir - hsym path to sort.csv, defaults to `:sort.csv
-  logdict:$[99h=type deps;
-    $[`log in key deps;
-      $[99h=type deps`log; deps`log; ()!()];
-      ()!()];
-    ()!()];
-  if[not count logdict;
-    '"di.sort: log dependency is required; pass `info`warn`error functions - ",
-     "see di.log for a default implementation or refer to confluence documentation";
-  ];
-  if[not all (`info`warn`error) in key logdict;
-    '"di.sort: log dict must have `info`warn`error keys; got: ",(", " sv string key logdict);
-  ];
-  .z.m.loginfo:logdict`info;
-  .z.m.logwarn:logdict`warn;
-  .z.m.logerr:logdict`error;
-  .z.m.defaultfile:$[`savedir in key deps; deps`savedir; `:sort.csv];
+  / wire the injectable log dependency so the module reports through the host's logger
+  / deps: a dict with a `log key -> `info`warn`error!({[c;m]};{[c;m]};{[c;m]})
+  / see di.log for a default implementation, or pass any matching dict
+  / example: srt.init[enlist[`log]!enlist logdep]
+  if[99h<>type deps;
+    '"di.sort: deps must be a dict with a `log key; see di.log for a default logger"];
+  if[not `log in key deps;
+    '"di.sort: log dependency is required; pass `info`warn`error functions keyed on `log"];
+  if[99h<>type deps`log;
+    '"di.sort: log value must be a dict of `info`warn`error functions"];
+  if[not all (`info`warn`error) in key deps`log;
+    '"di.sort: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  .z.m.log:deps`log;
   };
 
-/ internal - log and rethrow a csv load error
-csvloaderr:{[f;e]
-  / logs the failure then rethrows so getsortcsv surfaces the error to the caller
-  .z.m.logerr[`getsortcsv;"failed to open ",string[f],". the error was: ",e];
-  '"failed to open ",string[f],". the error was: ",e;
-  };
-
-getsortcsv:{[file]
-  / load and validate sort configuration from a CSV file; populates the params table
+readcsv:{[file]
+  / convenience for the common case where config lives in a csv - returns it, does not store
+  / pass the result to sorttab, e.g. srt.sorttab[srt.readcsv `:sort.csv;`trade;dirs]
+  / the csv must have columns tabname,att,column,sort in that order
   file:hsym file;
-  rawparams:@[
-    {.z.m.loginfo[`getsortcsv;"retrieving sort settings from ",string x]; ("SSSB";enlist",")0:x};
-    file;
-    csvloaderr[file]
-  ];
-  spc:cols rawparams;
-  badcols:spc where not spc in `tabname`att`column`sort;
-  if[count badcols;
-    '"unrecognised columns (",(", " sv string badcols),") in ",string file];
-  missingcols:(`tabname`att`column`sort) where not (`tabname`att`column`sort) in spc;
-  if[count missingcols;
-    '"missing required columns (",(", " sv string missingcols),") in ",string file];
-  at:distinct rawparams`att;
-  badatts:at where not at in validatts;
-  if[count badatts;
-    '"unrecognised type of attribute - ",(", " sv string badatts)];
-  .z.m.params:rawparams;
-  .z.m.loginfo[`getsortcsv;"loaded ",(string count rawparams)," sort config rows from ",string file];
+  t:@[readfile; file; readerr[file]];
+  .z.m.log[`info][`readcsv;"read ",(string count t)," sort param row(s) from ",string file];
+  :t;
   };
 
-sorttab:{[d]
-  / sort and apply attributes to on-disk partitions for one table
-  st:string t:first d;
-  if[0=count params; getsortcsv[.z.m.defaultfile]];
-  .z.m.loginfo[`sorttab;"sorting the ",st," table"];
-  sp:getsortparams[t;st];
+/ internal - read and parse a sort-config csv (the protected action in readcsv)
+readfile:{[file]
+  / kept named rather than inline so readcsv reads cleanly and matches the style guide
+  .z.m.log[`info][`readcsv;"reading sort params from ",string file];
+  :("SSSB";enlist",") 0: file;
+  };
+
+/ internal - log and rethrow a csv read failure
+readerr:{[file;e]
+  / surfaces the failure to the caller after logging it under the readcsv context
+  .z.m.log[`error][`readcsv;"failed to read ",string[file],": ",e];
+  '"failed to read ",string[file],": ",e;
+  };
+
+sorttab:{[config;tabname;dirs]
+  / sort and apply attributes to the on-disk partition directories for one table
+  / config: a sort-config table (build it directly or via readcsv); tabname: symbol; dirs: hsym or list of hsyms
+  / example: srt.sorttab[srt.readcsv `:sort.csv;`trade;`:/hdb/2024.01.01/trade]
+  checkconfig config;
+  st:string tabname;
+  .z.m.log[`info][`sorttab;"sorting the ",st," table"];
+  sp:getsortparams[config;tabname;st];
   if[not count sp; :()];
-  sortdir[sp] each distinct (),last d;
-  .z.m.loginfo[`sorttab;"finished sorting the ",st," table"];
+  sortdir[sp] each distinct (),dirs;
+  .z.m.log[`info][`sorttab;"finished sorting the ",st," table"];
   };
 
-getparams:{[]
-  / return the current sort configuration table
-  :params;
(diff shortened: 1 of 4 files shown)


jonnypress (commented):
I think we need to modify the interface a little. We should be able to pass in a table for the sort params, which should be possible to read from a csv file, but isn’t required to be in a csv file. Does it do that? Don’t be afraid to rethink the interface to make it more user friendly - it looks quite complex at the minute.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+  :t;
+  };
+
+/ internal - read and parse a sort-config csv (the protected action in readcsv)
+readfile:{[file]
+  / kept named rather than inline so readcsv reads cleanly and matches the style guide
+  .z.m.log[`info][`readcsv;"reading sort params from ",string file];
+  :("SSSB";enlist",") 0: file;
+  };
+
```
adamhamilton96, line 38: The format string `"SSSB"` in `readfile` parses all four CSV columns as `S S S B` (symbol symbol symbol boolean). The `att` column is a symbol, but the `column` column is also a symbol — that is fine. However, the `att` column will read the empty string `""` as the symbol `` ` `` which is the intended no-attribute sentinel, so that part is correct. The real bug is that a missing or mismatched column count in the CSV will silently misalign the parse (e.g. `tmp_sort_3col.csv` has only 3 columns but `"SSSB"` will still attempt to read 4 columns, producing a type error or wrong data rather than a clear config-validation error). This is a correctness/robustness issue: `0:` with a type-list longer than the actual column count causes a length error instead of the clear `di.sort:` prefixed error that `checkconfig` would produce, so the user gets a confusing low-level error rather than the documented one.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+sorttab:{[config;tabname;dirs]
+  / sort and apply attributes to the on-disk partition directories for one table
+  / config: a sort-config table (build it directly or via readcsv); tabname: symbol; dirs: hsym or list of hsyms
+  / example: srt.sorttab[srt.readcsv `:sort.csv;`trade;`:/hdb/2024.01.01/trade]
+  checkconfig config;
+  st:string tabname;
+  .z.m.log[`info][`sorttab;"sorting the ",st," table"];
+  sp:getsortparams[config;tabname;st];
+  if[not count sp; :()];
+  sortdir[sp] each distinct (),dirs;
```
adamhamilton96, line 71: `getsortparams` is called with `[config;tabname;st]` where `tabname` is the raw symbol and `st` is its string form. Inside `getsortparams` the parameter is named `t`, and `select from config where tabname=t` compares the config `tabname` column against the local `t` — but the local variable `t` shadows the column name `tabname` only if q resolves the column first. In q, inside `select … where tabname=t`, `tabname` refers to the column and `t` refers to the local variable; this works correctly. However, the variable name `t` is also the name of the function parameter for the table name, and the first line of the function body does `select from config where tabname=t` — `config` is both the table and a parameter name used elsewhere: this is fine in isolation, but within `getsortparams`, `config` is passed as the first parameter and is used as a table in the select. The real risk: `tabname` is both a column name in `config` AND a column name in the result. The `select from config where tabname=t` statement uses `t` (the local symbol parameter), but if `t` were ever a table it would be silently misinterpreted. This is a latent naming-collision risk rather than an immediate bug, but worth noting. More concretely: `getsortparams` returns `0#config` (the empty config table) on the no-match path, but `sorttab` checks `if[not count sp; :()]` — `count` of a 0-row table is 0, so this returns `()` correctly. No bug here, but the `0#config` shape is important and only works because `count` of an empty table is 0.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+    '"di.sort: unrecognised config column(s): ",", " sv string badcols];
+  missingcols:(`tabname`att`column`sort) where not (`tabname`att`column`sort) in c;
+  if[count missingcols;
+    '"di.sort: missing required config column(s): ",", " sv string missingcols];
+  if[not 1h=type t`sort;
+    '"di.sort: the sort column must be boolean"];
+  badatts:at where not (at:distinct t`att) in validatts;
+  if[count badatts;
+    '"di.sort: unrecognised attribute(s) in att column: ",", " sv string badatts];
+  };
```
adamhamilton96, line 92: `checkconfig` checks for unrecognised columns with `badcols:c where not c in \`tabname\`att\`column\`sort` and then separately checks for missing columns. However, a table with all four required columns PLUS extra columns will trigger the `badcols` error. That is intentional. But a table with only a subset of the four columns and no extra columns will pass the `badcols` check and only fail at `missingcols`. The real bug is in the `badcols` check itself: it uses `c where not c in \`tabname\`att\`column\`sort`, which flags ANY column not in the allowed set — including legitimate columns. This means a config table with exactly `\`tabname\`att\`column\`sort` passes, but one with those four in a different order also passes because `in` is order-independent. This is correct. No bug here on reflection — withdrawing this point.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+/ internal - sort one partition directory by the given columns
+sortcolumns:{[dloc;sortcols]
+  / split out of sortdir so the conditional body there stays a single statement
+  .z.m.log[`info][`sorttab;"sorting ",string[dloc]," by these columns: ",", " sv string sortcols];
+  .[xasc;(sortcols;dloc);
+    sorterr[sortcols;dloc]];
+  };
+
+/ internal - sort columns and apply attributes for a single on-disk partition directory
+sortdir:{[sp;dloc]
```
adamhamilton96, line 128: In `sortdir`, the `attrcols` filter is `select column,att from sp where not null att` but does NOT filter out the empty-symbol `` ` `` sentinel (which represents "no attribute"). In q, `null` of a symbol returns `0b` for `` ` `` (the empty symbol is NOT null — `null `` ` `` ` is `0b`). So rows with `att:`` ` `` `` (the no-attribute sentinel, which is the empty symbol) will pass the `not null att` filter and be included in `attrcols`, causing `applyattr` to be called with `att=`` ` ``. `applyattr` does guard against this with `if[null att; :()]`, but `null`` ` `` ` is `0b` in q, so that guard also fails to catch the empty symbol. The `@[{@[x;y;z#]};(dloc;colname;`` ` ``);attrerr[…]]` call will then attempt to apply `` ` ``# to the column on disk, which will either be a no-op or error. The `validatts` list includes `` ` `` as a valid att value meaning "none", but it should be explicitly excluded from the `attrcols` selection in `sortdir` using `where (not null att), att<>`` ` `` ` or equivalently `where att in \`p\`s\`g\`u`.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+
+/ internal - log an attribute application failure without rethrowing
+attrerr:{[dl;cn;at;e]
+  / logs failure and continues so other columns and partitions still get processed
+  .z.m.log[`error][`applyattr;"unable to apply ",string[at]," attr to the ",string[cn]," column in ",string[dl],". the error was: ",e];
+  :();
+  };
+
+/ internal - apply a single attribute to a specific column in an on-disk partition
+applyattr:{[dloc;colname;att]
```
adamhamilton96, line 144: In `applyattr`, the null-att guard `if[null att; :()]` uses `null`, which in q returns `0b` for the empty symbol `` ` ``. So passing `att:`` ` `` `` bypasses this guard. The comment says "sortdir only passes non-null atts", but as noted above, `not null`` ` `` ` is `1b`, so sortdir WILL pass the empty symbol through. The guard here should be `if[(null att) or att=`` ` ``; :()]` or simply `if[not att in \`p\`s\`g\`u; :()]` to be safe.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+/ internal - log and rethrow a csv read failure
+readerr:{[file;e]
+  / surfaces the failure to the caller after logging it under the readcsv context
+  .z.m.log[`error][`readcsv;"failed to read ",string[file],": ",e];
+  '"failed to read ",string[file],": ",e;
+  };
+
+sorttab:{[config;tabname;dirs]
```
adamhamilton96, line 60: `readerr` constructs the rethrown error message by string-concatenating `e` directly: `'"failed to read ",string[file],": ",e`. In q, the error value `e` passed to a trap handler is a string (character vector). If `e` is not a string (e.g. it is a symbol or other type in some edge cases), this concatenation will fail with a type error, swallowing the original error. This is a low-probability edge case but could mask real failures.

di/sort/sort.q
```diff
@@ -0,0 +1,134 @@
...
+  / config: a sort-config table (build it directly or via readcsv); tabname: symbol; dirs: hsym or list of hsyms
+  / example: srt.sorttab[srt.readcsv `:sort.csv;`trade;`:/hdb/2024.01.01/trade]
+  checkconfig config;
+  st:string tabname;
+  .z.m.log[`info][`sorttab;"sorting the ",st," table"];
+  sp:getsortparams[config;tabname;st];
+  if[not count sp; :()];
+  sortdir[sp] each distinct (),dirs;
+  .z.m.log[`info][`sorttab;"finished sorting the ",st," table"];
+  };
```
adamhamilton96, line 73: `sorttab` does not validate that `tabname` is a symbol before passing it to `getsortparams`. The test CSV (test.csv line ~72) includes a test `srt.sorttab[cfg;42;enlist\`:/]` that is expected to `fail`, but `checkconfig` only validates `config` — it does not validate `tabname`. The call `getsortparams[config;42;st]` where `st:string 42` = `"42"` will attempt `select from config where tabname=42`, which in q produces a type error (comparing a symbol column to a long). This will signal an error, so the test passes, but the error message will be a raw q type error rather than the documented `di.sort:`-prefixed error. The API contract says errors are prefixed `di.sort:` but this case breaks that contract.

adamhamilton96:
## DIReview Summary

**3 critical** | **4 warning(s)** | **0 suggestion(s)**

PR #103, opened 2026-06-16
Asyncdispatch

Patrickguest1998:
di.asyncdispatch (new, standalone)
The async multi-process query coordinator extracted from the gateway. Handles queuing queries, dispatching to available servers, collecting async results, applying join functions, timeout management, and error handling.

Standalone value: any process that needs scatter-gather across multiple backends.

Key data structures:

queryqueue   / pending and active queries
results      / per-query result accumulator (queryid -> servertype -> result)
clients      / connected client tracking

Pluggable hooks (the consumer sets these, not hardcoded):

formatresponse   / {[status;sync;result] ...} customise result delivery
availableservers / {[excludeinuse] ...} customise server selection
getnextqueryid   / {[] ...} customise scheduling (default: FIFO)
Injected deps: logging, timer (optional, for timeouts).
No framework dependency required.

Screenshot 2026-06-16 151600


di/asyncdispatch/test.csv
```diff
@@ -0,0 +1,66 @@
...
+run,0,0,q,ad.removeclienthandle[2i],1,,disconnect the client
+true,0,0,q,0~count select from qq[] where clienth=2i,null returntime,1,client unfinished queries cleared
+
+/ Test 8: pluggable hooks
+run,0,0,q,ad.setformatresponse[{[status;sync;result]result}],1,,override formatresponse to pass result through
+true,0,0,q,5~ad.formatresponse[1b;0b;5],1,,overridden formatresponse is used
+run,0,0,q,ad.setgetnextqueryid[{()}],1,,override scheduler to never pick a query
+true,0,0,q,0~count ad.getnextqueryid[],1,,overridden scheduler is used
+run,0,0,q,ad.setcallbacks[`myresult;`myerror],1,,override callback symbols
+true,0,0,q,(`myresult;`myerror)~(.m.di.0asyncdispatch.resultcallback;.m.di.0asyncdispatch.errorcallback),1,,callback symbols updated
```
jonnypress, line 66: These tests all have stubbed responses as far as I can see. Can you we add running up a new process to test against instead of stubbing?

di/asyncdispatch/init.q
```diff
@@ -0,0 +1,12 @@
+/ Load core functionality into root module namespace
+\l ::asyncdispatch.q
+
+export:([servers;queryqueue;clients;results;
+        errorprefix;querykeeptime;clearinactivetime;synccallsallowed;
```
jonnypress: We don’t need to export all the functions and tables. Can you work out which ones are useful to the user and reduce to that.

jonnypress (changes requested):

di/asyncdispatch/test.csv
Patrickguest1998: yep will do

di/asyncdispatch/init.q
Patrickguest1998: understood

PR #105, opened 2026-06-18
Feature grafana

Olly99999:
image

### DIFF
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -35,7 +35,9 @@ wired:0b;
 
 zpp:{[x]
   / parse a grafana http post request and dispatch to the matching handler
-  / cut at the first whitespace to isolate the api url from any function params
+  / kdb passes .z.pp (requeststring;headers); requeststring is "<endpoint> <body>"
+  / (method and leading / already stripped), so cut at the first space gives
+  / (endpoint;body) - e.g. "query {...}" -> ("query";"{...}")
   r:(0;n?" ")cut n:first x;
   rqt:.j.k r 1;
   .z.m.log[`info][`grafana;"received ",(r 0)," request"];
@@ -50,9 +52,17 @@ annotation:{[rqt]
   };
 
 query:{[rqt]
-  / dispatch a /query request to the timeseries or table builder by target type
-  rqtype:raze rqt[`targets]`type;
-  :.h.hy[`json]$[rqtype~"timeserie";tsfunc rqt;tbfunc rqt];
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
+  / a single target may arrive as a bare dict; normalise to a list of targets
+  tgts:$[99h=type tgts;enlist tgts;tgts];
+  / iterate by index so a list or a table of targets is handled the same way
+  build:{[rqt;tgts;i]t:tgts i;r:@[rqt;`targets;:;t];$[t[`type]~"timeserie";tsfunc r;tbfunc r]};
+  / strip the [ ] from each per-target array, drop empties, re-wrap as one array
+  inners:{1_-1_x}each build[rqt;tgts]each til count tgts;
+  inners:inners where 0<count each inners;
+  :.h.hy[`json]"[",(","sv inners),"]";
   };
 
 search:{[rqt]
@@ -173,8 +183,9 @@ tsfunc:{[x]
   mil:{floor epoch+(`long$x)%1000000};
   / ensure the time column is a timestamp
   if["p"<>meta[rqt][.z.m.timecol;`t];rqt:@[rqt;.z.m.timecol;+;.z.D]];
-  / restrict to the time range requested by grafana
-  range:"P"$-1_'x[`range]`from`to;
+  / restrict to the time range requested by grafana - grafana sends iso-8601 utc;
+  / strip a trailing Z only if present rather than blindly dropping the last char
+  range:"P"${$["Z"=last x;-1_x;x]}each x[`range]`from`to;
   rqt:?[rqt;enlist(within;.z.m.timecol;range);0b;()];
   / add the milliseconds-since-epoch column grafana expects
   rqt:@[rqt;`msec;:;mil rqt .z.m.timecol];
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -23,6 +23,9 @@ types:.Q.t!`array`boolean,(3#`null),(5#`number),11#`string;
 / milliseconds between 1970.01.01 and 2000.01.01
 epoch:946684800000;
 
+/ internal flag - set true once the http handlers have been installed
+wired:0b;
+
 / -----------------------------------------------------------------------------
 / request handling
 / -----------------------------------------------------------------------------
@@ -32,14 +35,14 @@ zpp:{[x]
   / cut at the first whitespace to isolate the api url from any function params
   r:(0;n?" ")cut n:first x;
   rqt:.j.k r 1;
-  .z.m.loginfo[`grafana;"received ",(r 0)," request"];
+  .z.m.log[`info][`grafana;"received ",(r 0)," request"];
   handler:$["query"~r 0;query;"search"~r 0;search;annotation];
-  :.[handler;enlist rqt;{[e].z.m.logerr[`grafana;"failed to process request: ",e];'e}];
+  :.[handler;enlist rqt;{[e].z.m.log[`error][`grafana;"failed to process request: ",e];'e}];
   };
 
 annotation:{[rqt]
   / annotation endpoint is not yet implemented
-  .z.m.logwarn[`grafana;"annotation url not yet implemented"];
+  .z.m.log[`warn][`grafana;"annotation url not yet implemented"];
   :`$"Annotation url nyi";
   };
 
@@ -237,20 +240,24 @@ init:{[deps]
   / examples:
   /   grafana.init[enlist[`log]!enlist mylog]
   /   grafana.init[`log`config!(mylog;enlist[`ticks]!enlist 500)]
-  logdict:$[99h=type deps;$[(`log in key deps)and not(::)~deps`log;deps`log;()!()];()!()];
-  if[not count logdict;'"di.grafana: log dependency is required; pass `info`warn`error functions - see di.log for a default implementation"];
-  .z.m.loginfo:logdict`info;
-  .z.m.logwarn:logdict`warn;
-  .z.m.logerr:logdict`error;
+  / validate the required log dependency - nested guards, no eager `and`
+  if[99h<>type deps;'"di.grafana: deps must be a dict with `log key"];
+  if[not`log in key deps;'"di.grafana: log dependency is required; pass `info`warn`error functions keyed on `log"];
+  if[99h<>type deps`log;'"di.grafana: log value must be a dict; pass `info`warn`error functions"];
+  if[not all`info`warn`error in key deps`log;'"di.grafana: log dict must have `info`warn`error keys; got: ",", " sv string key deps`log];
+  .z.m.log:deps`log;
   / overlay any supplied configuration onto the defaults
-  config:$[99h=type deps;$[(`config in key deps)and not(::)~deps`config;deps`config;()!()];()!()];
-  if[count config;
-    vars:`timecol`sym`timebackdate`ticks`del inter key config;
-    (.Q.dd[.z.M]each vars)set'config vars;
+  config:$[`config in key deps;deps`config;()!()];
+  if[99h=type config;
+    if[`timecol in key config;.z.m.timecol:config`timecol];
+    if[`sym in key config;.z.m.sym:config`sym];
+    if[`timebackdate in key config;.z.m.timebackdate:config`timebackdate];
+    if[`ticks in key config;.z.m.ticks:config`ticks];
+    if[`del in key config;.z.m.del:config`del];
    ];
   / install the http handlers once, preserving any existing definitions
-  if[not`wired in key .z.M;sethandlers[]];
-  .z.m.loginfo[`grafana;"initialised grafana json datasource adaptor"];
+  if[not wired;sethandlers[]];
+  .z.m.log[`info][`grafana;"initialised grafana json datasource adaptor"];
   };
 
 getconfig:{
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -222,6 +222,18 @@ tablesym:{[coln;rqt;symname]
 / initialisation
 / -----------------------------------------------------------------------------
 
+normlog:{[logdict]
+  / detect kx.log instance by presence of kx.log-specific keys (getlvl, sinks, fmts)
+  / kx.log functions are monadic - wrap each into binary {[c;m]} and embed context in the message
+  / plain {[c;m]} log dicts (info`warn`error only) pass through unchanged
+  $[any `getlvl`sinks`fmts in key logdict;
+    `info`warn`error!(
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`info;];
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`warn;];
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`error;]);
+    logdict]
+  };
+
 sethandlers:{
   / wrap any existing .z.pp/.z.ph so grafana requests are intercepted and every
   / other request falls through to the original handler
@@ -234,27 +246,26 @@ sethandlers:{
 
 init:{[deps]
   / wire the logging dependency and any configuration, then install the handlers
-  / deps - `log!enlist logdict or `log`config!(logdict;configdict)
-  /   logdict    - `info`warn`error!({[c;m]};{[c;m]};{[c;m]}) - required
-  /   configdict - optional overrides for `timecol`sym`timebackdate`ticks`del
+  / deps - a single dict holding the log dependency and any config overrides
+  /   `log        - `info`warn`error!({[c;m]};{[c;m]};{[c;m]}) - required
+  /   config keys - optional `timecol`sym`timebackdate`ticks`del, alongside `log
   / examples:
   /   grafana.init[enlist[`log]!enlist mylog]
-  /   grafana.init[`log`config!(mylog;enlist[`ticks]!enlist 500)]
+  /   grafana.init[`log`ticks!(mylog;500)]
   / validate the required log dependency - nested guards, no eager `and`
   if[99h<>type deps;'"di.grafana: deps must be a dict with `log key"];
   if[not`log in key deps;'"di.grafana: log dependency is required; pass `info`warn`error functions keyed on `log"];
   if[99h<>type deps`log;'"di.grafana: log value must be a dict; pass `info`warn`error functions"];
-  if[not all`info`warn`error in key deps`log;'"di.grafana: log dict must have `info`warn`error keys; got: ",", " sv string key deps`log];
-  .z.m.log:deps`log;
-  / overlay any supplied configuration onto the defaults
-  config:$[`config in key deps;deps`config;()!()];
-  if[99h=type config;
-    if[`timecol in key config;.z.m.timecol:config`timecol];
-    if[`sym in key config;.z.m.sym:config`sym];
-    if[`timebackdate in key config;.z.m.timebackdate:config`timebackdate];
-    if[`ticks in key config;.z.m.ticks:config`ticks];
-    if[`del in key config;.z.m.del:config`del];
-   ];
+  / normalise a kx.log instance into the binary {[c;m]} contract; plain dicts pass through
+  lg:normlog deps`log;
+  if[not all`info`warn`error in key lg;'"di.grafana: log dict must have `info`warn`error keys; got: ",", " sv string key lg];
+  .z.m.log:lg;
+  / overlay any supplied config - config values sit alongside `log in deps
+  if[`timecol in key deps;.z.m.timecol:deps`timecol];
+  if[`sym in key deps;.z.m.sym:deps`sym];
+  if[`timebackdate in key deps;.z.m.timebackdate:deps`timebackdate];
+  if[`ticks in key deps;.z.m.ticks:deps`ticks];
+  if[`del in key deps;.z.m.del:deps`del];
   / install the http handlers once, preserving any existing definitions
   if[not wired;sethandlers[]];
   .z.m.log[`info][`grafana;"initialised grafana json datasource adaptor"];
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -267,13 +267,19 @@ normlog:{[logdict]
     logdict]
   };
 
+isgrafana:{[x]
+  / true if the request's header dict carries the X-Grafana-Org-Id header; guard
+  / the type so a non-dict last element (malformed/non-http arg) falls through
+  $[99h=type h:last x;(`$"X-Grafana-Org-Id")in key h;0b]
+  };
+
 sethandlers:{
   / wrap any existing .z.pp/.z.ph so grafana requests are intercepted and every
   / other request falls through to the original handler
   .z.m.prevpp:$[@[{value x;1b};`.z.pp;0b];.z.pp;{[x]}];
-  .z.pp:{[x]$[(`$"X-Grafana-Org-Id")in key last x;.z.m.zpp x;.z.m.prevpp x]};
+  .z.pp:{[x]$[.z.m.isgrafana x;.z.m.zpp x;.z.m.prevpp x]};
   .z.m.prevph:$[@[{value x;1b};`.z.ph;0b];.z.ph;{[x]}];
-  .z.ph:{[x]$[(`$"X-Grafana-Org-Id")in key last x;"HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n";.z.m.prevph x]};
+  .z.ph:{[x]$[.z.m.isgrafana x;"HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n";.z.m.prevph x]};
   .z.m.wired:1b;
   };
 
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -1,0 +1,259 @@
+/ grafana json datasource adaptor for kdb-x
+/ implements the simpod json datasource api so a grafana server can query
+/ kdb+ tables and timeseries data over http - the /search endpoint populates
+/ the grafana dropdowns and the /query endpoint returns timeseries or table data
+
+/ -----------------------------------------------------------------------------
+/ configuration - defaults applied here, overridden via the config dict in init
+/ -----------------------------------------------------------------------------
+
+/ name of the time column used for timeseries queries
+timecol:`time;
+/ name of the sym column used to split data by instrument
+sym:`sym;
+/ how far back to look when finding distinct syms
+timebackdate:2D;
+/ number of ticks to return for table requests
+ticks:1000;
+/ delimiter separating the arguments within a query target
+del:".";
+
+/ json type for each kdb datatype, keyed by .Q.t character
+types:.Q.t!`array`boolean,(3#`null),(5#`number),11#`string;
+/ milliseconds between 1970.01.01 and 2000.01.01
+epoch:946684800000;
+
+/ -----------------------------------------------------------------------------
+/ request handling
+/ -----------------------------------------------------------------------------
+
+zpp:{[x]
+  / parse a grafana http post request and dispatch to the matching handler
+  / cut at the first whitespace to isolate the api url from any function params
+  r:(0;n?" ")cut n:first x;
+  rqt:.j.k r 1;
+  .z.m.loginfo[`grafana;"received ",(r 0)," request"];
+  handler:$["query"~r 0;query;"search"~r 0;search;annotation];
+  :.[handler;enlist rqt;{[e].z.m.logerr[`grafana;"failed to process request: ",e];'e}];
+  };
+
+annotation:{[rqt]
+  / annotation endpoint is not yet implemented
+  .z.m.logwarn[`grafana;"annotation url not yet implemented"];
+  :`$"Annotation url nyi";
+  };
+
+query:{[rqt]
+  / dispatch a /query request to the timeseries or table builder by target type
+  rqtype:raze rqt[`targets]`type;
+  :.h.hy[`json]$[rqtype~"timeserie";tsfunc rqt;tbfunc rqt];
+  };
+
+search:{[rqt]
+  / build the grafana dropdown options from the tables available in the process
+  tabs:tables[];
+  symtabs:tabs where sym in'cols each tabs;
+  timetabs:tabs where timecol in'cols each tabs;
+  rsp:string tabs;
+  if[count timetabs;
+    rsp,:s1:prefix["t";string timetabs];
+    rsp,:s2:prefix["g";string timetabs];
+    / suffix the numeric columns for the graph and other panel options
+    rsp,:raze(s2,'del),/:'c1:string {cols[x] where`number=types(0!meta x)`t}each timetabs;
+    rsp,:raze(prefix["o";string timetabs],'del),/:'c1;
+    if[count symtabs;
+      / suffix the distinct syms for the timeseries and other panel options
+      rsp,:raze(s1,'del),/:'c2:string each finddistinctsyms'[timetabs];
+      rsp,:raze(prefix["o";string timetabs],'del),/:'{x[0]cross del,'string finddistinctsyms x 1}each(enlist each c1),'timetabs;
+     ];
+   ];
+  :.h.hy[`json].j.j rsp;
+  };
+
+finddistinctsyms:{[x]
+  / distinct syms seen in table x within the configured lookback window
+  :?[x;enlist(>;timecol;(-;.z.p;timebackdate));1b;{x!x}enlist sym]sym;
+  };
+
+prefix:{[c;s]
+  / prefix string c and the delimiter to each string in s
+  :(c,del),/:s;
+  };
+
+/ -----------------------------------------------------------------------------
+/ fetching the last n ticks
+/ -----------------------------------------------------------------------------
+
+diskvals:{[x]
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-ticks)+til ticks;
+  :get'[.Q.ind[x;c]];
+  };
+
+memvals:{[x]
+  / last `ticks` rows of an in-memory table
+  :get'[?[x;enlist(within;`i;count[x]-ticks,0);0b;()]];
+  };
+
+catchvals:{[x]
+  / try the on-disk path first, fall back to the in-memory path on error
+  :@[diskvals;x;{[x;y]memvals x}[x]];
+  };
+
+/ -----------------------------------------------------------------------------
+/ target parsing helpers
+/ -----------------------------------------------------------------------------
+
+istype:{[targ;char]
+  / test whether the target is prefixed with char followed by the delimiter
+  :(char,del)~2#targ;
+  };
+isfunc:istype[;"f"];
+istab:istype[;"t"];
+
+/ -----------------------------------------------------------------------------
+/ building json responses
+/ -----------------------------------------------------------------------------
+
+tabresponse:{[colname;coltype;rqt]
+  / build a table response in the json datasource schema
+  :.j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table);
+  };
+
+tbfunc:{[rqt]
+  / process a table request and return the json datasource table response
+  rqt:raze rqt[`targets]`target;
+  symname:0b;
+  / strip the type prefix: f.t.func drops 4, f.func drops 2, t.tab leaves the
+  / table name plus an optional sym
+  rqt:0!value $[isfunc[rqt]&istab 2_rqt;4_rqt;
+                isfunc rqt;2_rqt;
+                istab rqt;[rqt:`$del vs rqt;if[2<count rqt;symname:rqt 2];rqt 1];
+                rqt];
+  colname:cols rqt;
+  coltype:types(0!meta rqt)`t;
+  / filter to a single sym if one was supplied in the target
+  if[-11h=type symname;rqt:?[rqt;enlist(=;sym;enlist symname);0b;()]];
+  :tabresponse[colname;coltype;rqt];
+  };
+
+tsfunc:{[x]
+  / process a timeseries request and route to the correct panel/sym builder
+  targ:raze x[`targets]`target;
+  / split the target into its delimited arguments
+  numargs:count args:$[isfunc targ;(0;1+targ?del)cut targ:2_targ;`$del vs targ];
+  tyargs:$[10h=abs type args 0;`$1#;]args 0;
+  coln:cols rqt:0!value args 1;
+  / convert a timestamp to milliseconds since the unix epoch
+  mil:{floor epoch+(`long$x)%1000000};
+  / ensure the time column is a timestamp
+  if["p"<>meta[rqt][timecol;`t];rqt:@[rqt;timecol;+;.z.D]];
+  / restrict to the time range requested by grafana
+  range:"P"$-1_'x[`range]`from`to;
+  rqt:?[rqt;enlist(within;timecol;range);0b;()];
+  / add the milliseconds-since-epoch column grafana expects
+  rqt:@[rqt;`msec;:;mil rqt timecol];
+  / dispatch on the number of arguments and the panel type
+  $[(2<numargs)and`g~tyargs;graphsym[args 2;rqt];
+    (2<numargs)and`t~tyargs;tablesym[coln;rqt;args 2];
+    (2=numargs)and`g~tyargs;graphnosym[coln;rqt];
+    (2=numargs)and`t~tyargs;tablenosym[coln;rqt];
+    (4=numargs)and`o~tyargs;othersym[args;rqt];
+    (3=numargs)and`o~tyargs;othernosym[args 2;rqt];
+    (2=numargs)and`o~tyargs;othernosym[coln except timecol;rqt];
+    `$"Wrong input"]
+  };
+
+buildnosym:{[x;y;z]
+  / append one column's datapoints to the response accumulator
+  :y,`target`datapoints!(z 0;value each ?[x;();0b;z!z]);
+  };
+
+nosymresponse:{[rqt;colname]
+  / build a no-sym datapoints response across all requested columns
+  :.j.j buildnosym[rqt]\[();colname];
+  };
+
+othernosym:{[coln;rqt]
+  / timeseries response for a non-specific panel with no sym split
+  colname:coln cross`msec;
+  :nosymresponse[rqt;colname];
+  };
+
+graphnosym:{[coln;rqt]
+  / timeseries response for a graph panel with no sym split (numeric columns only)
+  coln:-1_coln where`number=types(0!meta rqt)`t;
+  colname:coln cross`msec;
+  :nosymresponse[rqt;colname];
+  };
+
+tablenosym:{[coln;rqt]
+  / timeseries response for a table panel with no sym split
+  coltype:types -1_(0!meta rqt)`t;
+  :tabresponse[coln;coltype;rqt];
+  };
+
+othersym:{[args;rqt]
+  / timeseries response for a non-specific panel returning a single sym's data
+  outcol:args[2],`msec;
+  data:flip value flip?[rqt;enlist(=;sym;enlist args 3);0b;outcol!outcol];
+  :.j.j enlist`target`datapoints!(args 3;data);
+  };
+
+graphsym:{[colname;rqt]
+  / timeseries response for a graph panel returning each sym's data
+  syms:`$string ?[rqt;();1b;{x!x}enlist sym]sym;
+  outcol:colname,`msec;
+  build:{[outcol;rqt;x;y]data:flip value flip?[rqt;enlist(=;sym;enlist y);0b;outcol!outcol];:x,`target`datapoints!(y;data)};
+  :.j.j build[outcol;rqt]\[();syms];
+  };
+
+tablesym:{[coln;rqt;symname]
+  / timeseries response for a table panel filtered to a single sym
+  coltype:types -1_(0!meta rqt)`t;
+  rqt:?[rqt;enlist(=;sym;enlist symname);0b;()];
+  :tabresponse[coln;coltype;rqt];
+  };
+
+/ -----------------------------------------------------------------------------
+/ initialisation
+/ -----------------------------------------------------------------------------
+
+sethandlers:{
+  / wrap any existing .z.pp/.z.ph so grafana requests are intercepted and every
+  / other request falls through to the original handler
+  .z.m.prevpp:$[@[{value x;1b};`.z.pp;0b];.z.pp;{[x]}];
+  .z.pp:{[x]$[(`$"X-Grafana-Org-Id")in key last x;.z.m.zpp x;.z.m.prevpp x]};
+  .z.m.prevph:$[@[{value x;1b};`.z.ph;0b];.z.ph;{[x]}];
+  .z.ph:{[x]$[(`$"X-Grafana-Org-Id")in key last x;"HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n";.z.m.prevph x]};
+  .z.m.wired:1b;
+  };
+
+init:{[deps]
+  / wire the logging dependency and any configuration, then install the handlers
+  / deps - `log!enlist logdict or `log`config!(logdict;configdict)
+  /   logdict    - `info`warn`error!({[c;m]};{[c;m]};{[c;m]}) - required
+  /   configdict - optional overrides for `timecol`sym`timebackdate`ticks`del
+  / examples:
+  /   grafana.init[enlist[`log]!enlist mylog]
+  /   grafana.init[`log`config!(mylog;enlist[`ticks]!enlist 500)]
+  logdict:$[99h=type deps;$[(`log in key deps)and not(::)~deps`log;deps`log;()!()];()!()];
+  if[not count logdict;'"di.grafana: log dependency is required; pass `info`warn`error functions - see di.log for a default implementation"];
+  .z.m.loginfo:logdict`info;
+  .z.m.logwarn:logdict`warn;
+  .z.m.logerr:logdict`error;
+  / overlay any supplied configuration onto the defaults
+  config:$[99h=type deps;$[(`config in key deps)and not(::)~deps`config;deps`config;()!()];()!()];
+  if[count config;
+    vars:`timecol`sym`timebackdate`ticks`del inter key config;
+    (.Q.dd[.z.M]each vars)set'config vars;
+   ];
+  / install the http handlers once, preserving any existing definitions
+  if[not`wired in key .z.M;sethandlers[]];
+  .z.m.loginfo[`grafana;"initialised grafana json datasource adaptor"];
+  };
+
+getconfig:{
+  / return the currently active configuration
+  :`timecol`sym`timebackdate`ticks`del!(timecol;sym;timebackdate;ticks;del);
+  };
--- a/di/grafana/grafana.q
+++ b/di/grafana/grafana.q
@@ -4,7 +4,8 @@
 / the grafana dropdowns and the /query endpoint returns timeseries or table data
 
 / -----------------------------------------------------------------------------
-/ configuration - defaults applied here, overridden via the config dict in init
+/ configuration - load-time defaults; overridden via the deps dict in init and
+/ read back through .z.m at every call site
 / -----------------------------------------------------------------------------
 
 / name of the time column used for timeseries queries
@@ -17,6 +18,8 @@ timebackdate:2D;
 ticks:1000;
 / delimiter separating the arguments within a query target
 del:".";
+/ allow f. function targets to evaluate arbitrary q (off by default - see grafana.md)
+allowfunctions:0b;
 
 / json type for each kdb datatype, keyed by .Q.t character
 types:.Q.t!`array`boolean,(3#`null),(5#`number),11#`string;
@@ -55,19 +58,19 @@ query:{[rqt]
 search:{[rqt]
   / build the grafana dropdown options from the tables available in the process
   tabs:tables[];
-  symtabs:tabs where sym in'cols each tabs;
-  timetabs:tabs where timecol in'cols each tabs;
+  symtabs:tabs where .z.m.sym in'cols each tabs;
+  timetabs:tabs where .z.m.timecol in'cols each tabs;
   rsp:string tabs;
   if[count timetabs;
     rsp,:s1:prefix["t";string timetabs];
     rsp,:s2:prefix["g";string timetabs];
     / suffix the numeric columns for the graph and other panel options
-    rsp,:raze(s2,'del),/:'c1:string {cols[x] where`number=types(0!meta x)`t}each timetabs;
-    rsp,:raze(prefix["o";string timetabs],'del),/:'c1;
+    rsp,:raze(s2,'.z.m.del),/:'c1:string {cols[x] where`number=types(0!meta x)`t}each timetabs;
+    rsp,:raze(prefix["o";string timetabs],'.z.m.del),/:'c1;
     if[count symtabs;
       / suffix the distinct syms for the timeseries and other panel options
-      rsp,:raze(s1,'del),/:'c2:string each finddistinctsyms'[timetabs];
-      rsp,:raze(prefix["o";string timetabs],'del),/:'{x[0]cross del,'string finddistinctsyms x 1}each(enlist each c1),'timetabs;
+      rsp,:raze(s1,'.z.m.del),/:'c2:string each finddistinctsyms'[timetabs];
+      rsp,:raze(prefix["o";string timetabs],'.z.m.del),/:'{x[0]cross .z.m.del,'string finddistinctsyms x 1}each(enlist each c1),'timetabs;
      ];
    ];
   :.h.hy[`json].j.j rsp;
@@ -75,12 +78,12 @@ search:{[rqt]
 
 finddistinctsyms:{[x]
   / distinct syms seen in table x within the configured lookback window
-  :?[x;enlist(>;timecol;(-;.z.p;timebackdate));1b;{x!x}enlist sym]sym;
+  :?[x;enlist(>;.z.m.timecol;(-;.z.p;.z.m.timebackdate));1b;{x!x}enlist .z.m.sym] .z.m.sym;
   };
 
 prefix:{[c;s]
   / prefix string c and the delimiter to each string in s
-  :(c,del),/:s;
+  :(c,.z.m.del),/:s;
   };
 
 / -----------------------------------------------------------------------------
@@ -89,13 +92,13 @@ prefix:{[c;s]
 
 diskvals:{[x]
   / last `ticks` rows of an on-disk partitioned table
-  c:(count[x]-ticks)+til ticks;
+  c:(count[x]-.z.m.ticks)+til .z.m.ticks;
   :get'[.Q.ind[x;c]];
   };
 
 memvals:{[x]
   / last `ticks` rows of an in-memory table
-  :get'[?[x;enlist(within;`i;count[x]-ticks,0);0b;()]];
+  :get'[?[x;enlist(within;`i;count[x]-.z.m.ticks,0);0b;()]];
   };
 
 catchvals:{[x]
@@ -109,10 +112,28 @@ catchvals:{[x]
 
 istype:{[targ;char]
   / test whether the target is prefixed with char followed by the delimiter
-  :(char,del)~2#targ;
+  :(char,.z.m.del)~2#targ;
   };
 isfunc:istype[;"f"];
 istab:istype[;"t"];
+
+resolvetab:{[t]
+  / resolve a table-name symbol to its unkeyed table, rejecting anything that is
+  / not a known table - the name is looked up, never evaluated as code
+  if[not t in tables[];
+    .z.m.log[`error][`grafana;"unknown table: ",string t];
+    '"di.grafana: unknown table ",string t];
+  :0!value t;
+  };
+
+evalfunc:{[s]
+  / evaluate an f. function-target expression - gated behind allowfunctions
+  / because it executes arbitrary q (disabled by default)
+  if[not .z.m.allowfunctions;
+    .z.m.log[`error][`grafana;"function target rejected; allowfunctions is disabled"];
+    '"di.grafana: function targets are disabled (set allowfunctions to enable)"];
+  :value s;
+  };
 
 / -----------------------------------------------------------------------------
 / building json responses
@@ -127,16 +148,15 @@ tbfunc:{[rqt]
   / process a table request and return the json datasource table response
   rqt:raze rqt[`targets]`target;
   symname:0b;
-  / strip the type prefix: f.t.func drops 4, f.func drops 2, t.tab leaves the
-  / table name plus an optional sym
-  rqt:0!value $[isfunc[rqt]&istab 2_rqt;4_rqt;
-                isfunc rqt;2_rqt;
-                istab rqt;[rqt:`$del vs rqt;if[2<count rqt;symname:rqt 2];rqt 1];
-                rqt];
+  / f. targets evaluate a q expression (opt-in); t./bare targets resolve a known
+  / table by name and are never evaluated as code
+  rqt:$[isfunc rqt;0!evalfunc $[istab 2_rqt;4_rqt;2_rqt];
+        istab rqt;[parts:`$.z.m.del vs rqt;if[2<count parts;symname:parts 2];resolvetab parts 1];
+        resolvetab(`$rqt)];
   colname:cols rqt;
   coltype:types(0!meta rqt)`t;
   / filter to a single sym if one was supplied in the target
-  if[-11h=type symname;rqt:?[rqt;enlist(=;sym;enlist symname);0b;()]];
+  if[-11h=type symname;rqt:?[rqt;enlist(=;.z.m.sym;enlist symname);0b;()]];
   :tabresponse[colname;coltype;rqt];
   };
 
@@ -144,18 +164,20 @@ tsfunc:{[x]
   / process a timeseries request and route to the correct panel/sym builder
   targ:raze x[`targets]`target;
   / split the target into its delimited arguments
-  numargs:count args:$[isfunc targ;(0;1+targ?del)cut targ:2_targ;`$del vs targ];
-  tyargs:$[10h=abs type args 0;`$1#;]args 0;
-  coln:cols rqt:0!value args 1;
+  numargs:count args:$[isfunc targ;(0;1+targ?.z.m.del)cut targ:2_targ;`$.z.m.del vs targ];
+  / panel-type indicator (g/t/o) as a symbol - args 0 is a char string on the f.
+  / path (take its first char) and already a symbol otherwise
+  tyargs:$[10h=abs type args 0;`$1#args 0;args 0];
+  coln:cols rqt:$[isfunc targ;0!evalfunc args 1;resolvetab args 1];
   / convert a timestamp to milliseconds since the unix epoch
   mil:{floor epoch+(`long$x)%1000000};
   / ensure the time column is a timestamp
-  if["p"<>meta[rqt][timecol;`t];rqt:@[rqt;timecol;+;.z.D]];
+  if["p"<>meta[rqt][.z.m.timecol;`t];rqt:@[rqt;.z.m.timecol;+;.z.D]];
   / restrict to the time range requested by grafana
   range:"P"$-1_'x[`range]`from`to;
-  rqt:?[rqt;enlist(within;timecol;range);0b;()];
+  rqt:?[rqt;enlist(within;.z.m.timecol;range);0b;()];
   / add the milliseconds-since-epoch column grafana expects
-  rqt:@[rqt;`msec;:;mil rqt timecol];
+  rqt:@[rqt;`msec;:;mil rqt .z.m.timecol];
   / dispatch on the number of arguments and the panel type
   $[(2<numargs)and`g~tyargs;graphsym[args 2;rqt];
     (2<numargs)and`t~tyargs;tablesym[coln;rqt;args 2];
@@ -163,7 +185,7 @@ tsfunc:{[x]
     (2=numargs)and`t~tyargs;tablenosym[coln;rqt];
     (4=numargs)and`o~tyargs;othersym[args;rqt];
     (3=numargs)and`o~tyargs;othernosym[args 2;rqt];
-    (2=numargs)and`o~tyargs;othernosym[coln except timecol;rqt];
+    (2=numargs)and`o~tyargs;othernosym[coln except .z.m.timecol;rqt];
     `$"Wrong input"]
   };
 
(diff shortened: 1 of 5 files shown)


Olly99999:
## Summary
Extracts the `.grafana` namespace from TorQ's `grafana.q` into a standalone kdb-x module: **`di.grafana`**. The module turns a kdb+ process into a [Grafana SimPod JSON datasource](https://github.com/simPod/GrafanaJsonDatasource): it installs HTTP handlers that answer Grafana's `/search`, `/query` and connection-test requests with the JSON the plugin expects, serving the process's tables as timeseries and table panels. It satisfies the `di.*` module contract: dependency injection via `init`, a clean exported API, and no hard dependencies beyond an injected logger.

## Background
TorQ's `grafana.q` implements the Grafana JSON-datasource adaptor as a `.grafana` namespace that hooks `.z.pp`/`.z.ph` through TorQ's `.dotz` handler manager. This PR is part of the broader TorQ → kdb-x modularisation effort: make each functional area independently loadable, testable, and injectable, removing the dependency on TorQ's global process framework.

## Changes

### New files

| File | Description |
| --- | --- |
| `di/grafana/grafana.q` | Core implementation — `init`, `getconfig`, the `.z.pp`/`.z.ph` wrappers, and the search/query request builders |
| `di/grafana/init.q` | Module entry point — loads `grafana.q` and declares the export list |
| `di/grafana/test.csv` | k4unit test suite (32 tests) |
| `di/grafana/grafana.md` | Module README |

| Aspect | TorQ `.grafana` | `di.grafana` |
| --- | --- | --- |
| **Handlers** | `.dotz.set` / `.dotz.getcommand` (TorQ handler manager) | `init` closure-wraps `.z.pp`/`.z.ph` directly, preserving any pre-existing handler so non-Grafana traffic falls through untouched. Follows the `di.timer` precedent (`.z.ts`); no `di.handlers` module exists yet to inject. Verified by a fall-through test. |
| Logging | None | Required injected `log` dependency (`info`/`warn`/`error`, `{[ctx;msg]}`). Request dispatch trapped; failures logged under the `grafana` context |
| Config | `@[value;` `.grafana.x` `;default]` globals | Module-local defaults + optional config dict in `init` (`timecol`, `sym`, `timebackdate`, `ticks`, `del`) |
| Global state / namespace | `\d .grafana` | Flat module namespace; mutable state on `.z.m`/`.z.M` |
| Dependency validation | N/A | `init` validates a non-null log dep and errors with a `di.grafana`-prefixed message if absent |

## Exported API

```q
grafana:use`di.grafana

grafana.init[deps]      / wire dependencies + config, install handlers; required before use
grafana.getconfig[]     / return the currently active configuration dict
```

Everything else is driven by inbound Grafana HTTP requests via the installed handlers, so it is intentionally not exported.

`deps` forms:

```q
grafana.init[enlist[`log]!enlist logdep]              / logger only, default config
grafana.init[`log`config!(logdep;`ticks!500)]         / logger + config overrides
```

## Query target syntax

`/search` emits encoded metric strings that `/query` parses back; `del` (default `.`) separates the arguments:

| Target | Meaning |
| --- | --- |
| `t.<table>` | table panel, whole table |
| `t.<table>.<sym>` | table panel filtered to one sym |
| `g.<table>` | graph, one series per numeric column |
| `g.<table>.<col>` | graph, one series per sym for a column |
| `o.<table>...` | "other" panels (stat/gauge) |
| `f....` | target is a function call rather than a table |

## Test coverage

Run via k4unit: `k4unit.moduletest` `` `di.grafana ``

| Area | Tests |
| --- | --- |
| Export surface | only `init` + `getconfig` exported |
| `init` — dependency validation | rejects `::` and a null log dep |
| `init` — wiring | `wired` flag set; handlers installed |
| Config | defaults applied; override takes; `getconfig` reflects it |
| Parsing helpers | `isfunc` / `istab` / `istype` / `prefix` |
| Data fetch | `finddistinctsyms`, `memvals`, `catchvals` |
| `/query` table | `tbfunc` returns the columns/rows/table schema |
| `/query` timeseries | `tsfunc` returns target/datapoints numeric series |
| Handler routing | `/search` POST + GET test-connection via live handlers; non-Grafana POST falls through to the pre-existing handler |

**Total: 32 tests, all passing.**

## Notes

- `di.grafana` has no hard module dependencies — only the injected `log` dict is required.
- `di.log` satisfies the log contract out of the box:

```q
```

- The handler wrapper preserves any pre-existing `.z.pp`/`.z.ph`, so adding the module to a process that already serves HTTP does not break existing endpoints.
- The `/annotations` endpoint returns a "not yet implemented" marker, matching the TorQ original.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  rqt:.j.k r 1;
+  .z.m.log[`info][`grafana;"received ",(r 0)," request"];
+  handler:$["query"~r 0;query;"search"~r 0;search;annotation];
+  :.[handler;enlist rqt;{[e].z.m.log[`error][`grafana;"failed to process request: ",e];'e}];
+  };
+
+annotation:{[rqt]
+  / annotation endpoint is not yet implemented
+  .z.m.log[`warn][`grafana;"annotation url not yet implemented"];
+  :`$"Annotation url nyi";
```
DI-Software-Engineering, line 51: In `search`, the variable `rsp` is built by appending string table names (`string tabs`), then later appending prefixed targets (`prefix["t";string timetabs]` etc). The initial `string tabs` entries have no prefix, so they appear as raw table names alongside `t.*`/`g.*` targets. This is almost certainly wrong — bare table names cannot be dispatched by the prefix-based routing in `tbfunc`/`tsfunc`, and `zpp` → `query` will fall through to `annotation` for them. Remove the bare `string tabs` line or replace it with appropriate prefixed entries.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+query:{[rqt]
+  / dispatch a /query request to the timeseries or table builder by target type
+  rqtype:raze rqt[`targets]`type;
+  :.h.hy[`json]$[rqtype~"timeserie";tsfunc rqt;tbfunc rqt];
+  };
```
DI-Software-Engineering, line 66: `finddistinctsyms` uses a functional select that references the module-level names `timecol`, `sym`, and `timebackdate` by value (not via `.z.m.*`). After `init` overrides these via `.z.m.timecol` etc., the local namespace variables are never updated, so `finddistinctsyms` (and other functions referencing `timecol`, `sym`, `ticks`, `timebackdate`, `del` by bare name) will always use the initial default values instead of the configured ones. The `init` function assigns `deps` values to `.z.m.timecol`, `.z.m.sym`, etc. but the rest of the code reads the unqualified names, which resolve to the module-level constants set at load time.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  :.h.hy[`json].j.j rsp;
+  };
+
+finddistinctsyms:{[x]
+  / distinct syms seen in table x within the configured lookback window
+  :?[x;enlist(>;timecol;(-;.z.p;timebackdate));1b;{x!x}enlist sym]sym;
+  };
+
+prefix:{[c;s]
+  / prefix string c and the delimiter to each string in s
```
DI-Software-Engineering, line 95: `memvals` uses `?[x;enlist(within;`i;count[x]-ticks,0);0b;()]`. The `within` check `\`i within count[x]-ticks,0` is evaluated as `\`i within (count[x]-ticks),0` — meaning the lower bound is `count[x]-ticks` and the upper bound is `0`, which is always an empty range for any positive `count`. The intended expression is likely `(count[x]-ticks),count[x]-1` or `?[x;();0b;()]` followed by a tail take. This means `memvals` silently returns an empty table for every in-memory table.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+/ target parsing helpers
+/ -----------------------------------------------------------------------------
+
+istype:{[targ;char]
+  / test whether the target is prefixed with char followed by the delimiter
+  :(char,del)~2#targ;
+  };
+isfunc:istype[;"f"];
+istab:istype[;"t"];
+
```
DI-Software-Engineering, line 129: In `tsfunc`, `args 1` is used as a table reference (passed to `value`) when the target is a plain `g.table.col` string. When `isfunc` is false, `args` is `` `$del vs targ `` — a symbol list. `args 1` is thus a symbol like `` `trade ``, and `value` on a symbol executes it, which is fine for a global. But when `isfunc` is true, `args` is a 2-element list from `(0;1+targ?del)cut targ:2_targ` — a pair of *strings*, not symbols. `args 1` would then be a string, and `0!value args 1` would attempt `value` on a string, which *evaluates* it as q code rather than looking up a table. This is an injection risk: a crafted Grafana target of the form `f.<q expression>` will be executed by the server.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+/ -----------------------------------------------------------------------------
```
DI-Software-Engineering, line 148: In `tsfunc`, when `isfunc` is false the branch is `` `$del vs targ `` giving a symbol list; `args 0` is then a symbol like `` `g ``. The subsequent check `10h=abs type args 0` tests for a char/string — a symbol has type -11h, so `tyargs` will be assigned via `` `$1# `` applied to a symbol, yielding `` `g `` (symbol), not `` `g `` (symbol from char). The downstream `$[(2<numargs)and\`g~tyargs;...]` comparisons use `` `g `` (symbol) for `tyargs` and `` `g `` (symbol) for the literal, so they coincidentally match — but when `isfunc` is true `args 0` is a string (type 10h), `tyargs` becomes `` `$1#"g" `` = `` `g ``, so both branches produce the same result. The logic is confusing and fragile but the `isfunc` path hides the mismatch. Flag for clarity: the `10h=abs type args 0` branch should be the `isfunc`-false (symbol) branch, not the string branch; the condition is inverted relative to the comment.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  / split the target into its delimited arguments
+  numargs:count args:$[isfunc targ;(0;1+targ?del)cut targ:2_targ;`$del vs targ];
+  tyargs:$[10h=abs type args 0;`$1#;]args 0;
+  coln:cols rqt:0!value args 1;
+  / convert a timestamp to milliseconds since the unix epoch
+  mil:{floor epoch+(`long$x)%1000000};
+  / ensure the time column is a timestamp
+  if["p"<>meta[rqt][timecol;`t];rqt:@[rqt;timecol;+;.z.D]];
+  / restrict to the time range requested by grafana
+  range:"P"$-1_'x[`range]`from`to;
```
DI-Software-Engineering: `sethandlers` stores the previous `.z.pp`/`.z.ph` into `.z.m.prevpp`/`.z.m.prevph` and then defines the new `.z.pp`/`.z.ph` as plain lambda closures that reference `.z.m.prevpp` etc. by name. However, `.z.m.wired` is set at the end of `sethandlers`, but `init` checks the *unqualified* `wired` variable — which is always `0b` (it was set at module load and never updated via `.z.m.wired`). So `sethandlers` is called on every `init` invocation, re-wrapping the handlers each time and creating a chain of growing closures. The guard `if[not wired;sethandlers[]]` should read `if[not .z.m.wired;sethandlers[]]`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+ticks:1000;
+/ delimiter separating the arguments within a query target
+del:".";
+
+/ json type for each kdb datatype, keyed by .Q.t character
+types:.Q.t!`array`boolean,(3#`null),(5#`number),11#`string;
+/ milliseconds between 1970.01.01 and 2000.01.01
+epoch:946684800000;
+
+/ internal flag - set true once the http handlers have been installed
```
DI-Software-Engineering, line 29: In `zpp`, `r:(0;n?" ")cut n:first x` splits the raw HTTP request line at the first space. For a `POST /query HTTP/1.1` request, `r 0` will be the method (e.g. `"POST"`), not the path (e.g. `"query"`). The handler dispatch `$["query"~r 0;...]` then compares the method string to the path, and will never match, so every request falls through to `annotation`. The split should extract the path component (between the first and second spaces), not the method.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+wired:0b;
+
+/ -----------------------------------------------------------------------------
+/ request handling
+/ -----------------------------------------------------------------------------
+
+zpp:{[x]
+  / parse a grafana http post request and dispatch to the matching handler
+  / cut at the first whitespace to isolate the api url from any function params
+  r:(0;n?" ")cut n:first x;
```
DI-Software-Engineering, line 41: In `query`, `rqtype:raze rqt[\`targets]\`type` retrieves the `type` field from all targets and razes them. If multiple targets are present with different types (mixed timeserie/table), the raze produces a concatenated string and the comparison `rqtype~"timeserie"` will be false, silently sending a mixed request to `tbfunc`. The dispatch logic does not handle multi-target requests correctly.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-ticks)+til ticks;
+  :get'[.Q.ind[x;c]];
+  };
+
+memvals:{[x]
+  / last `ticks` rows of an in-memory table
+  :get'[?[x;enlist(within;`i;count[x]-ticks,0);0b;()]];
+  };
+
```
DI-Software-Engineering, line 113: In `tbfunc`, `rqt:raze rqt[\`targets]\`target` retrieves all targets and razes them. If `targets` contains multiple entries (Grafana can send multiple targets per query), the strings are concatenated by `raze`, producing a garbled target name. All of `tsfunc` and `tbfunc` assume a single target; multi-target requests are silently corrupted rather than rejected or iterated.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  / cut at the first whitespace to isolate the api url from any function params
+  r:(0;n?" ")cut n:first x;
+  rqt:.j.k r 1;
+  .z.m.log[`info][`grafana;"received ",(r 0)," request"];
+  handler:$["query"~r 0;query;"search"~r 0;search;annotation];
+  :.[handler;enlist rqt;{[e].z.m.log[`error][`grafana;"failed to process request: ",e];'e}];
+  };
+
+annotation:{[rqt]
+  / annotation endpoint is not yet implemented
```
DI-Software-Engineering, line 49: In `query`, `rqtype:raze rqt[`targets]`type` extracts the `type` field from the targets list. When multiple targets are present, `raze` concatenates all their type strings together (e.g. `"timeserietimeserie"` or `"timetimeserie"`), so the subsequent `~"timeserie"` test will silently fail for more than one target. Each target should be dispatched individually, or the type should be taken from a single/first target rather than razing all.

DI-Software-Engineering, line 49: `query` dispatches the entire `rqt` (with all targets) to either `tsfunc` or `tbfunc`, but those functions extract `raze rqt[`targets]`target` which collapses all targets into one string. Grafana `/query` requests regularly contain multiple targets; only the first (or a garbled concatenation) will be processed, silently dropping the others. The response should loop over targets.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+/ -----------------------------------------------------------------------------
+/ fetching the last n ticks
+/ -----------------------------------------------------------------------------
+
+diskvals:{[x]
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-ticks)+til ticks;
+  :get'[.Q.ind[x;c]];
```
DI-Software-Engineering, line 106: `memvals` uses `` `i `` as a virtual column in a functional select: `?[x;enlist(within;`i;count[x]-ticks,0);0b;()]`. The lower bound `count[x]-ticks` can be negative when the table has fewer rows than `ticks`, producing an invalid range (e.g. `-500` to `0`). This will return zero rows instead of the whole table. The lower bound should be clamped to `0`, e.g. `0|count[x]-ticks`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+/ building json responses
+/ -----------------------------------------------------------------------------
+
+tabresponse:{[colname;coltype;rqt]
+  / build a table response in the json datasource schema
+  :.j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table);
+  };
+
+tbfunc:{[rqt]
+  / process a table request and return the json datasource table response
```
DI-Software-Engineering, line 158: In `tsfunc`, `rqt:0!value args 1` evaluates `args 1` as a kdb+ value. When `isfunc` is false the target is split by `del` and `args` is a symbol list (`` `$del vs targ ``), so `args 1` is a plain symbol and `value` on it looks up a variable — that is acceptable. But when `isfunc` is true the target is a string (the function-call string is kept as a character list), so `args 1` is a character-list substring that `value` will *execute as code*. A Grafana user (or attacker) controlling the `target` field can inject arbitrary q code. The function path should use a safe lookup rather than `value` on a user-supplied string.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  / table name plus an optional sym
+  rqt:0!value $[isfunc[rqt]&istab 2_rqt;4_rqt;
+                isfunc rqt;2_rqt;
+                istab rqt;[rqt:`$del vs rqt;if[2<count rqt;symname:rqt 2];rqt 1];
+                rqt];
+  colname:cols rqt;
+  coltype:types(0!meta rqt)`t;
+  / filter to a single sym if one was supplied in the target
+  if[-11h=type symname;rqt:?[rqt;enlist(=;sym;enlist symname);0b;()]];
+  :tabresponse[colname;coltype;rqt];
```
DI-Software-Engineering, line 170: `tsfunc` converts the time column with `rqt:@[rqt;timecol;+;.z.D]` when the column type is not `"p"`. `.z.D` is a `date`, not a `timespan`/`timestamp`. Adding a date to most time types (e.g. `time` which is a timespan) uses kdb+ temporal arithmetic, but the result type and correctness depend on the original column type. If the column is already a `datetime` or `timestamp`-like type other than `"p"`, this coercion may produce unexpected results or a type error.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+tsfunc:{[x]
+  / process a timeseries request and route to the correct panel/sym builder
```
DI-Software-Engineering, line 174: `range:"P"$-1_'x[`range]`from`to` drops the last character of the Grafana ISO-8601 datetime strings (the `Z`) before casting to timestamp. This fragile manual trim assumes the strings always end in exactly one character to drop. A malformed or differently formatted timestamp from Grafana (e.g. with milliseconds `"2024-01-01T00:00:00.000Z"`) will still have its last character dropped but may then fail to parse or parse incorrectly. Use a robust parse or strip the trailing `Z` explicitly with `ssr`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  $[(2<numargs)and`g~tyargs;graphsym[args 2;rqt];
+    (2<numargs)and`t~tyargs;tablesym[coln;rqt;args 2];
+    (2=numargs)and`g~tyargs;graphnosym[coln;rqt];
+    (2=numargs)and`t~tyargs;tablenosym[coln;rqt];
+    (4=numargs)and`o~tyargs;othersym[args;rqt];
+    (3=numargs)and`o~tyargs;othernosym[args 2;rqt];
+    (2=numargs)and`o~tyargs;othernosym[coln except timecol;rqt];
+    `$"Wrong input"]
+  };
+
```
DI-Software-Engineering, line 202: `buildnosym` uses `value each ?[x;();0b;z!z]` to extract column values, where `z` is a two-element list `(colname;`msec)`. Selecting two columns returns a table with two columns; `value each` on a table returns a list of two column vectors (one per row of `value`). The datapoints structure Grafana expects is a list of `(value;milliseconds)` pairs. `flip value flip` (as used in `othersym`/`graphsym`) would be correct here; `value each` on the table produces per-column vectors, not per-row pairs.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+tablenosym:{[coln;rqt]
+  / timeseries response for a table panel with no sym split
+  coltype:types -1_(0!meta rqt)`t;
+  :tabresponse[coln;coltype;rqt];
+  };
+
+othersym:{[args;rqt]
+  / timeseries response for a non-specific panel returning a single sym's data
+  outcol:args[2],`msec;
```
DI-Software-Engineering, line 234: In `tbfunc`, when the target is a plain table name (not prefixed with `f.` or `t.`), the code falls into the `istab` branch anyway because `istab rqt` is tested last and returns false, so `rqt` is passed through unchanged as a string. The final else branch `rqt` would be the raw target string. Then `0!value rqt` attempts to evaluate the raw string. A target like `"trade"` (no prefix) would not match `isfunc` or `istab`, so `rqt` remains the original string and `value` executes it — which is another code-injection vector for non-prefixed targets. The else/fallthrough case should reject unknown prefixes rather than calling `value` on the raw input.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+
+sethandlers:{
+  / wrap any existing .z.pp/.z.ph so grafana requests are intercepted and every
+  / other request falls through to the original handler
+  .z.m.prevpp:$[@[{value x;1b};`.z.pp;0b];.z.pp;{[x]}];
+  .z.pp:{[x]$[(`$"X-Grafana-Org-Id")in key last x;.z.m.zpp x;.z.m.prevpp x]};
+  .z.m.prevph:$[@[{value x;1b};`.z.ph;0b];.z.ph;{[x]}];
+  .z.ph:{[x]$[(`$"X-Grafana-Org-Id")in key last x;"HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n";.z.m.prevph x]};
+  .z.m.wired:1b;
+  };
```
DI-Software-Engineering, line 284: `init` guards against re-installing handlers with `if[not wired;sethandlers[]]`, reading the module-level `wired` variable. But `sethandlers` writes `.z.m.wired` (the namespaced copy) while the guard reads the bare `wired` (the top-level copy). After the first `init` call `.z.m.wired` is `1b` but `wired` remains `0b`, so `sethandlers` will be called on every subsequent `init`, re-wrapping `.z.pp`/`.z.ph` each time and creating a growing chain of nested closures.

di/grafana/grafana.q
```diff
@@ -0,0 +1,277 @@
...
+  / timeseries response for a graph panel returning each sym's data
+  syms:`$string ?[rqt;();1b;{x!x}enlist sym]sym;
+  outcol:colname,`msec;
+  build:{[outcol;rqt;x;y]data:flip value flip?[rqt;enlist(=;sym;enlist y);0b;outcol!outcol];:x,`target`datapoints!(y;data)};
+  :.j.j build[outcol;rqt]\[();syms];
+  };
+
+tablesym:{[coln;rqt;symname]
+  / timeseries response for a table panel filtered to a single sym
+  coltype:types -1_(0!meta rqt)`t;
```
DI-Software-Engineering, line 249: In `sethandlers`, the `.z.pp` handler detects Grafana requests by checking `(`$"X-Grafana-Org-Id") in key last x`. The kdb+ HTTP POST handler receives a two-element list: the body string and the header dictionary. If the request arrives with no headers (or `x` is not a two-element list), `last x` is the body string and `key` on a string returns indices (`0 1 2 ...`), so the symbol will never be found — this is benign. However, if `x` is a plain string (e.g. non-HTTP invocation), `last x` is a character and `key` errors. More critically, a plain `GET` from `.z.ph` where headers are absent will cause `.z.pp` to error rather than fall through. The handler should guard `type last x` before calling `key`.

DI-Software-Engineering:
## DIReview Summary

**2 critical** | **8 warning(s)** | **0 suggestion(s)**

di/grafana/grafana.q
Olly99999: Bare table names **are** dispatched, not dropped.

`tbfunc` has an explicit else branch that resolves an unprefixed target via `value`, so a raw table name is served as a whole-table **table panel** — intended behaviour, and a faithful port of the TorQ original:

```q
rqt:0!value $[isfunc[rqt]&istab 2_rqt; 4_rqt;   / f.t.xxx
              isfunc rqt;                2_rqt;   / f.xxx
              istab rqt;  [rqt:`$del vs rqt; ...; rqt 1];   / t.xxx
              rqt];                               / else: bare name -> value
```

Also, `zpp` routes on the request **URL** (`query`/`search`), not on the target:

```q
handler:$["query"~r 0;query;"search"~r 0;search;annotation];
```

so `annotation` is the fallback for an unknown *endpoint* (e.g. `/annotations`), never for an unknown *target*. A metric request arrives as `query …`, so it always reaches `query` → `tbfunc`/`tsfunc`; a target can't fall through to `annotation`.

Verified locally — a bare `"trade"` target returns a normal `{columns, rows, type: "table"}` response:

```q
q).m.di.0grafana.tbfunc enlist[`targets]!enlist `target`type!("trade";"table")
"[{\"columns\":[{\"text\":\"time\",\"type\":\"string\"},...],\"rows\":[...],\"type\":\"table\"}]"
```

Keeping `string tabs` as-is. The one true edge — a bare name dropped into a **graph** panel — returns `` `$"Wrong input" ``, which is benign (not an `annotation` fallthrough, not a crash) and matches the TorQ original.

di/grafana/grafana.q
Olly99999: The configured values were in fact being applied (no stale defaults), but the comment correctly spotted that the bare-name reads made that non-obvious, so I've made them explicit.

**Why it was already correct:** there's no `\d` namespace switch here. The module loads into `.m.di.0grafana`, and `.z.m` is an alias for that *same* namespace. So the load-time default, `init`'s `.z.m.timecol:` write, and a bare `timecol` read inside a function all resolve to the **same slot** (`.m.di.0grafana.timecol`) — `init`'s writes were visible to the bare reads. (Same pattern `di.timer` uses for `enabled`/`nextruntime`.)

**What changed:** every config read — `timecol`, `sym`, `timebackdate`, `ticks`, `del`, and the `wired` flag — now goes through `.z.m.` explicitly, matching the "read mutable state via `.z.m`" convention and removing the ambiguity. The load-time bare defaults remain as the initial values of those `.z.m` slots. For example:

```q
finddistinctsyms:{[x]
  :?[x;enlist(>;.z.m.timecol;(-;.z.p;.z.m.timebackdate));1b;{x!x}enlist .z.m.sym] .z.m.sym;
  };
```

**Verified** — config overrides flow through end-to-end (40/40 k4unit tests pass), e.g.:

```q
q)grafana.init[`log`ticks!(logdep;5)]
q)grafana.getconfig[][`ticks]
5      / getconfig reads .z.m.ticks
```

di/grafana/grafana.q
Olly99999: This parses the other way round — q is strictly right-to-left, so `,0` binds to `ticks` first, not to `(count[x]-ticks)`.

`count[x]-.z.m.ticks,0` evaluates as:

1. `.z.m.ticks,0` → `(ticks;0)`  (e.g. `1000 0`)
2. `count[x] - (ticks;0)` → `(count[x]-ticks; count[x])`

So the range is `(count-ticks; count)` — lower bound `count-ticks`, **upper bound `count[x]`** (not `0`). `i within (count-ticks; count)` therefore selects the last `ticks` rows (or all rows when `ticks ≥ count`). This is the unchanged TorQ original.

Worth noting your suggested `(count[x]-ticks; count[x]-1)` yields the same rows — the code just uses `count[x]` as a harmless (non-existent) upper index.

Verified — the test suite asserts `10 = count memvals` on a 10-row table, which passes (it would be `1` if the upper bound were `0`):

```q
q)10 - 1000,0          / shows ,0 binds first -> upper bound is count, not 0
-990 10
q)trade:([]time:.z.p-til 10;sym:10#`a;price:10?1f)
q)count .m.di.0grafana.memvals trade
10
```

So `memvals` returns the rows, not an empty table.

di/grafana/grafana.q
Olly99999: This was a real RCE surface (inherited from the TorQ original, where `f.` is an "evaluate this expression" target and bare targets were `value`d too). Fixed, secure by default:

- **Table targets (`t.`/`g.`/`o.`/bare) are now resolved by name, not evaluated.** They're looked up as symbols against `tables[]` (`resolvetab`); an unknown name is rejected. No target string reaches `value` through these paths.
- **`f.` function targets are gated behind a new `allowfunctions` flag (default `0b`).** With it off they're rejected (and not evaluated); with it on, trusted deployments keep the eval feature. A `Security` section in the README documents this and the port trust boundary.

Added k4unit coverage: `f.` rejected-and-not-evaluated by default, unknown-table rejected, and `f.` evaluating only after opting in.

```q
q)grafana.init[enlist[`log]!enlist logdep]               / allowfunctions off
q).m.di.0grafana.tbfunc enlist[`targets]!enlist `target`type!("f.1+1";"table")
'di.grafana: function targets are disabled (set allowfunctions to enable)
```

di/grafana/grafana.q
Olly99999: The condition isn't inverted, and the two branches aren't coincidental. `args 0` has a different *shape* in the two paths, and the test normalises both to the same symbol:

- `isfunc` **false** → `args` is `` `$del vs targ`` → **symbols**, so `args 0` is `` `g`` (type `-11h`). `abs -11h = 11h`, so `10h=11h` is **false** → identity branch → `tyargs` = `` `g``. (The `` `$1# `` branch is *not* taken for a symbol.)
- `isfunc` **true** → `args` is a `cut` of a string → **char strings**, so `args 0` is `"g…"` (type `10h`) → **true** → `` `$1#`` → `` `$"g"`` = `` `g``.

So `10h=abs type args 0` reads as "is this the string (f.) form? then take its first char as a symbol; otherwise it's already a symbol, leave it" — correct as written, no hidden mismatch.

Agreed the point-free `$[…;`$1#;]args 0` is hard to read, so I've made it explicit (behaviour identical):

```q
/ panel-type indicator (g/t/o) as a symbol - args 0 is a char string on the f.
/ path (take its first char) and already a symbol otherwise
tyargs:$[10h=abs type args 0;`$1#args 0;args 0];
```

di/grafana/grafana.q
Olly99999: The guard already reads `.z.m.wired` on the current branch (in a follow-up commit), so the suggested change is in.

On the premise though: `wired` and `.z.m.wired` were never separate variables. `.z.m` is the loader's alias for the module's own namespace (`.m.di.0grafana`), so the load-time `wired:0b` and `sethandlers`'s `.z.m.wired:1b` write the **same slot** — `init`'s bare `wired` read saw the update, and `sethandlers` ran only once (no growing closure chain). Verifiable:

```q
q).m.di.0grafana.wired                         / 0b  (slot from load-time `wired:0b`)
q)grafana.init[enlist[`log]!enlist logdep]
q).m.di.0grafana.wired                         / 1b  (sethandlers' .z.m.wired:1b wrote it)
q)p1:.m.di.0grafana.prevpp; grafana.init[enlist[`log]!enlist logdep]; .m.di.0grafana.prevpp~p1
1b                                             / re-init does not re-wrap
```

Switched it (and the other config reads) to explicit `.z.m.` regardless — same slot, clearer intent.

di/grafana/grafana.q
Olly99999: Verified against a real HTTP POST — kdb does not pass `.z.pp` the raw request line. It passes a 2-item list `(requeststring; headers)`, and `requeststring` is the **endpoint followed by the body** (method and leading `/` already stripped), so `r 0` is the path, not the method:

```q
q)\p 8903
q).z.pp:{`CAP set x; "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\nok"}
q)/ shell: curl -s -X POST -d '{"targets":[{"target":"t.trade","type":"table"}]}' http://localhost:8903/query
q)count CAP
2
q)first CAP
"query {\"targets\":[{\"target\":\"t.trade\",\"type\":\"table\"}]}"
```

So `(0;n?" ")cut first x` gives `r 0 ~ "query"`, `$["query"~r 0;…]` matches, and the request goes to `query` — never `annotation`. I've added a clarifying comment on the split, and a regression test that posts a real-shape `"query {body}"` (body built with `.j.j`) through the installed `.z.pp` and asserts a table response.

di/grafana/grafana.q
Olly99999: Multi-target requests were mishandled. The old `query` did `raze rqt[`targets]`type` / `raze ...`target` across *all* targets, so two queries got concatenated into one nonsense target. `query` now dispatches **each target by its own type** and merges the per-target JSON arrays.

Before — a two-query panel:
```q
q)t:`target`type!("t.trade";"table")
q).m.di.0grafana.query enlist[`targets]!enlist (t;t)
'di.grafana: unknown table tradet        / "t.trade" + "t.trade" razed into one
```
After:
```q
q).m.di.0grafana.query enlist[`targets]!enlist (t;t)
"[ {...table...}, {...table...} ]"       / each target built, then merged
```

Implementation notes:
- The change is confined to `query`; `tsfunc`/`tbfunc` and the leaf builders are untouched, and single-target output is unchanged.
- It iterates targets by index and normalises the `targets` field (Grafana sends an array; a lone target can arrive as an object), so list/table/dict shapes are all handled.
- Per-target arrays are merged as **strings**, not via a `.j.k`/`.j.j` round-trip — re-parsing would float-ify the epoch-ms longs and risk precision loss on timestamps.
- Added a k4unit test that runs a two-target request through `query` and asserts both tables are present in the response.

di/grafana/grafana.q
Olly99999: Same root cause as the multi-target comment above, and fixed by the same change. `query` now iterates the targets and calls `tbfunc`/`tsfunc` **once per target** with a single-target request, so neither builder ever sees more than one target — the `raze rqt[`targets]`target` here flattens exactly that one target (intentional), it no longer concatenates across targets.

```q
q)t:`target`type!("t.trade";"table")
q).m.di.0grafana.query enlist[`targets]!enlist (t;t)   / two targets
"[ {...table...}, {...table...} ]"                      / iterated + merged, not corrupted
```

`tsfunc`/`tbfunc` are intentionally single-target; the per-target dispatch and merge live in `query`. Covered by the new multi-target k4unit test.

di/grafana/grafana.q
Olly99999: This is the same multi-target issue, and the fix does exactly what you suggest: each target is dispatched individually. The `rqtype:raze rqt[`targets]`type` line is gone; `query` now iterates the targets and dispatches each on its own `type` (`tsfunc` for `"timeserie"`, `tbfunc` otherwise), then merges the per-target results. So mixed-type and multi-target requests are handled correctly, not silently dropped.

Covered by the new multi-target k4unit test.

Olly99999: Fixed - `query` now loops over the targets.

di/grafana/grafana.q
Olly99999: This is the same parse as the earlier `memvals` note. q is right-to-left, so `,0` binds to `ticks`, giving the range `(count-ticks; count)`, not `(count-ticks; 0)`:

```q
q)10 - 1000,0      / ,0 binds first -> (count-ticks; count)
-990 10
```

The **upper bound is `count`**, not `0`. When `count < ticks` the lower bound is negative, but `i within (negative; count)` still selects every row (all `i` in `0..count-1`), i.e. the whole table — the intended "fewer than `ticks` rows ⇒ return all" behaviour, not zero rows.

di/grafana/grafana.q
Olly99999: Fixed same hardening as the `f.` injection in `tbfunc`. `tsfunc`'s resolution is now:

```q
coln:cols rqt:$[isfunc targ;0!evalfunc args 1;resolvetab args 1];
```

- Non-`f.` targets: `args 1` is a symbol, resolved via `resolvetab`, which looks the name up against `tables[]` — never `value` on a string.
- `f.` targets: go through `evalfunc`, gated behind the `allowfunctions` flag (default `0b`). With it off they're rejected; only an explicit opt-in enables expression evaluation.

So a crafted `f.<q>` target can't execute by default. Same mechanism as the table-builder fix; k4unit covers `f.` rejected-by-default, unknown-table rejected, and `f.` evaluating only when enabled.

di/grafana/grafana.q
Olly99999: Verified — for the time-column types Grafana data realistically uses, this is correct and doesn't error:

```q
q)v:09:00:00.000          / a `time` value (type -19h)
q)v + .z.D
2026.06.29D09:00:00.000000000      / correct timestamp
```

`time + date` and `timespan + date` both yield the right timestamp, and a `timestamp` column skips the branch (`"p"<>...`). So `p`/`t`/`n` are all handled correctly.

The one type where `+.z.D` would double-count is `datetime` (`"z"`) — but that's deprecated and banned by the repo style guide, so it's out of scope. Leaving as-is.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
+/ grafana json datasource adaptor for kdb-x
```
DI-Software-Engineering, line 1: All top-level variable assignments (`timecol`, `sym`, `timebackdate`, `ticks`, `del`, `allowfunctions`, `types`, `epoch`, `wired`) and all function definitions are in the root (`.`) namespace rather than inside `\d .di.grafana` / `\d .` guards. TorQ convention requires every file to open and close a named namespace. These symbols will pollute the global namespace.

di/grafana/grafana.q
Olly99999: The `-1_` works for the format Grafana actually sends.

Confirmed empirically: `"P"$` rejects the trailing `Z` (`"P"$"…000Z"` → `0Np`) but parses the ISO string fine once it's gone (`"P"$"…000"` → a timestamp). So stripping the `Z` is necessary and correct — the risk is only that blindly dropping the *last char* corrupts any string that doesn't end in `Z`. Now it strips conditionally:

```q
range:"P"${$["Z"=last x;-1_x;x]}each x[`range]`from`to;
```

So a `Z`-terminated ISO string (what Grafana sends) parses as before, and anything else is passed to `"P"$` intact rather than truncated. Also switched the timeseries tests to a real ISO-8601 range (`"2000-01-01T00:00:00.000Z"…`) so they exercise the actual `"P"$`-on-ISO path, which the previous kdb-format test strings didn't.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+/ implements the simpod json datasource api so a grafana server can query
+/ kdb+ tables and timeseries data over http - the /search endpoint populates
+/ the grafana dropdowns and the /query endpoint returns timeseries or table data
+
+/ -----------------------------------------------------------------------------
+/ configuration - load-time defaults; overridden via the deps dict in init and
+/ read back through .z.m at every call site
```
DI-Software-Engineering, line 8: All top-level variables (`timecol`, `sym`, `timebackdate`, `ticks`, `del`, `allowfunctions`, `types`, `epoch`, `wired`) are declared outside any namespace. TorQ convention requires every file to open with `\d .ns` and close with `\d .`. Without this, these names pollute the root namespace and can collide with other modules.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+/ number of ticks to return for table requests
+ticks:1000;
+/ delimiter separating the arguments within a query target
+del:".";
+/ allow f. function targets to evaluate arbitrary q (off by default - see grafana.md)
+allowfunctions:0b;
+
+/ json type for each kdb datatype, keyed by .Q.t character
+types:.Q.t!`array`boolean,(3#`null),(5#`number),11#`string;
+/ milliseconds between 1970.01.01 and 2000.01.01
```
DI-Software-Engineering, line 26: Single-slash `/` is used throughout as an inline comment character, but in q a single `/` at the start of a line opens a multiline comment block. All inline comments should use `//` to avoid silently commenting out subsequent code if a `/` happens to be the first non-whitespace token on a line.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+
+annotation:{[rqt]
+  / annotation endpoint is not yet implemented
+  .z.m.log[`warn][`grafana;"annotation url not yet implemented"];
+  :`$"Annotation url nyi";
+  };
+
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
```
DI-Software-Engineering, line 56: In `zpp`, the request string is split at the first space with `(0;n?" ")cut n:first x`. When the body contains no space (e.g. an empty-body `search` request), `n?" "` returns `count n`, so the cut produces `(n;n)` — both elements are the full string, and `r 1` is the entire request string rather than the body. `.j.k` then fails or silently misparses. Fix: use `(" "vs r:first x)` and take the tail, or guard when no space is found.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+/ internal flag - set true once the http handlers have been installed
+wired:0b;
+
+/ -----------------------------------------------------------------------------
+/ request handling
+/ -----------------------------------------------------------------------------
+
+zpp:{[x]
+  / parse a grafana http post request and dispatch to the matching handler
+  / kdb passes .z.pp (requeststring;headers); requeststring is "<endpoint> <body>"
```
DI-Software-Engineering, line 38: `r:(0;n?" ")cut n:first x` cuts the entire request string `n` at the first space, giving `(endpoint; body)`. However, `r 1` is then passed directly to `.j.k`. If the HTTP body itself contains no space (e.g. an empty body `"{}"`), `n?" "` returns `count n` and the cut produces `(full-string;"")`, so `.j.k ""` will throw a parse error. The body extraction should be done more robustly, e.g. `(1+n?" ")_n`.

di/grafana/grafana.q
DI-Software-Engineering, line 56: The `zpp` handler calls `.z.m.log` before the handler guard, meaning it will crash with a null-dereference error if `.z.pp` is ever called before `init` has been run (`.z.m.log` is not set at that point). The handler should be installed only after `init`, but the code path that sets `.z.pp` (via `sethandlers`) is reachable only after `init`, so this is safe only if `sethandlers` is never called independently — but there is no enforcement preventing that.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  / (method and leading / already stripped), so cut at the first space gives
+  / (endpoint;body) - e.g. "query {...}" -> ("query";"{...}")
+  r:(0;n?" ")cut n:first x;
+  rqt:.j.k r 1;
```
DI-Software-Engineering, line 42: The `annotation` function returns `` :`$"Annotation url nyi" `` — a symbol — but all other handlers return an HTTP response string (the output of `.h.hy`). Grafana (or `.z.pp`) will send this symbol back as the HTTP response body, likely producing a garbled or empty HTTP response rather than a valid JSON reply.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+/ -----------------------------------------------------------------------------
+/ fetching the last n ticks
+/ -----------------------------------------------------------------------------
+
+diskvals:{[x]
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-.z.m.ticks)+til .z.m.ticks;
+  :get'[.Q.ind[x;c]];
+  };
+
```
DI-Software-Engineering, line 108: In `memvals`, the `within` constraint is `(within;\`i;count[x]-.z.m.ticks,0)`. The right-hand side `` count[x]-.z.m.ticks,0 `` is parsed right-to-left as `` count[x]-(.z.m.ticks,0) ``, i.e. subtracting the two-element list `(ticks;0)` from `count[x]`, not creating the intended `(count[x]-ticks; count[x]-0)` range. This produces wrong or type-error results. Fix to `((count[x]-.z.m.ticks);count[x]-1)` or equivalent.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+annotation:{[rqt]
+  / annotation endpoint is not yet implemented
+  .z.m.log[`warn][`grafana;"annotation url not yet implemented"];
+  :`$"Annotation url nyi";
+  };
+
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
```
DI-Software-Engineering, line 57: In `search`, `types` is indexed with `(0!meta x)\`t` which returns a char vector. The `types` dict is keyed by `.Q.t` characters. Any column whose type character is not in `.Q.t` (e.g. temporal types not listed, or extended types) will produce a null/missing lookup result without error, silently dropping that column from the numeric-column filter. This is latent but worth guarding.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  :get'[?[x;enlist(within;`i;count[x]-.z.m.ticks,0);0b;()]];
+  };
+
+catchvals:{[x]
+  / try the on-disk path first, fall back to the in-memory path on error
+  :@[diskvals;x;{[x;y]memvals x}[x]];
+  };
+
+/ -----------------------------------------------------------------------------
+/ target parsing helpers
```
DI-Software-Engineering, line 120: `diskvals` computes indices as `c:(count[x]-.z.m.ticks)+til .z.m.ticks`. If the table has fewer rows than `ticks`, this produces negative starting indices. `.Q.ind` with out-of-range indices will error or return wrong data. Guard with `c:(neg .z.m.ticks sublist til count x)` or clamp the start to 0.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
+  / a single target may arrive as a bare dict; normalise to a list of targets
+  tgts:$[99h=type tgts;enlist tgts;tgts];
+  / iterate by index so a list or a table of targets is handled the same way
+  build:{[rqt;tgts;i]t:tgts i;r:@[rqt;`targets;:;t];$[t[`type]~"timeserie";tsfunc r;tbfunc r]};
+  / strip the [ ] from each per-target array, drop empties, re-wrap as one array
```
DI-Software-Engineering, line 62: `finddistinctsyms` uses a functional select with `1b` (distinct) on a single grouped column. The result is a table; indexing it with `.z.m.sym` (a symbol) on a table returns the column as a list, which is correct only because the distinct flag produces a single-key grouped result. However if the sym column is absent or the table is empty this will signal a type/rank error that is not caught here, potentially crashing the `/search` handler.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  :.j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table);
+  };
+
+tbfunc:{[rqt]
+  / process a table request and return the json datasource table response
+  rqt:raze rqt[`targets]`target;
+  symname:0b;
+  / f. targets evaluate a q expression (opt-in); t./bare targets resolve a known
+  / table by name and are never evaluated as code
+  rqt:$[isfunc rqt;0!evalfunc $[istab 2_rqt;4_rqt;2_rqt];
```
DI-Software-Engineering, line 163: In `tsfunc`, when `isfunc targ` is true the `args` variable is assigned as `(0;1+targ?.z.m.del)cut targ:2_targ`. On the `f.` path `args 1` is expected to be the q expression to evaluate; but if the delimiter does not appear in the stripped target, `targ?.z.m.del` returns `count targ` and `args 1` is an empty string, causing a silent empty-string `value` call rather than a clear error.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  };
+
+search:{[rqt]
+  / build the grafana dropdown options from the tables available in the process
+  tabs:tables[];
+  symtabs:tabs where .z.m.sym in'cols each tabs;
+  timetabs:tabs where .z.m.timecol in'cols each tabs;
+  rsp:string tabs;
+  if[count timetabs;
+    rsp,:s1:prefix["t";string timetabs];
```
DI-Software-Engineering, line 75: `memvals` uses `` ?[x;enlist(within;`i;count[x]-.z.m.ticks,0);0b;()] ``. The right-hand side of `within` must be a 2-element list `(lo;hi)`, so `count[x]-.z.m.ticks,0` produces a 2-element list only when ticks < count. When `count x <= ticks` (table has fewer rows than ticks), `count[x]-ticks` is zero or negative, giving a range like `(-5;0)` — `i` will never satisfy `within` and the result is an empty table even though all rows should be returned.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  rqt:raze rqt[`targets]`target;
+  symname:0b;
+  / f. targets evaluate a q expression (opt-in); t./bare targets resolve a known
+  / table by name and are never evaluated as code
+  rqt:$[isfunc rqt;0!evalfunc $[istab 2_rqt;4_rqt;2_rqt];
+        istab rqt;[parts:`$.z.m.del vs rqt;if[2<count parts;symname:parts 2];resolvetab parts 1];
+        resolvetab(`$rqt)];
+  colname:cols rqt;
+  coltype:types(0!meta rqt)`t;
+  / filter to a single sym if one was supplied in the target
```
DI-Software-Engineering, line 168: In `tsfunc`, the expression `if["p"<>meta[rqt][.z.m.timecol;\`t];rqt:@[rqt;.z.m.timecol;+;.z.D]]` adds `.z.D` (a date atom) to the time column to coerce it to a timestamp, but adding a date to a time gives a datetime, not a timestamp (`p`). This will not produce a timestamp column, and the subsequent `within` comparison on the timestamp range will type-error. Use `` `timestamp$rqt .z.m.timecol `` or an explicit cast instead.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  };
+
+finddistinctsyms:{[x]
+  / distinct syms seen in table x within the configured lookback window
+  :?[x;enlist(>;.z.m.timecol;(-;.z.p;.z.m.timebackdate));1b;{x!x}enlist .z.m.sym] .z.m.sym;
+  };
+
+prefix:{[c;s]
+  / prefix string c and the delimiter to each string in s
+  :(c,.z.m.del),/:s;
```
DI-Software-Engineering, line 96: `tbfunc` strips the prefix from an `f.` target with `$[istab 2_rqt;4_rqt;2_rqt]`: it drops the first 2 chars (`f.`), checks if the remainder starts with `t.`, then drops 4 chars from the original `rqt` or 2. But `rqt` at this point is still the full original string (e.g. `"f.t.trade"`). Dropping 4 chars from `"f.t.trade"` gives `"trade"`, while `evalfunc` receives `"trade"` — that is actually correct for `"f.t.trade"`. However for `"f.g.trade"` (not `istab`) it drops only 2 chars giving `"g.trade"`, which `evalfunc` then calls `value` on, evaluating the symbol `` `g `` applied to `trade` rather than evaluating a function expression. The intent of `f.` targets is to eval arbitrary q; stripping the inner type prefix before evaluation is fragile and undocumented.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+
+/ -----------------------------------------------------------------------------
+/ fetching the last n ticks
+/ -----------------------------------------------------------------------------
+
+diskvals:{[x]
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-.z.m.ticks)+til .z.m.ticks;
+  :get'[.Q.ind[x;c]];
+  };
```
DI-Software-Engineering, line 107: In `tsfunc`, `args` is built as `` `$.z.m.del vs targ `` for non-`f.` targets, giving a symbol list. Then `args 1` is passed to `resolvetab` which already expects a symbol — fine. But for the `f.` path, `args` is `(0;1+targ?.z.m.del)cut targ` — a list of strings — and `args 1` (a string) is passed to `evalfunc`. `evalfunc` calls `value` on a string, which evaluates it as q code. If the expression in `args 1` itself contains the delimiter character the cut at `targ?.z.m.del` (first occurrence only) will silently truncate the expression. This is a correctness issue for any `f.` target whose q expression contains a `.`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  rqt:$[isfunc rqt;0!evalfunc $[istab 2_rqt;4_rqt;2_rqt];
+        istab rqt;[parts:`$.z.m.del vs rqt;if[2<count parts;symname:parts 2];resolvetab parts 1];
+        resolvetab(`$rqt)];
+  colname:cols rqt;
+  coltype:types(0!meta rqt)`t;
+  / filter to a single sym if one was supplied in the target
+  if[-11h=type symname;rqt:?[rqt;enlist(=;.z.m.sym;enlist symname);0b;()]];
+  :tabresponse[colname;coltype;rqt];
+  };
+
```
DI-Software-Engineering, line 172: `range` is a two-element list of ISO-8601 strings that is passed directly as the `within` right argument. `within` expects a two-element list of values of the same type as the column (timestamps), not strings. The ISO-8601 strings are never converted to timestamps, so the `within` filter will always type-error or silently fail. Strings must be cast: e.g. `` "P"$each range ``.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  / last `ticks` rows of an on-disk partitioned table
+  c:(count[x]-.z.m.ticks)+til .z.m.ticks;
+  :get'[.Q.ind[x;c]];
+  };
+
+memvals:{[x]
+  / last `ticks` rows of an in-memory table
+  :get'[?[x;enlist(within;`i;count[x]-.z.m.ticks,0);0b;()]];
+  };
+
```
DI-Software-Engineering, line 113: `if["p"<>meta[rqt][.z.m.timecol;\`t];rqt:@[rqt;.z.m.timecol;+;.z.D]]` attempts to coerce non-timestamp time columns by adding `.z.D` (today as a date). Adding a date to an integer or other non-temporal type is a type error. The intent is presumably to promote a time (`t`) column to a timestamp, but the operation is wrong for most non-timestamp types and will signal an error rather than coercing cleanly.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  / table by name and are never evaluated as code
+  rqt:$[isfunc rqt;0!evalfunc $[istab 2_rqt;4_rqt;2_rqt];
+        istab rqt;[parts:`$.z.m.del vs rqt;if[2<count parts;symname:parts 2];resolvetab parts 1];
+        resolvetab(`$rqt)];
+  colname:cols rqt;
+  coltype:types(0!meta rqt)`t;
+  / filter to a single sym if one was supplied in the target
+  if[-11h=type symname;rqt:?[rqt;enlist(=;.z.m.sym;enlist symname);0b;()]];
+  :tabresponse[colname;coltype;rqt];
+  };
```
DI-Software-Engineering, line 171: The `range` variable shadows `"P"$` in the next line: `range:"P"${...}each x[\`range]\`from\`to` — the `"P"$` prefix IS present and produces a two-element timestamp list, so range filtering is actually typed correctly. However the variable name `range` also shadows the outer q built-in `range` if used, which is a latent hazard. (Withdraw if not a real issue.) — Actually reading again: `range:"P"${$["Z"=last x;-1_x;x]}each x[\`range]\`from\`to` does cast to timestamp. Retract the previous finding about strings — the cast is present. This finding is withdrawn.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  / restrict to the time range requested by grafana - grafana sends iso-8601 utc;
+  / strip a trailing Z only if present rather than blindly dropping the last char
+  range:"P"${$["Z"=last x;-1_x;x]}each x[`range]`from`to;
+  rqt:?[rqt;enlist(within;.z.m.timecol;range);0b;()];
+  / add the milliseconds-since-epoch column grafana expects
+  rqt:@[rqt;`msec;:;mil rqt .z.m.timecol];
+  / dispatch on the number of arguments and the panel type
+  $[(2<numargs)and`g~tyargs;graphsym[args 2;rqt];
+    (2<numargs)and`t~tyargs;tablesym[coln;rqt;args 2];
+    (2=numargs)and`g~tyargs;graphnosym[coln;rqt];
```
DI-Software-Engineering, line 195: In `graphnosym`, the time column is removed with `-1_coln where \`number=types(0!meta rqt)\`t`. This removes the *last* column of the filtered list, not necessarily the time column. If the numeric columns do not include the time column at the end, a data column is silently dropped. If the time column (type `p`) is not `number` in the types dict (it is typed as `string` per the `types` table), it is not in `coln` at all, and `-1_` removes a legitimate data column. Fix: explicitly exclude the time column: `coln except enlist .z.m.timecol`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  };
+
+memvals:{[x]
+  / last `ticks` rows of an in-memory table
+  :get'[?[x;enlist(within;`i;count[x]-.z.m.ticks,0);0b;()]];
+  };
+
+catchvals:{[x]
+  / try the on-disk path first, fall back to the in-memory path on error
+  :@[diskvals;x;{[x;y]memvals x}[x]];
```
DI-Software-Engineering, line 116: `range` is built as `"P"$` applied to each ISO-8601 string after stripping a trailing `Z`. `"P"$` parses a kdb+ timestamp literal (format `yyyy.mm.ddTHH:MM:SS.nnnnnnnnn`), but Grafana sends ISO-8601 UTC strings like `"2000-01-01T00:00:00.000Z"` (dashes, not dots). `"P"$` will fail to parse this format and return a null timestamp, making the `within` filter on line 117 silently pass or reject all rows, producing wrong query results.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+    .z.m.log[`error][`grafana;"unknown table: ",string t];
+    '"di.grafana: unknown table ",string t];
+  :0!value t;
+  };
+
+evalfunc:{[s]
+  / evaluate an f. function-target expression - gated behind allowfunctions
+  / because it executes arbitrary q (disabled by default)
+  if[not .z.m.allowfunctions;
+    .z.m.log[`error][`grafana;"function target rejected; allowfunctions is disabled"];
```
DI-Software-Engineering, line 143: `graphnosym` does `coln:-1_coln where \`number=types(0!meta rqt)\`t` — it drops the last column of the numeric columns. The intent is presumably to exclude `msec` (which was just appended), but `msec` is not in `cols rqt` at the point `coln` is computed (it is added to `rqt` afterwards in `tsfunc`). If the last numeric column happens to be a genuine data column it will be silently dropped from the graph.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  / append one column's datapoints to the response accumulator
+  :y,`target`datapoints!(z 0;value each ?[x;();0b;z!z]);
+  };
+
+nosymresponse:{[rqt;colname]
+  / build a no-sym datapoints response across all requested columns
+  :.j.j buildnosym[rqt]\[();colname];
+  };
+
+othernosym:{[coln;rqt]
```
DI-Software-Engineering, line 213: In `graphsym`, `syms` is computed by `` `$string ?[rqt;();1b;{x!x}enlist .z.m.sym] .z.m.sym ``. The functional select with `1b` (distinct) returns a table; indexing it with `.z.m.sym` returns a column list, and `` `$string `` casts it. However the distinct-select already returns symbols, so `` `$string `` round-trips correctly only for simple symbols; symbols containing special characters will be mangled. More critically, if `rqt` has already been time-filtered and contains zero rows, this produces an empty list and the scan over `syms` silently returns `()`, yielding an empty JSON array with no error.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  mil:{floor epoch+(`long$x)%1000000};
+  / ensure the time column is a timestamp
+  if["p"<>meta[rqt][.z.m.timecol;`t];rqt:@[rqt;.z.m.timecol;+;.z.D]];
+  / restrict to the time range requested by grafana - grafana sends iso-8601 utc;
+  / strip a trailing Z only if present rather than blindly dropping the last char
+  range:"P"${$["Z"=last x;-1_x;x]}each x[`range]`from`to;
+  rqt:?[rqt;enlist(within;.z.m.timecol;range);0b;()];
+  / add the milliseconds-since-epoch column grafana expects
+  rqt:@[rqt;`msec;:;mil rqt .z.m.timecol];
+  / dispatch on the number of arguments and the panel type
```
DI-Software-Engineering, line 192: `sethandlers` assigns directly to `.z.pp` and `.z.ph` (`.z.*` handler assignment) rather than using a TorQ-approved mechanism such as `.dotz.set`. Per TorQ conventions, direct `.z.*` assignment should be avoided in favour of `.dotz.set`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+
+othersym:{[args;rqt]
+  / timeseries response for a non-specific panel returning a single sym's data
+  outcol:args[2],`msec;
+  data:flip value flip?[rqt;enlist(=;.z.m.sym;enlist args 3);0b;outcol!outcol];
+  :.j.j enlist`target`datapoints!(args 3;data);
+  };
+
+graphsym:{[colname;rqt]
+  / timeseries response for a graph panel returning each sym's data
```
DI-Software-Engineering, line 240: In `normlog`, the detection test is `any \`getlvl\`sinks\`fmts in key logdict`. A plain `info`warn`error` dict that happens to also contain a `getlvl`, `sinks`, or `fmts` key would be misidentified as a kx.log instance and its functions would be wrapped into a different calling convention, silently breaking the logger. The detection should require *all three* keys (use `all`) or check a more unique key combination.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  targ:raze x[`targets]`target;
+  / split the target into its delimited arguments
+  numargs:count args:$[isfunc targ;(0;1+targ?.z.m.del)cut targ:2_targ;`$.z.m.del vs targ];
+  / panel-type indicator (g/t/o) as a symbol - args 0 is a char string on the f.
+  / path (take its first char) and already a symbol otherwise
+  tyargs:$[10h=abs type args 0;`$1#args 0;args 0];
+  coln:cols rqt:$[isfunc targ;0!evalfunc args 1;resolvetab args 1];
+  / convert a timestamp to milliseconds since the unix epoch
+  mil:{floor epoch+(`long$x)%1000000};
+  / ensure the time column is a timestamp
```
DI-Software-Engineering, line 184: `normlog` detects a `kx.log` instance by checking for `getlvl`/`sinks`/`fmts` keys, then reads `logdict\`info` and `logdict\`warn` etc. If the passed dict does not have those keys (e.g. a partial kx.log object) the extraction will return a null/missing value without error, and the resulting wrapped function will fail at call time with a confusing error rather than a clear validation message.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+  syms:`$string ?[rqt;();1b;{x!x}enlist .z.m.sym] .z.m.sym;
+  outcol:colname,`msec;
+  build:{[outcol;rqt;x;y]data:flip value flip?[rqt;enlist(=;.z.m.sym;enlist y);0b;outcol!outcol];:x,`target`datapoints!(y;data)};
+  :.j.j build[outcol;rqt]\[();syms];
+  };
+
+tablesym:{[coln;rqt;symname]
+  / timeseries response for a table panel filtered to a single sym
+  coltype:types -1_(0!meta rqt)`t;
+  rqt:?[rqt;enlist(=;.z.m.sym;enlist symname);0b;()];
```
DI-Software-Engineering, line 250: In `sethandlers`, `.z.m.prevpp` is set to `{[x]}` (a no-op returning the generic null) as the fallback when `.z.pp` is not defined. A no-op POST handler returns null/nothing to Grafana's fallthrough clients, silently dropping requests instead of returning an HTTP error response. If no pre-existing `.z.pp` exists, a 501 or empty response should be returned explicitly.

di/grafana/grafana.q
```diff
@@ -0,0 +1,311 @@
...
+/ -----------------------------------------------------------------------------
+
+istype:{[targ;char]
+  / test whether the target is prefixed with char followed by the delimiter
+  :(char,.z.m.del)~2#targ;
+  };
+isfunc:istype[;"f"];
+istab:istype[;"t"];
+
+resolvetab:{[t]
```
DI-Software-Engineering, line 130: `buildnosym` calls `value each ?[x;();0b;z!z]` — `value` on a table row (a dict) returns its values as a list, not a `(value;msec)` datapoints pair. For Grafana's timeseries datapoints schema each entry must be `(value; epoch_ms)`. Calling `value each` on each row-dict gives a list of all column values per row, not the expected `(numericValue; msecValue)` pair, producing malformed datapoints JSON.

DI-Software-Engineering:
## DIReview Summary

**4 critical** | **11 warning(s)** | **0 suggestion(s)**

di/grafana/grafana.q
DI-Software-Engineering, line 143: In `tbfunc`, when the target is an `f.` path, the code is `0!evalfunc $[istab 2_rqt;4_rqt;2_rqt]`. Here `rqt` is the *target string* (renamed from the outer `rqt`), not the request dict. `istab 2_rqt` checks if the expression after stripping `f.` starts with `t.`, which is intended to detect `f.t.<expr>`. But `4_rqt` drops the first 4 chars of the original string, not of the already-stripped `2_rqt`, so for `f.t.expr` the evaluated expression is `t.expr` (still has the `t.` prefix) rather than `expr`. The strip should be `4_rqt` vs `2_rqt` from the already-`2_`-stripped string: use `s:2_rqt; evalfunc $[istab s;2_s;s]`.

di/grafana/grafana.q
DI-Software-Engineering, line 1: All top-level variables (`timecol`, `sym`, `timebackdate`, `ticks`, `del`, `allowfunctions`, `types`, `epoch`, `wired`) and functions are defined outside any namespace (`\d` is never set). Per TorQ convention every file should open with `\d .ns` and close with `\d .`, otherwise all symbols pollute the root namespace and risk clashing with user data.

DI-Software-Engineering, line 1: All inline comments use a single `/` (e.g. `/ grafana json datasource adaptor`). In q, a `/` at the start of a line (or after whitespace at the start) opens a multiline comment block until the next `\`. Single-line comments should use `//`. This means the file is likely interpreted with large sections as multi-line comments depending on indentation, silently skipping code.

DI-Software-Engineering:
## DIReview Summary

**4 critical** | **11 warning(s)** | **0 suggestion(s)**

di/grafana/grafana.q
Olly99999: `value each` here is correct — `each` over a *table* iterates **rows**, not columns, so `value each ?[…]` applies `value` to each row dict and yields that row's `(value; ms)` pair. It's equivalent to `flip value flip`:

```q
q)t:([]price:1 2 3f;msec:10 20 30)
q)(value each t) ~ flip value flip t
1b
```

Confirmed in the actual response — `datapoints` comes out as `[[value, ms], …]` pairs:
```
[{"target":"price","datapoints":[[86.71843,1782306589550],[79.02208,1782306588550],...]}]
```
So no change needed; `nosymresponse`/`buildnosym` already emit the per-row pair structure Grafana expects.

DI-Software-Engineering:
All changed files were excluded by the skip blocklist — nothing to review.

di/grafana/grafana.q
Olly99999: Already addressed by the same hardening as the `f.` injection. The bare/`else` branch no longer calls `value` on the raw target — it resolves by name:

```q
rqt:$[isfunc rqt;0!evalfunc …;
      istab rqt;[…;resolvetab parts 1];
      resolvetab(`$rqt)];          / bare: `$rqt -> symbol -> lookup in tables[]
```

`resolvetab` turns the bare string into a symbol and rejects anything not in `tables[]`, so `"trade"` resolves but `"1+1"` / `"system …"` are rejected, never executed. Added a test that proves a bare q-expression target isn't evaluated:

```q
q).test.pwned:0b
q)@[.m.di.0grafana.tbfunc;enlist[`targets]!enlist (enlist[`target]!enlist ".test.pwned:1b");{x}]
q).test.pwned          / still 0b — the string was rejected, not run
0b
```

di/grafana/grafana.q
Olly99999: Switched the guard to explicit `.z.m.wired` regardless

di/grafana/grafana.q
Olly99999: Added a typed guard and factored the duplicated header check into one predicate:

```q
isgrafana:{[x]$[99h=type h:last x;(`$"X-Grafana-Org-Id")in key h;0b]};
.z.pp:{[x]$[.z.m.isgrafana x;.z.m.zpp x;.z.m.prevpp x]};
.z.ph:{[x]$[.z.m.isgrafana x;"HTTP/1.1 200 OK\r\nConnection: close\r\n\r\n";.z.m.prevph x]};
```

Now any non-dict `last x` (malformed or non-HTTP invocation) falls through to the pre-existing handler instead of erroring on `key`. Behaviour is unchanged for real requests (dict header → same check). Added a test for the malformed-arg fall-through.

di/grafana/grafana.q
```diff
@@ -0,0 +1,317 @@
...
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
+  / a single target may arrive as a bare dict; normalise to a list of targets
+  tgts:$[99h=type tgts;enlist tgts;tgts];
+  / iterate by index so a list or a table of targets is handled the same way
+  build:{[rqt;tgts;i]t:tgts i;r:@[rqt;`targets;:;t];$[t[`type]~"timeserie";tsfunc r;tbfunc r]};
+  / strip the [ ] from each per-target array, drop empties, re-wrap as one array
+  inners:{1_-1_x}each build[rqt;tgts]each til count tgts;
```
DI-Software-Engineering, line 63: The `inners` filtering step (`inners where 0<count each inners`) drops empty per-target results silently. A target that returns an empty timeseries or table is legitimately empty (no data in the time range), and Grafana expects an entry in the array even for empty results (otherwise panel series disappear unexpectedly for targets that have no data). The empty-result case should be preserved rather than filtered out.

di/grafana/grafana.q
```diff
@@ -0,0 +1,317 @@
...
+
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
+  / a single target may arrive as a bare dict; normalise to a list of targets
+  tgts:$[99h=type tgts;enlist tgts;tgts];
+  / iterate by index so a list or a table of targets is handled the same way
+  build:{[rqt;tgts;i]t:tgts i;r:@[rqt;`targets;:;t];$[t[`type]~"timeserie";tsfunc r;tbfunc r]};
+  / strip the [ ] from each per-target array, drop empties, re-wrap as one array
```
DI-Software-Engineering, line 62: Each per-target result from `tsfunc`/`tbfunc` is already a JSON string (produced by `.h.hy[\`json]`). Stripping the first and last character with `1_-1_x` assumes the string always starts with `[` and ends with `]`. If either builder ever returns an object `{}` or an error string, this silently corrupts or truncates the response rather than erroring. No guard is applied before the slice.

di/grafana/grafana.q
```diff
@@ -0,0 +1,317 @@
...
+  / annotation endpoint is not yet implemented
+  .z.m.log[`warn][`grafana;"annotation url not yet implemented"];
+  :`$"Annotation url nyi";
+  };
+
+query:{[rqt]
+  / dispatch each target to the timeseries or table builder by its own type and
+  / merge the per-target json arrays - grafana can send several targets at once
+  tgts:rqt`targets;
+  / a single target may arrive as a bare dict; normalise to a list of targets
```
DI-Software-Engineering, line 58: `rqt[\`targets]` after `.j.k` on a Grafana JSON body is always a list of dicts (JSON array), so `99h=type tgts` will be `0b` and the `enlist` branch is never taken. However, the real problem is that `tgts i` where `tgts` is a list of dicts returns the dict at index `i`, but `build` then calls `@[rqt;\`targets;:;t]` replacing `targets` with a bare dict. `tsfunc`/`tbfunc` downstream access `rqt[\`targets]` as a list and then index into it — replacing it with a bare dict breaks that indexing. The per-target dispatch should pass the single target directly (not re-inserted as `targets`) or wrap it in `enlist`.

di/grafana/grafana.q
```diff
@@ -0,0 +1,317 @@
...
+  / plain {[c;m]} log dicts (info`warn`error only) pass through unchanged
+  $[any `getlvl`sinks`fmts in key logdict;
+    `info`warn`error!(
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`info;];
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`warn;];
+      {[fn;c;m] fn[string[c],": ",m]}[logdict`error;]);
+    logdict]
+  };
+
+isgrafana:{[x]
```
DI-Software-Engineering, line 270: `isgrafana` uses `last x` to get the headers. For a normal `.z.pp` argument `x` is `(requeststring;headerdict)`, so `last x` is the header dict — correct. But for `.z.ph`, Grafana's GET health-check, the argument shape may differ across kdb versions (it can be a single string or a 2-element list). The `.z.ph` handler now delegates to `isgrafana` which calls `last x`; if `x` is a plain string `last x` returns the last character (a char), `99h=type` is `0b`, and the guard correctly falls through — but this is a fragile coincidence rather than an intentional check. The original code also had this latent issue, but widening `isgrafana` to both handlers without documenting the different arg shapes increases the risk.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **3 warning(s)** | **1 suggestion(s)**

### Suggestions
- **di/grafana/grafana.q:192** — The `range` fix conditionally strips a trailing `Z` with `$["Z"=last x;-1_x;x]`. `"P"$` in kdb+ accepts ISO-8601 timestamps with a trailing `Z` directly (it treats `Z` as UTC); the strip is harmless but unnecessary and introduces a branch. If future Grafana versions use a `+00:00` suffix instead of `Z`, the cast will fail silently and produce a null timestamp, causing an empty result rather than an error.

PR #107, opened 2026-06-22
Asyncdispatch

Patrickguest1998:
Here is async dispatch. Added tests that weren't stubbed and got rid of unnecessary functions in init.q

di.asyncdispatch (new, standalone)
The async multi-process query coordinator extracted from the gateway. Handles queuing queries, dispatching to available servers, collecting async results, applying join functions, timeout management, and error handling.

Standalone value: any process that needs scatter-gather across multiple backends.

Key data structures:

queryqueue   / pending and active queries
results      / per-query result accumulator (queryid -> servertype -> result)
clients      / connected client tracking
Pluggable hooks (the consumer sets these, not hardcoded):

formatresponse   / {[status;sync;result] ...} customise result delivery
availableservers / {[excludeinuse] ...} customise server selection
getnextqueryid   / {[] ...} customise scheduling (default: FIFO)
Injected deps: logging, timer (optional, for timeouts).
No framework dependency required.


di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+
+addclientdetails:{[h].z.M.clients insert (cp[];h;.z.u;.z.a;.z.h)} / record client identity on connect for audit and orphan-query cleanup on disconnect
+
+removeclienthandle:{[h] / on client disconnect, mark their pending queries errored so result slots are not leaked
+  update error:1b,returntime:.z.m.cp[] from .z.M.queryqueue where clienth=h, null returntime;
+  .z.m.results:(exec queryid from .z.m.queryqueue where clienth=h)_results};
+
+addquery:{[query;servertype;join;postback;timeout;sync] / enqueue a query without dispatching; caller must call runnextquery[] to trigger dispatch
+  .z.M.queryqueue upsert (queryid;cp[];.z.w;query;servertype;join;{$[11h=type x;enlist x;x]}postback;timeout;0Np;0b;sync);
+  .z.m.queryid:queryid+1};
```
DI-Software-Engineering, line 52: `addquery` increments `queryid` with `.z.m.queryid:queryid+1`, but `queryid` on the right-hand side resolves to the module-local variable at definition time in `{...}` context — inside the lambda body, bare `queryid` without `.z.m.` prefix will resolve to the global/outer-context value rather than the updated module variable. This means the counter never advances past its initial value after the first call: every subsequent call overwrites the same `queryid=0` row. Use `.z.m.queryid:.z.m.queryid+1` to be consistent and correct.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+setavailableservers:{.z.m.availableservers:x} / swap in a custom routing strategy without forking core dispatch
+
+addclientdetails:{[h].z.M.clients insert (cp[];h;.z.u;.z.a;.z.h)} / record client identity on connect for audit and orphan-query cleanup on disconnect
+
+removeclienthandle:{[h] / on client disconnect, mark their pending queries errored so result slots are not leaked
+  update error:1b,returntime:.z.m.cp[] from .z.M.queryqueue where clienth=h, null returntime;
+  .z.m.results:(exec queryid from .z.m.queryqueue where clienth=h)_results};
+
+addquery:{[query;servertype;join;postback;timeout;sync] / enqueue a query without dispatching; caller must call runnextquery[] to trigger dispatch
+  .z.M.queryqueue upsert (queryid;cp[];.z.w;query;servertype;join;{$[11h=type x;enlist x;x]}postback;timeout;0Np;0b;sync);
```
DI-Software-Engineering, line 51: The `upsert` in `addquery` also references bare `queryid` (not `.z.m.queryid`) for the key value it inserts. If the module convention requires `.z.m.` for reads of module-local mutable state inside lambdas, this is broken in the same way: every call inserts with `queryid=0`, clobbering the previous queued query. Change to `.z.m.queryid`.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+
+getnextqueryid:{ / pick the oldest FIFO-eligible query whose required servertypes are all currently idle
+  avail:exec distinct servertype from availableservers 1b;
+  runnable:0!select from .z.m.queryqueue where null returntime, not queryid in key .z.m.results, {all x in y}[;avail] each servertype;
+  1 sublist select from runnable where time=min time};
+setgetnextqueryid:{.z.m.getnextqueryid:x} / inject a priority or custom scheduling strategy without forking the module
+
+addserverresult:{[qid;data] / fill one result slot; once all slots for a query are filled, run the join function and reply to the client
+  if[not qid in key results;:()];
+  st:first exec servertype from .z.m.servers where handle=.z.w;
```
DI-Software-Engineering, line 65: `addserverresult` uses `results[qid;1]` to fetch the slot dictionary, then mutates `slots` locally and writes back with `results[qid]:(results[qid;0];slots)`. But `results[qid]` was previously set in `runnextquery` as `(clienth;slots-dict)`, a 2-element list. After the write-back the structure is `(clienth; updated-slots)`, which is correct — **but** on the very same path, `results[qid;0]` is the clienth (an int) and `results[qid;1]` is the slots dict. After `finishquery` clears the entry and `runnextquery` is called (line 68), `addserverresult` checks `if[not qid in key results;:()]` at line 63 — but it does **not** re-check after `runnextquery[]` on line 68. `runnextquery` may call `addserverresult` recursively (if a queued query immediately triggers another result callback in-process). The second guard at line 70 (`if[not qid in key results;:()]`) does protect against that, so the logic is fine — but the `results[qid]:(results[qid;0];slots)` write on line 67 overwrites the whole entry using an index-assign on a global dict (`results`) without `.z.m.` prefix, whereas elsewhere mutable globals are updated via `.z.m.*`. This is inconsistent and may bypass the module's mutation mechanism, corrupting the shared `results` dict.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+removequeries:{[age] / prevent queryqueue growing unboundedly; purge completed queries older than age
+  .z.m.queryqueue:0!delete from .z.m.queryqueue where not null returntime, .z.m.cp[]>returntime+age};
+
+getnextqueryid:{ / pick the oldest FIFO-eligible query whose required servertypes are all currently idle
+  avail:exec distinct servertype from availableservers 1b;
+  runnable:0!select from .z.m.queryqueue where null returntime, not queryid in key .z.m.results, {all x in y}[;avail] each servertype;
+  1 sublist select from runnable where time=min time};
+setgetnextqueryid:{.z.m.getnextqueryid:x} / inject a priority or custom scheduling strategy without forking the module
+
+addserverresult:{[qid;data] / fill one result slot; once all slots for a query are filled, run the join function and reply to the client
```
DI-Software-Engineering, line 63: `addserverresult` reads `results` and writes `results[qid]` using bare unqualified name (no `.z.m.` prefix), inconsistent with every other mutable-state access in the module (which uses `.z.m.*`). If `.z.m` is required for module-namespace writes, these bare assignments silently write to a different (global) `results` variable, causing the accumulator to be permanently out of sync.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+  finishquery[qid;res 0]};
+
+addservererror:{[qid;err] / short-circuit a query on backend failure; free the server and notify the client before moving on
+  sendclientreply[qid;errorprefix,err;0b];
+  update inuse:0b from .z.M.servers where handle in .z.w;
+  runnextquery[];
+  finishquery[qid;1b]};
+
+sendclientreply:{[qid;result;status] / deliver result or error to the client, handling sync vs async send and postback wrapping in one place
+  qd:queryqueue[qid];
```
DI-Software-Engineering, line 86: `sendclientreply` passes `not status` as the second argument to `-30!` (the error flag), but then wraps `result` through `formatresponse[1b;1b;result]` only on the success branch. On the error branch (`not status`) the raw `result` (already prefixed by `errorprefix`) is passed directly to `-30!` without going through `formatresponse`. This inconsistently bypasses `formatresponse` for sync errors, breaking the pluggable hook contract described in the documentation.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+  tosend:$[()~qd`postback;result;qd[`postback],enlist[qd`query],enlist result];
+  $[qd`sync;
+    @[-30!;(qd`clienth;not status;$[status;formatresponse[1b;1b;result];result]);{}];
+    @[neg qd`clienth;formatresponse[status;0b;tosend];()]]};
+
+finishquery:{[qid;err] / remove query from the live results accumulator and stamp its completion time; keeps queryqueue and results consistent
+  .z.m.results:(qid,())_results;
+  update error:err,returntime:.z.m.cp[] from .z.M.queryqueue where queryid in qid};
+
+serverexecute:{[qid;query] / runs on the backend; traps errors locally so a crash posts an error reply rather than silently dropping the result
```
DI-Software-Engineering, line 97: `checktimeout` calls `finishquery[qids;1b]` where `qids` is a list of queryids. `finishquery` does `update error:err,returntime:... where queryid in qid` — the `in` handles a list, so that part is fine. But the first line of `finishquery` is `.z.m.results:(qid,())_results`. When `qid` is a list, `(qid,())` is already a list, so the drop `_` removes all those keys at once — that is correct. However `sendclientreply` is called via `each` but `finishquery` is called once with the whole list; if any of those queryids was not in `results` the drop is a no-op. This is fine, but the inconsistency between per-element `sendclientreply` and bulk `finishquery` means that if `sendclientreply` errors on one qid, `finishquery` still runs for all — silently swallowing the send failure.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+  res:@[{(0b;value x)};query;{(1b;"server ",(string .z.h),":",(string system"p"),": ",x)}];
+  @[neg .z.w;$[res 0;(errorcallback;qid;res 1);(resultcallback;qid;res 1)];
+    {@[neg .z.w;(errorcallback;x;"failed to return result: ",y);()]}[qid]]};
+
+sendquerytoserver:{[qid;query;handles] / fan the query out to all required handles and mark them in-use atomically to prevent double-dispatch
+  (neg handles,:())@\:(serverexecute;qid;query);
+  update inuse:1b from .z.M.servers where handle in handles};
```
DI-Software-Engineering, line 104: `removeserverhandle` computes `qids` (in-flight queries using the disconnected handle) by iterating `each key .z.m.results`. After `finishquery[qids;1b]` removes those entries from `results`, `qids2` is computed with `not queryid in key .z.m.results` — this correctly excludes the now-removed in-flight entries. However, `sendclientreply` for `qids` is called before `finishquery`, and `sendclientreply` checks `if[qd`error;:()]` using `queryqueue`. Since `finishquery` has not yet run, `error` is still `0b` and the send proceeds. But there is no guard against sending a second reply to the same client if the same queryid appears in both `qids` and `qids2` (e.g. a query that is both in-flight and somehow also queued). The two lists are not guaranteed disjoint by construction.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+  handles:avail types;
+  qid:torun`queryid;
+  slots:types!(count[types],())#enlist(0Ni;(::);0b);
+  slots[types;0]:handles;
+  results[qid]:(torun`clienth;slots);
+  sendquerytoserver[qid;torun`query;handles]};
+
+checktimeout:{ / periodic scan to error queries that have waited beyond their timeout, preventing them from hanging indefinitely
+  qids:exec queryid from .z.m.queryqueue where not timeout=0Wn, null returntime, .z.m.cp[]>time+timeout;
+  if[count qids;
```
DI-Software-Engineering, line 120: In `execquery`, the sync branch calls `.[{[q;s;j;t]addquery[q;s;j;();t;1b];runnextquery[]};(query;servertype;join;timeout);{-30!(.z.w;1b;x)}]`. The error handler `{-30!(.z.w;1b;x)}` attempts to use `.z.w` inside the error trap — but `.z.w` is the **current** connection handle at the time the trap fires. If the trap fires asynchronously or from a different call stack, `.z.w` may not be the original client handle. The client handle should be captured before the protected call (e.g. `ch:.z.w`) and closed over in the error lambda.

di/asyncdispatch/asyncdispatch.q
```diff
@@ -0,0 +1,159 @@
...
+removeclienthandle:{[h] / on client disconnect, mark their pending queries errored so result slots are not leaked
+  update error:1b,returntime:.z.m.cp[] from .z.M.queryqueue where clienth=h, null returntime;
+  .z.m.results:(exec queryid from .z.m.queryqueue where clienth=h)_results};
+
+addquery:{[query;servertype;join;postback;timeout;sync] / enqueue a query without dispatching; caller must call runnextquery[] to trigger dispatch
+  .z.M.queryqueue upsert (queryid;cp[];.z.w;query;servertype;join;{$[11h=type x;enlist x;x]}postback;timeout;0Np;0b;sync);
+  .z.m.queryid:queryid+1};
+
+removequeries:{[age] / prevent queryqueue growing unboundedly; purge completed queries older than age
+  .z.m.queryqueue:0!delete from .z.m.queryqueue where not null returntime, .z.m.cp[]>returntime+age};
```
DI-Software-Engineering, line 55: `removequeries` reassigns `queryqueue` with `.z.m.queryqueue:0!delete from .z.m.queryqueue where ...`. `queryqueue` is defined as a keyed table (`([queryid:...] ...)`). After `delete ... from` on a keyed table the result remains keyed, but `0!` unkeys it, so the reassignment replaces the keyed table with a plain table. Subsequent `upsert` and keyed-table operations will fail with a type error on the next `addquery` call.

PR #110, opened 2026-06-29
Feature serverselect

ascottDI:
# di.serverselect — TorQ Modularisation PR

## Summary

Extracts the server-selection logic from TorQ's `.gw` namespace (`gatewaylib.q` and `gateway.q`) into a standalone kdb-x module: `di.serverselect`. The module maintains a registered pool of backend servers and selects from them by servertype or attribute requirements. It satisfies the di.* module contract: dependency injection via `init`, a clean exported API, and no hard module dependencies beyond an injected logger.

---

## Background

TorQ's gateway uses a `.gw` namespace to track connected backend servers and route queries to them. Server registration, active-flag management, and selection logic were entangled with the gateway's query execution and connection-handling code. This PR extracts the server-selection layer — the pool of registered servers and the logic to pick from it — as an independently loadable and testable unit.

This PR is part of the broader TorQ → kdb-x modularisation effort. The extracted module removes the dependency on TorQ's global process framework and can be used in any gateway or proxy that needs to maintain a backend server pool.

---

### New files

| File | Description |
|---|---|
| `di/serverselect/serverselect.q` | Core implementation — `init`, `addserverfull`, `addserverattr`, `addserver`, `setserveractive`, `getserverstable`, `addserversfromtable`, `getservers`, `selector`, `getserverbytype`, `gethandlebytype`, `gethpbytype`, `getserverids` and internal helpers |
| `di/serverselect/init.q` | Module entry point — loads `serverselect.q` and declares the export list |
| `di/serverselect/test.csv` | k4unit unit-test suite (128 assertions) |
| `di/serverselect/integration.q` | End-to-end integration test (76 assertions) |
| `di/serverselect/serverselect.md` | Module README |

| Aspect | TorQ `.gw` | `di.serverselect` |
|---|---|---|
| Logging | Hard-coded `.lg.o`/`.lg.e` calls | Injected `log` dependency, required via `init` (no fallback) |
| Server state | Global `.gw.servers` table | `.z.m.servers` module-local mutable state |
| Error handling | Mix of silent failures and hard exits | Every error path logged then signalled with a `di.serverselect:` prefix |
| Attribute matching | Inlined in query dispatch | Extracted as `attributematch` (internal); exposed via the `getservers` `attribmatch` column |
| Bulk registration | `.servers.SERVERS` hard-coded lookup | Generic `addserversfromtable[proctypes;conntable]` accepting any connection table |

---

Logging is an injected dependency, wired via `init` — there is **no default logger** and the module does **not** load `kx.log` itself; initialising the logging framework is the job of the start-up script or the user. `init` **must** be called before any other function.

- **`init[deps]`** takes a single dict carrying the required `log` dependency (plus any future optional config). It errors immediately (plain signal) if `deps` is not a dict, is missing `` `log ``, or `log` is not a dict exposing `` `info`warn`error ``.
- **`normlog`** normalises the injected logger to a binary `` `info`warn`error!{[c;m]} `` dict — each function takes a context symbol `c` and a message string `m`. It accepts **either**:
  - a whole `kx.log` instance (`(use\`kx.log)[\`createLog][]`) — detected by its `getlvl`/`sinks`/`fmts` keys; its monadic functions are wrapped to `{[c;m]}`, folding context in as a `"ctx: msg"` prefix; or
  - a bespoke binary `` `info`warn`error!({[c;m]};{[c;m]};{[c;m]}) `` dict — passed through unchanged.
- Every call site uses the binary form `.z.m.log[\`info][\`ctx;"msg"]` (which maps 1:1 with TorQ's `.lg.o[\`ctx;"msg"]`).
- Errors are routed through an internal `raiseerror[ctx;msg]` helper that logs via `.z.m.log[\`error]` **then** signals `'"di.serverselect: ",ctx,": ",msg`, so every failure is observable in the log as well as thrown.

---

## Exported API

```q
srvsel:use`di.serverselect

/ wire the required log dependency first (pass the whole kx.log instance; normlog wraps it)
srvsel.init[enlist[`log]!enlist (use`kx.log)[`createLog][]]

srvsel.addserverfull[h;pname;st;hp;att]   / register a server with full details
srvsel.addserverattr[h;st;att]            / register a server with servertype and attributes
srvsel.addserver[h;st]                    / register a server with no attributes
srvsel.setserveractive[h;active]          / mark a server active (1b) or inactive (0b)
srvsel.getserverstable[]                  / return the full registered server table
srvsel.addserversfromtable[types;ctab]    / bulk-register from a connection table
srvsel.getservers[nameortype;lookups;req] / look up active servers with attribute scoring
srvsel.selector[servertable;selection]    / pick one row using roundrobin / any / last strategy
srvsel.getserverbytype[ptype;col;sel]     / return one column value for a servertype
srvsel.gethandlebytype[ptype;sel]         / convenience projection: return handle
srvsel.gethpbytype[ptype;sel]             / convenience projection: return hpup
srvsel.getserverids[att]                  / return server IDs by servertype list or attribute dict
```

---

## Hardening and fixes

- **`updatestats` keyed on `serverid`, not `handle`.** `handle` is not unique (the table is keyed on `serverid`, and a freed handle can be reused after reconnect). Keying stats updates on the unique `serverid` prevents a single selection from skewing `hits`/`lastp` across every server that happens to share a handle. Covered by a dedicated regression test.
- **Input validation on the public mutators**, routed through `raiseerror` so the failure is logged: `addserverfull`/`setserveractive` check handle (int) and servertype/active-flag types; `addserversfromtable` checks the connection table has the required `w`/`proctype`/`attributes` columns.
- **`getservers` validates `nameortype`** — it must be `` `servertype `` or `` `procname `` (when `lookups` is not `` ` ``); any other value errors rather than silently falling through to a `procname` lookup.
- **`selector` empty-table behaviour documented** — it returns a null-valued row; the `*bytype` helpers guard against this and return `()` when no active server matches.

---

## Test coverage

Two suites, both green.

### Unit tests — `test.csv` (k4unit), 128 assertions

| Area | Coverage |
|---|---|
| `init` — dependency injection | accepts a valid log dep; rejects non-dict `deps`, missing `log`, non-dict `log`, log dict missing a key; `di.serverselect:` error prefix |
| `init` — injected logger used | a capturing binary logger receives info messages on registration and error messages on failure paths |
| registration | `addserver`/`addserverattr`/`addserverfull` — row counts, handles, servertypes, procname/hpup population, serverid autoincrement, type-error rejection |
| `setserveractive` | active flag toggled; missing handle is a no-op; type-error rejection |
| `updatestats` | a selection updates only the chosen `serverid`, even when a handle is shared |
| `getservers` | column schema incl. `attribmatch`; servertype/procname filters; null lookups; per-attribute match scoring; invalid `nameortype` rejected |
| `selector` | `roundrobin`/`last`/`any` strategies; single-row and empty-table behaviour; unknown strategy rejected |
| `getserverbytype` / `gethandlebytype` / `gethpbytype` | column value returned; `()` for unknown type; round-robin rotation; `hits` increment |
| `getserverids` — symbol path | IDs for registered types; rejects null / unregistered / all-inactive |
| `getserverids` — attribute path | date/sym match; cross and independent matching; servertype-scoped dict; `besteffort` strict mode |
| `addserversfromtable` | bulk register; skips already-active handles; proctype filter; `` `ALL ``; optional `procname`/`hpup` columns; missing-column rejection |

### Integration test — `integration.q`, 76 assertions

Drives every exported function through a realistic gateway lifecycle (register → activate/deactivate → query → select → bulk-register → error handling → `init` re-wiring with a capturing logger), with a `PASS`/`FAIL` summary and a non-zero exit code equal to the number of failures.

### Running the tests

```bash
```

```q
/ unit tests
k4unit:use`di.k4unit;
k4unit.moduletest`di.serverselect;
```

```bash
/ integration test
QPATH=/path/to/kx/mod:/path/to/kdbx-modules q integration.q
```

---

## Integration notes

- `di.serverselect` has no hard module dependencies. The injected `log` is required — `init` throws if it is not provided, and must be called before any other function.
- The caller wires `kx.log` by passing the **whole** `createLog[]` instance (`srvsel.init[enlist[\`log]!enlist (use\`kx.log)[\`createLog][]]`); `normlog` detects and adapts it. A stripped 3-key dict of kx.log's monadic functions must **not** be passed (it would bypass detection and fail with `'rank`).
- `addserversfromtable` accepts a TorQ `.servers.SERVERS`-style table directly: columns `w` (handle), `proctype`, `attributes` are required; `procname` and `hpup` are optional.
- `getserverids` supports the full TorQ gateway attribute-matching contract: cross-product matching (default), independent matching, per-servertype scoping, and `besteffort` mode. Its result is consumed by the gateway via `inter/:` + `first each` (and `raze` for emptiness checks), which is robust to its per-path return shape.
- All error messages are prefixed `di.serverselect:` to identify the source in stack traces.

### Test results screenshot

image
image


di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+  free:cand where not inuse each cand;
+  if[n>count free;
+    '"could not find ",string[n]," free ports scanning up from ",string start];
+  :n#free;
+  };
+ports:findfree[baseport;count names];
+
+launch:{[port]
+  / start a bare q listener detached (no script needed - identity is injected over IPC)
+  system qbin," -p ",string[port]," -q </dev/null >/dev/null 2>&1 &";
```
DI-Software-Engineering: The `inuse` probe opens a handle with `hopen` but only closes it on success (the `hclose` path). When the port is free, `hopen` throws and `0b` is returned without any handle to close — that is fine. However, the success branch returns `1b` immediately after `hclose`, which is correct. The real problem is the timeout: `hopen (hp;500)` with a 500 ms timeout will block for up to 500 ms per port when scanning for free ports, making the scan of `10*count names` = 70 candidates potentially take 35 seconds on a machine where many ports are filtered (rather than refused). A shorter timeout (e.g. 50 ms) would be safer here, or the probe should use a non-blocking connect approach.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+  };
+
+connect:{[port]
+  / open a handle, retrying for up to ~10s while the child binds and accepts
+  hp:`$":localhost:",string port;
+  step:{[hp;h] if[not null h; :h]; system"sleep 0.2"; @[hopen;(hp;500);{0Ni}]}[hp];
+  :50 step/ 0Ni;
```
DI-Software-Engineering, line 92: `baseport:$[0<count s:getenv`SSPORT; "I"$s; 20000+`int$.z.i mod 20000]` — when `$SSPORT` is not set, `getenv` returns an empty string `""`, so `count s` is `0` and the else branch is taken (correct). But when `$SSPORT` IS set, `"I"$s` casts a char-vector to an int-vector (type 6h), not a single int (type 6i). `findfree` then calls `start+til 10*n` where `start` is an int-vector, producing a matrix rather than a vector of candidate ports, so `inuse each cand` will fail at runtime. The fix is `"i"$s` (lowercase, int atom) or `first "I"$s`.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+  };
+
+/ launch the fleet, connect, capture each child's own pid (.z.i) for exact-pid teardown
+-1 "selected free ports: ",.Q.s1 ports;
+launch each ports;
+handles:connect each ports;
+pids:{$[null x; 0Ni; @[x;".z.i";{0Ni}]]} each handles;
+/ guarantee the fleet is reaped on ANY exit (normal, error-abort, or exit code)
+.z.exit:{teardown[]};
+if[any null handles;
```
DI-Software-Engineering, line 112: `.z.exit` is assigned directly (`".z.exit:{teardown[]}"`) rather than via `.dotz.set`, violating the TorQ convention and potentially silently overwriting an existing handler (e.g. one set by the TorQ framework itself). Use `.dotz.set[`.z.exit; {teardown[]}]` instead.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+  -2 "failed to connect to child processes on ports: ",(" " sv string ports where null handles);
```
DI-Software-Engineering, line 113: If `any null handles` is true, `exit 3` is called. `.z.exit` fires and `teardown` is invoked, but `handles` at that point may contain `0Ni` nulls; the line `@[hclose;;{}] each handles where not null handles` in `teardown` should be safe. However, the processes that *did* start but whose handles are null (connect timed out) will only be reaped by the `pkill` fallback which matches `"bin/q [-]p <port>"`. If `qbin` is a custom path that does not contain `bin/q`, the `pkill` pattern won't match and those children will be orphaned. The `pkill` pattern should be derived from `qbin` or use a more robust signal.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+  / fallback so children are reaped even when we never connected (no pid).
+  / the [-]p bracket keeps the pattern from matching this very kill command.
+  live:pids where not null pids;
+  if[count live; @[system;"kill ",(" " sv string live)," 2>/dev/null; true";{}]];
+  {@[system;"pkill -f \"bin/q [-]p ",string[x],"\" 2>/dev/null; true";{}]} each ports;
+  @[hclose;;{}] each handles where not null handles;
+  };
+
+/ launch the fleet, connect, capture each child's own pid (.z.i) for exact-pid teardown
+-1 "selected free ports: ",.Q.s1 ports;
```
DI-Software-Engineering, line 107: `connect` uses `50 step/ 0Ni` (a fixed-point `/` convergence scan). Because `step` returns `0Ni` (a null int) on failure and an open handle integer on success, convergence will never naturally terminate if the child never comes up — `0Ni ~ 0Ni` is `1b` so q's fixed-point `/` would stop after the *first* failure, returning `0Ni` immediately rather than retrying 50 times. The correct idiom for a counted retry is `50 step\` (scan, discard) followed by a check, or a `do[50; ...]` loop, or `({not null x} step\)` with a separate retry limit.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+chk["handles stored are the real open handles";handles til 5;exec handle from srvsel.getserverstable[]];
+chk["servertypes as registered";`rdb`rdb`hdb`hdb`gateway;exec servertype from srvsel.getserverstable[]];
+chk["addserverfull populated procname";`gw1;first exec procname from srvsel.getserverstable[] where servertype=`gateway];
+chk["addserverfull populated hpup";gwhp;first exec hpup from srvsel.getserverstable[] where servertype=`gateway];
+chk["addserver -> empty attributes";(()!());first exec attributes from srvsel.getserverstable[] where handle=handles 0];
+chk["all active by default";1b;all exec active from srvsel.getserverstable[]];
+
+/ =========================================================================
+hdr"STEP 3  getservers (lookup + attribute match scoring)";
+chk["lookups=` returns all active";5;count srvsel.getservers[`;`;()!()]];
```
DI-Software-Engineering, line 156: STEP 9 kills `pids 0` (rdb1) and then sleeps 0.3 s before deactivating it. However `pids 0` may be `0Ni` (if the `@[x;".z.i";{0Ni}]` capture failed), in which case `system"kill 0Ni ..."` passes a literal string `"0Ni"` to the shell — which will fail silently but leave rdb1 alive. The test then deactivates the handle and assumes rdb1 is dead, but subsequent `route` calls may still reach rdb1 via its open handle. Add a null check before killing, or assert `not null pids 0` at the start of STEP 9.

di/serverselect/integration.q
```diff
@@ -0,0 +1,262 @@
...
+chk["procname=gw1 returns the gateway handle";enlist handles 4;exec handle from srvsel.getservers[`procname;`gw1;()!()]];
+chk["unknown servertype returns empty";0;count srvsel.getservers[`servertype;`nope;()!()]];
+chk["attribmatch column present";1b;`attribmatch in cols srvsel.getservers[`;`;()!()]];
+chk["complete date match flagged for some hdb on d1";1b;
+  any {x[`date]0} each (srvsel.getservers[`servertype;`hdb;(enlist`date)!enlist enlist d1])`attribmatch];
+
+/ =========================================================================
+hdr"STEP 4  selector (strategies on a known table)";
+seltab:([]serverid:101 102 103i;handle:201 202 203i;lastp:(2024.01.03D0;2024.01.01D0;2024.01.02D0));
+chk["roundrobin picks oldest lastp";202i;(srvsel.selector[seltab;`roundrobin])`handle];
```
DI-Software-Engineering, line 168: STEP 9 routes 6 queries after killing rdb1 and storing results as `postresp`. Then it asserts `not `rdb1 in postresp`. However `route` returns `` `NONE `` when `gethandlebytype` returns `()`, and the open handle for rdb1 is still registered (only deactivated). If the dead rdb1 handle is somehow selected before deactivation takes effect (race), or if `hopen` to a dead process returns an error swallowed by the route trap, `route` returns `` `NONE ``. The assertion `not `NONE in postresp` catches that, but the assertion `not `rdb1 in postresp` would still pass in that case — leaving the route failure undetected in one scenario. This is acceptable defensive test logic, but the ordering of `kill` → `sleep` → `setserveractive` means there is a window where rdb1 is still active and its handle is dead, so the very first `route` call in `postresp` could hit the dead handle and return `` `NONE `` rather than `rdb2`/`rdb3`, causing `not `NONE in postresp` to fail spuriously.

di/serverselect/serverselect.q
```diff
@@ -0,0 +1,275 @@
...
+
+getserverids:{[att]
+  / return server IDs matching a servertype list or attribute requirement dictionary
+  / att: symbol list of servertypes, or dict of attribute requirements (optionally keyed on `servertype)
+  / dispatch the symbol-list path to getserveridsbytype and the dict path to getserveridstype
+  if[99h<>type att; :getserveridsbytype att];
+  / validate attribute-requirement values are atoms or simple vectors (control keys excepted);
+  / a nested/general-list value (type 0h) would otherwise crash the cross matcher with a raw 'type
+  reqkeys:key[att] except `servertype`attributetype`besteffort;
+  if[count badkeys:reqkeys where 0h=type each att reqkeys;
```
DI-Software-Engineering: The new validation `reqkeys where 0h=type each att reqkeys` indexes `att` by `reqkeys` (a symbol vector) using indexing-at-depth, which in q returns corresponding values from the dict. `type each` on the resulting values is correct. However, `att reqkeys` when `reqkeys` is a single-element symbol vector returns a list of one value — `type each` returns a list of one type — so the comparison `0h=type each ...` and `where` work correctly. The edge case to check is when `reqkeys` is empty (no non-control keys): `att reqkeys` returns an empty list, `type each` returns `()`, `0h=()` is `()`, `count badkeys` is 0, and the branch is skipped. That is correct. No bug here — this note is for confirmation only (no finding to post).

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **7 warning(s)** | **0 suggestion(s)**

di/serverselect/integration.q
ascottDI: Good point, changed to 50ms

di/serverselect/integration.q
ascottDI: Incorrect, casting using the lowercase "i" will results in ascii for each char instead of int

di/serverselect/integration.q
ascottDI: Ran as a standalone test so not necessary, additionally the use of .dotz.set would result in a failure as it is not available here.

di/serverselect/integration.q
ascottDI: Actual error, 
Fixed. processes that started but never connected had no captured pid (came from .z.i), so were only reachable via pkill which a custom $QBIN breaks.

launch now captures the shell pid (& echo $!) -> exact pid for every child, no .z.i reliance.
pkill fallback rekeyed to [-]p <port> -q (unique port + flag), qbin-independent.

di/serverselect/integration.q
ascottDI: Incorrect, reviewer has the "/" converge semantics inverted. Correctly calls 50 times

di/serverselect/integration.q
ascottDI: Added chk["rdb1 pid captured";1b;not null pids 0] at the start of STEP 9. Note the .z.i capture you cite was already replaced with a shell-$! capture in the prior fix from comment on line 98-101, so pids 0 is now essentially always set. Also, routing can't actually reach rdb1 — route selects only active servers and we deactivate it. The real risk the guard closes: a null pid → kill "0Ni" no-ops silently, letting STEP 9 pass while only exercising deactivation rather than a dead backend. Now it fails loudly instead.

di/serverselect/integration.q
ascottDI: Not an error
the reviewer is misreading the ordering. setserveractive[handles 0;0b] runs before the postresp route loop, and as q is single-threaded, there's no race condition and no window where a route call sees rdb1 active-but-dead (that window is entirely between kill and setserveractive, with no routing in it).

di/serverselect/serverselect.q
ascottDI: Yes this is correct, works just as stated

di/serverselect/integration.q
```diff
@@ -0,0 +1,271 @@
...
+  cand:start+til 10*n;
+  free:cand where not inuse each cand;
+  if[n>count free;
+    '"could not find ",string[n]," free ports scanning up from ",string start];
+  :n#free;
+  };
+ports:findfree[baseport;count names];
+
+launch:{[port]
+  / start a bare q listener detached and capture its shell background pid ($!) - so every
```
DI-Software-Engineering, line 75: The return value of `launch` is now the shell pid string parsed as an int, but if the `system` call returns an empty list (e.g. on a platform where `echo $!` is unavailable or the shell strips the output), `first` of an empty list returns a null, and `"I"$` of a null/empty result would silently produce `0Ni` — which the rest of the code already guards against, but the failure mode is silent. The real risk is that `first system …` on some q versions returns a char list (the raw string), not a string atom, so `"I"$` parses correctly; however if the command fails entirely, `system` raises a signal and the `@[…;…;{}]`-style trap is absent here. Wrap the call in a trap: `@[{"I"$first system x}; cmd; {0Ni}]` to prevent an uncaught signal from aborting the entire launch phase.

di/serverselect/integration.q
```diff
@@ -0,0 +1,271 @@
...
+  step:{[hp;h] if[not null h; :h]; system"sleep 0.2"; @[hopen;(hp;500);{0Ni}]}[hp];
+  :50 step/ 0Ni;
+  };
+
+teardown:{[]
+  / always-run cleanup. kill every launched child by its captured pid (exact). the pkill
+  / fallback is a last resort for the rare case a pid was never captured - keyed on our own
+  / unique port and -q flag (NOT the qbin path, which is configurable), so a custom qbin
+  / cannot break it. the [-]p bracket keeps the pattern from matching this very kill command.
+  live:pids where not null pids;
```
DI-Software-Engineering, line 100: The pkill fallback pattern `"[-]p ",string[x]," -q"` matches any process with that port number and `-q` flag, not just processes spawned by this test run. On a shared host another q process started on the same port (possible if `findfree` races) would be killed. The previous pattern anchored on `bin/q` to narrow the match; removing that anchor widens it unnecessarily. Reintroduce a narrower anchor (e.g. include the qbin basename or the process group) so the fallback cannot kill unrelated processes.

di/serverselect/integration.q
```diff
@@ -0,0 +1,271 @@
...
+  if[count live; @[system;"kill ",(" " sv string live)," 2>/dev/null; true";{}]];
+  {@[system;"pkill -f \"[-]p ",string[x]," -q\" 2>/dev/null; true";{}]} each ports;
+  @[hclose;;{}] each handles where not null handles;
+  };
+
+/ launch the fleet (capturing each child's shell pid for exact-pid teardown), then connect
+-1 "selected free ports: ",.Q.s1 ports;
+pids:launch each ports;
```
DI-Software-Engineering, line 108: `pids:launch each ports` replaces the previous two-step sequence where `pids` was populated *after* connecting and querying `.z.i` from each child. The new code assigns pids immediately from the shell's `$!`, which is the pid of the background shell job, not necessarily the pid of the `q` process itself (on some systems the shell forks an intermediate process). If the shell pid and the q process pid differ, `kill <pid>` in teardown and in STEP 9 will target the wrong process. Verify that the shell used by `system` on all target platforms makes `$!` equal to the q process pid, or revert to querying `.z.i` over IPC after connecting.

DI-Software-Engineering:
## DIReview Summary

di/serverselect/integration.q
ascottDI: Doesn't apply to launch's command form. system only signals on a plain failing command; launch uses qbin … & echo $!$, which exits 0 even when the binary is missing (verified: bad qbin still returns the backgrounded pid, no signal), and echo $! always emits a pid so the empty -> null branch can't fire. An uncaptured pid is already handled. Teardown filters nulls and falls back to pkill on the port pattern, and connect works by port not pid. Adding @[…;{0Ni}] would only mask the one real failure (a broken $QBIN), turning an immediate 'os into a 10s timeout and murky assertion failures — worse in a test harness.

di/serverselect/integration.q
ascottDI: The qbin anchor was removed on purpose: $QBIN is configurable and can include an env prefix, this project's own launcher is QHOME=… …/q. Which never appears in the child's argv (verified: ps shows the child as /opt/…/q -p <port> -q, no QHOME=). Anchoring on qbin would then fail to match and orphan every child. Port + -q avoids that, and the over-kill risk is low anyway: ports are pid-derived and verified free at launch, with exact-pid kill as the primary reaper, so the pkill sweep normally hits an already-dead target. Leaving as-is.

di/serverselect/integration.q
ascottDI: Verified $! equals the child's own pid here: launched a q child with the exact pattern, captured $!, compared to its .z.i over IPC, identical (and ps shows the pid is the q process, …/q -p <port> -q). The intermediate-process case applies to compound backgrounded commands; launch backgrounds a simple command, which POSIX shells exec directly, so $! is the q pid (redirections add no process). Linux-only work. The switch was also deliberate, $! captures a pid for every child at launch, including one that never connects, whereas .z.i-over-IPC orphans any child that hangs during startup.

di/serverselect/serverselect.q
```diff
@@ -0,0 +1,270 @@
...
+
+updatestats:{[sid]
+  / internal - update last-access timestamp and hit count for a single server
+  / keyed on serverid (unique) rather than handle, which may be shared by multiple servers
+  .z.m.servers:update lastp:.z.p,hits:hits+1i from servers where serverid=sid;
+  };
+
+addserverfull:{[h;pname;st;hp;att]
+  / register a server with full details: handle, procname, servertype, hpup and attribute dictionary
+  if[not -6h=type h;.z.m.logerr[`addserverfull;err:"di.serverselect: handle must be an int; got type ",string type h];'err];
```
DI-Software-Engineering: The error variable `err` is assigned inside the `if` condition expression but `'err` is a separate statement — in q, `if` takes a condition and zero or more body expressions separated by semicolons, all inside the brackets. The pattern used here is:
```
if[not …; .z.m.logerr[…;err:"…"]; 'err]
```
This is valid q (two body expressions inside the `if`). However the same pattern was also used in a single-line form without brackets enclosing both body expressions in some earlier style (`raiseerror` was a single function call). Verify each call site carefully: every occurrence must have the signal `'err` inside the same `if[…;…;…]` bracket group, not dangling outside it as a bare statement that always executes. Audit every transformed call site — e.g. line 65 (`setserveractive`) — to confirm `'err` is inside the `if[…]` brackets and not a standalone expression that unconditionally throws.

DI-Software-Engineering:
## DIReview Summary

di/serverselect/serverselect.q
ascottDI: Audited all 16 sites — every 'err is inside its if[…] group, none dangling. Single-line guards end in 'err];; multi-line ones have 'err; immediately followed by the closing ];. The line-187 case (logerr message spans lines via a nested $[…]) is fine too: the ]; after activeserversmsg closes the logerr call, and 'err sits inside the if. Behavioural proof: an unconditional throw in a happy-path guard like setserveractive would fail its tests

di/serverselect/serverselect.q
```diff
@@ -0,0 +1,514 @@
...
+  .z.m.logwarn:deps[`log]`warn;
+  .z.m.logerr:deps[`log]`error;
+  .z.m.cp:getopt[deps;`cp;{.z.p}];
+  .z.m.clearinactivetime:getopt[deps;`clearinactivetime;0D01:00];
+  .z.m.maxcrossproduct:getopt[deps;`maxcrossproduct;1000000];
+  / seed the live selection strategy from the built-in default; setselector replaces it at runtime
+  .z.m.selector:selector;
+  };
+
+nextserverid:{
```
DI-Software-Engineering, line 122: `checkopt[deps;`maxcrossproduct;{(-7h=type x) and 0<x};"a positive long"]` rejects `0` as maxcrossproduct since `0<0` is false. The docs say `0W` disables the bound; `0` is not discussed. The error message says "a positive long" and `0` would be rejected. But the actual semantics in `getserverids` is `if[.z.m.maxcrossproduct<sz:...]` — if maxcrossproduct were `0`, every non-empty request would fail (since `sz>=1`). So `0` being rejected is arguably correct. No defect. Disregarding.

di/serverselect/init.q
```diff
@@ -0,0 +1,27 @@
...
+/ is done here; init fans it out into .z.m.loginfo/.z.m.logwarn/.z.m.logerr, called as
+/ .z.m.loginfo[`ctx;"msg"]
+
+/ module version, read from the VERSION file rather than hardcoded, so a release bump touches one
+/ plain-text file. read module-relative (`:::` resolves to di/serverselect) and BEFORE the export
+/ line, since export:([...]) evaluates each name.
+/ NB `version` must STAY in the export: di.depcheck resolves a dependency's version from the export
+/ dict (checkdepversion) and classes a missing one as a FAILURE - which makes di.depcheck.init throw
+/ for any process loading a module that declares this one as a hard dependency
+/ trim, and fail LOUD on a missing/unreadable/empty VERSION, rather than a bare `first read0`: a raw
```
DI-Software-Engineering, line 16: `VERSION` file loses its trailing newline (as shown by the `\ No newline at end of file` in the diff). The `read0` call in `init.q` reads the file and takes `first`, then `trim`s. If the file has no trailing newline, `read0` returns a one-element list with the version string (no trailing `\r` or space issues). `trim first read0` handles this correctly. However, other tooling (e.g. `cat`, `diff`, some CI version-check scripts) that expects a newline-terminated file may be broken. This is a minor portability concern but not a code defect in the q module itself.

di/serverselect/serverselect.q
```diff
@@ -0,0 +1,514 @@
...
+  requireinit`removeinactive;
+  if[not -16h=type age;raiseerror[`removeinactive;"age must be a timespan; got type ",string type age]];
+  if[null age;raiseerror[`removeinactive;"age must not be null"]];
+  if[age<0D;raiseerror[`removeinactive;"age must not be negative; got ",string age]];
+  / 0Wn means "retain forever" and MUST short-circuit: it cannot be expressed by the comparison below,
+  / because disconnecttime+0Wn overflows the timestamp range (wrapping back to the year 1734), so every
+  / inactive row would compare as aged out - the exact opposite of an infinite retention. 0Nn is
+  / rejected above for the same reason: disconnecttime+0Nn is 0Np, and cp[]>0Np is true for everything
+  if[0Wn=age;:()];
+  .z.m.servers:delete from servers where not active,not null disconnecttime,.z.m.cp[]>disconnecttime+age;
```
DI-Software-Engineering, line 204: `removeinactive` does not call `requireinit` before checking `not -16h=type age`. If called before `init`, the age type check and subsequent checks may run, and `if[0Wn=age;:()]` would return without touching `.z.m` — but the `delete` at the end references `.z.m.servers` and `.z.m.cp`, which are unset, producing a cryptic `'.m.di.0serverselect.servers` error instead of the expected `'di.serverselect: removeinactive: init must be called first`. Looking at the actual code: `requireinit\`removeinactive` is at line 200 (first line of the function body). Checking: `removeinactive:{[age] requireinit\`removeinactive; ...}`. The test confirms this at the requireinit section. So `requireinit` IS called first. No defect.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **2 warning(s)** | **5 suggestion(s)**

### Suggestions
- **di/serverselect/serverselect.q:119** — The `checkopt` validator for `maxcrossproduct` uses `{(-7h=type x) and 0<x}`, which rejects `0W` (positive infinity long). But `0W` is documented and tested as the "disable the bound" sentinel, and the code in `getserverids` explicitly compares `.z.m.maxcrossproduct<sz` meaning `0W` must pass `init`. The type check `(-7h=type x)` is correct for `0W`, but `0<0W` is `1b`, so `0W` actually does pass the check. However the error message says "a positive long" and `0W` is technically valid — the real defect is that the check also rejects `0W` being passed as `maxcrossproduct` only if the condition `0<x` were the issue. Re-examining: `0W` is a long atom with type `-7h` and `0<0W` is `1b`, so `0W` passes. But separately: the test at line 536 in test.csv asserts `fail` for `srvsel.init[`log`maxcrossproduct!(noop;0W)]`... wait, that test asserts `fail` on `0W`? No, looking again at test.csv line 536: `run,0,0,q,"srvsel.init[`log`maxcrossproduct!(noop;0W)]",1,1,re-init with 0W to disable the bound` — that is a `run` (expected to succeed). So `0W` must pass `checkopt`. Since `-7h=type 0W` and `0<0W` are both true, `0W` does pass. No defect here after careful re-examination. Disregarding this finding.
- **di/serverselect/serverselect.q:59** — The `requiredict` check for an empty dictionary `()!()` relies on `not $[0=count k:key d; not 98h=type k; 11h=type k]`. When `d` is `()!()`, `key d` is an empty general list (type `0h`), so `count k` is `0` and the branch taken is `not 98h=type k`, which evaluates to `not 98h=0h` → `not 0b` → `1b`, meaning the condition `not (1b)` = `0b`, so the `raiseerror` is NOT triggered and `()!()` is accepted. That is the intended behaviour. However when `d` is an empty keyed table `([a:`int$()]b:`int$())`, `key d` is an empty table (type `98h`), so `count k` is `0` (empty table has 0 rows) and the branch is again `not 98h=type k` → `not 98h=98h` → `not 1b` → `0b`, so `raiseerror` is NOT triggered and an empty keyed table is accepted. But the test at the G3 section asserts `fail` for `srvsel.addserverattr[9i;`hdb;([a:`int$()]b:`int$())]`. This means `requiredict` must reject an empty keyed table, but the logic as written would NOT reject it because an empty keyed table also has `count key = 0`. The guard `0=count k:key d` is true for both `()!()` and an empty keyed table, so both take the `not 98h=type k` branch — but `98h=type k` differs: for `()!()` it is `0h` giving `not 0b`=`1b` (accepted), for an empty keyed table `key` returns an empty table of type `98h` giving `not 1b`=`0b` (rejected). Actually for an empty keyed table `count key kt` where `kt:([a:`int$()]b:`int$())` — `key kt` returns the key table which has 0 rows but type `98h`. So `count k` is `0` AND `98h=type k` is `1b`, so `not 98h=type k` is `0b`, so `raiseerror` IS triggered. The logic is correct. Disregarding.
- **di/serverselect/serverselect.q:79** — `selectorarity` handles a projection (type `104h`) by calling `selectorarity first value f` recursively to get the underlying rank, then subtracts `count where not (::)~/:1_value f`. For a trailing partial application like `{[h;t;s]h}[9i]`, `value f` is `({[h;t;s]h}; 9i)`, so `1_value f` is `enlist 9i`, and `(::)~/:enlist 9i` is `enlist 0b`, so `count where not` is `1`. The underlying rank is `3` (from `selectorarity {[h;t;s]h}`), so the result is `3-1=2`. This is correct. But for `{[h;t;s]h}[9i;`t]`, `1_value f` is `(9i;`t)`, count where not is `2`, result is `3-2=1`. That also looks correct per the test. No defect. Disregarding.
- **di/serverselect/serverselect.q:366** — In `getserverids`, the cross-product size check uses `prd count each value req`. `req` is the result of `normreq`, which promotes atoms to one-element lists via `{(),x} each value req`. After `normreq`, values of `req` are lists. `count each value req` gives the count of each requirement list. `prd` of that gives the cross-product size. However, `normreq` is called with `ctrlkeys _ att` (flat form) or `att\`attrs` (nested form), and this happens before the `badkeys` check (nested list check). If `req` has a value that is a general list (type `0h`, e.g. `enlist enlist 2024.01.01`), `normreq` wraps it in `()` making it `(enlist 2024.01.01)` — a one-element general list. Then `count` of that is `1`, contributing `1` to the product. The `badkeys` check at line ~357 catches this: `0h=type each value req` would be true for the wrapped general list (`type enlist enlist 2024.01.01` is `0h`). Wait — after `normreq`, the value `enlist enlist 2024.01.01` gets wrapped as `(),(enlist enlist 2024.01.01)` = `enlist enlist 2024.01.01` (since `(),` on a list is identity). Actually `{(),x}[enlist enlist 2024.01.01]` = `(),(enlist enlist 2024.01.01)` which is a general list containing one element. Its type is `0h`. So the badkeys check still fires. No defect. Disregarding.
- **di/serverselect/serverselect.q:349** — `getserverids` validates the `servertype` control key with `not 11h=abs type ctl\`servertype`, which accepts both an atom symbol (`-11h`) and a symbol list (`11h`). However, when `servertype` is an atom symbol (e.g. `` `hdb ``), `distinct (),ctl\`servertype` correctly makes it a one-element list. The fan-out `fanoutids[req;besteffort;attype] each distinct (),ctl\`servertype` then works correctly. No defect.

di/serverselect/integration.q
```diff
@@ -0,0 +1,353 @@
...
+
+teardown:{[]
+  / always-run cleanup. kill every launched child by its captured pid (exact). the pkill
+  / fallback is a last resort for the rare case a pid was never captured - keyed on our own
+  / unique port and -q flag (NOT the qbin path, which is configurable), so a custom qbin
+  / cannot break it. the [-]p bracket keeps the pattern from matching this very kill command.
+  live:pids where not null pids;
+  if[count live; @[system;"kill ",(" " sv string live)," 2>/dev/null; true";{}]];
+  {@[system;"pkill -f \"[-]p ",string[x]," -q\" 2>/dev/null; true";{}]} each ports;
+  @[hclose;;{}] each handles where not null handles;
```
DI-Software-Engineering, line 103: The `connect` function's retry loop `50 step/ 0Ni` uses converge (`/` without a count), not a counted loop. Because `step` returns the handle integer on success and `0Ni` on failure, the converge will stop as soon as two successive calls return the same value — meaning it stops immediately on the first successful connect (correct), but it also stops if two consecutive retries both return `0Ni` (i.e. after just *two* failed attempts, not fifty). The intent is to retry up to 50 times; use `50 step/ 0Ni` as a *counted* iterate by making the count an integer: this already is an integer `50`, so in q `n f/ seed` with integer `n` applies `f` exactly `n` times. Verify this is the intended semantics — if so it is correct as written and this note can be disregarded.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **1 warning(s)** | **3 suggestion(s)**

### Suggestions
- **di/serverselect/serverselect.q:183** — `nextserverid` reads `serverid` (the module-level variable) directly by bare name rather than through `.z.m.serverid`, then writes the result back as `.z.m.serverid`. On the first call, `.z.m.serverid` does not exist yet (it is only written by this function, never initialised in `init` or at module level under `.z.m`), so `serverid` resolves to the top-level `0i` initialisation, which is correct on the first call. However, subsequent calls read `.z.m.serverid` via the bare name `serverid` only if the module namespace aliases it; if the bare name keeps resolving to the top-level `0i` constant every call, every server would get `serverid=1i`. Confirm that the bare `serverid` read on this line resolves to `.z.m.serverid` after the first write, or change it to `.z.m.serverid+1i` explicitly.
- **di/serverselect/serverselect.q:126** — `checkopt` for `maxcrossproduct` requires `(-7h=type x) and 0<x`, which rejects `0W` (positive infinity for longs, type `-7h`, and `0W>0` is `1b`). But the doc and test both state `0W` disables the bound and is accepted. `0W` satisfies `-7h=type 0W` and `0W>0` is `1b`, so the check actually passes for `0W` — that is correct. However the check `0<x` also rejects `0` as a bound (zero would mean reject every non-empty cross product, which may be intentional). Verify that `0` is intentionally rejected; if a zero bound should be allowed, change to `0<=x`.
- **di/serverselect/serverselect.q:270** — In `getserverids`, the `@[getserveridstype[req;besteffort;attype]; `all; raisecaught[`getserverids;]]` call passes the atom `` `all `` as the second argument to `.[]` — which means it is used as the *argument* to `getserveridstype[req;besteffort;attype]`. `getserveridstype` is a 4-argument function that has already been projected to 3 arguments, so this single remaining argument is `typ`. Inside `getserveridstype`, `typ=`all`` triggers the branch that returns all active servers regardless of servertype, which is the intended behaviour. The trap `raisecaught[`getserverids;]` is a projection awaiting one string argument — the error string — which is correct for `@[f;x;h]`. This appears correct. No defect.

PR #111, opened 2026-06-30
Feature HTML

Ruairi-wq2:
The big issue or concern I have with this module is that, the exported functions allow for html serving over the browser, but in order to make use, the user would need to have their own html doc which calls the exported functions in the module. The components for doing so are shipped with the module, but implementation could be awkward. I include a integration script with a test.html which demonstrates the use case. I have added a brief section on building a client HTML page but am not certain if this is the correct way to go about this.

### DIFF
--- a/di/html/html.q
+++ b/di/html/html.q
@@ -1,0 +1,201 @@
+/ di.html - websocket pub/sub and html page serving
+
+/ list of table names registered for pub/sub
+subtables:`symbol$();
+
+/ subscriber list per table: each entry is (handle; sym-filter)
+subs:()!();
+
+/ modifier function per table: transforms data before sending to subscriber
+modifier:()!();
+
+/ cache of resolved ip addresses: int -> symbol
+ipacache:(`int$())!`symbol$();
+
+/ home directory for html files - set by init
+homedir:"";
+
+/ logging functions - required injected deps, wired by init via .z.m
+
+/ converts a list of timestamps or dates to iso 8601 strings e.g. "2024-01-02T12:00:00Z"
+jstsiso8601:{[x] {("-" sv "." vs string `date$x),"T",string[`second$x],"Z"} each x};
+
+/ converts a list of dates to javascript epoch milliseconds
+/ kdb dates are days since 2000-01-01; 10957 is the day offset from 1970-01-01 to 2000-01-01
+jstsfromd:{[x] "j"$86400000 * 10957 + `long$x};
+
+/ converts a list of time, second, or minute values to milliseconds since midnight
+jstsfromt:{[x] "j"$"t"$x};
+
+/ converts a list of months to javascript epoch milliseconds via the first day of each month
+jstsfromm:{[x] jstsfromd `date$x};
+
+/ maps kdb type shorts to their javascript converter function
+/ types not listed here are left unchanged by jsformat
+typemap:12 13 14 16 17 18 19h!(jstsiso8601;jstsfromm;jstsfromd;jstsfromt;jstsfromt;jstsfromt;jstsfromt);
+
+jsformat:{[tbl]
+  / applies the correct javascript converter to each column of a table that needs it
+  / columns whose type is not in the typemap are passed through unchanged (via ::)
+  coldict:flip 0!tbl;
+  k:key coldict;
+  colvals:value coldict;
+  converters:typemap type each colvals;
+  :flip k!converters@'colvals;
+  };
+
+updformat:{[msgtype;msgdata]
+  / wraps an upd message into a name/data dictionary with javascript-formatted table data
+  formatteddata:(key msgdata)!(msgdata`tablename;jsformat msgdata`tabledata);
+  :(`name`data)!(msgtype;formatteddata);
+  };
+
+/ filter applied before sending data to a subscriber - returns full table (no filtering)
+sel:{[tbl;syms] tbl};
+
+del:{[tbl;handle]
+  / removes a handle from the subscriber list for a table
+  / if handle is not found, ? returns count of list and drop has no effect
+  idx:subs[tbl;;0]?handle;
+  .z.m.subs:@[subs;tbl;_;idx];
+  };
+
+add:{[tbl;syms]
+  / adds the current handle to the subscriber list for a table
+  / if already subscribed, updates the sym filter by taking union with new syms
+  / if not subscribed, appends a new (handle; syms) pair to the list
+  / returns (tablename; current data) so the subscriber can initialise their local copy
+  i:subs[tbl;;0]?.z.w;
+  .z.m.subs:$[(count subs tbl)>i;
+    .[subs;(tbl;i;1);union;syms];
+    @[subs;tbl;,;enlist(.z.w;syms)]
+    ];
+  :(tbl;$[99=type v:value tbl;sel[v;syms];@[0#v;`sym;`g#]]);
+  };
+
+closehandle:{[handle]
+  / removes the given handle from all subscriber lists when a connection closes
+  del[;handle] each subtables;
+  };
+
+replace:{[str;findreplace]
+  / applies each find->replace pair to the string in sequence using ssr
+  :(ssr/)[str;string key findreplace;value findreplace];
+  };
+
+ipa:{[ipint]
+  / resolves an ip address integer to a hostname symbol, caching results
+  / tries .Q.host first; falls back to converting the ip bytes manually if that fails
+  if[not `~r:ipacache ipint;:r];
+  hostname:.Q.host ipint;
+  r:$[`~hostname;`$"."sv string "i"$0x0 vs ipint;hostname];
+  .z.m.ipacache:@[ipacache;ipint;:;r];
+  :r;
+  };
+
+/ returns the current listen port as a string
+getport:{[] string system "p"};
+
+execdict:{[inputdict]
+  / extracts the func key and any additional args from a dictionary and calls the function
+  / args are passed to the function in the order the keys appear after func
+  if[not `func in key inputdict;'"no func in dictionary"];
+  f:value inputdict`func;
+  args:value inputdict _ `func;
+  :$[1=count key inputdict;f @ 1;f . args];
+  };
+
+/ websocket message handler - module-level so it carries the module context for evaluate
+wshandler:{neg[.z.w] -8!.j.j[evaluate[.j.k -9!x]];};
+
+addtables:{[tablelist]
+  / registers a list of tables for pub/sub and sets their default modifier
+  / can be called multiple times; already-registered tables are ignored
+  tablelist,:();
+  new:tablelist except subtables;
+  .z.m.subtables:subtables,new;
+  .z.m.subs:subs,new!(count new)#();
+  .z.m.modifier:modifier,new!(count new)#{-8!.j.j updformat["upd";`tablename`tabledata!(x 1;x 2)]};
+  if[count new;.z.m.lginfo[`html;"registered tables: ",", " sv string new]];
+  };
+
+pub:{[tbl;data]
+  / publishes data to all current subscribers of a table
+  / applies the table modifier before sending (default modifier json-encodes the data)
+  {[tbl;data;s]
+    if[count data:sel[data;s 1];
+      (neg first s) modifier[tbl]@(`upd;tbl;data)];
+    }[tbl;data] each subs tbl;
+  };
+
+sub:{[tbl;syms]
+  / subscribes the current handle to a table with an optional sym filter
+  / pass backtick as tbl to subscribe to all registered tables
+  / removes any existing subscription for this handle before re-adding
+  if[tbl~`;:sub[;syms] each subtables];
+  if[not tbl in subtables;'tbl];
+  del[tbl;.z.w];
+  :add[tbl;syms];
+  };
+
+wssub:{[tbl]
+  / subscribes via websocket, no return value
+  sub[tbl;`];
+  };
+
+end:{[eodval]
+  / broadcasts end-of-day message to all subscriber handles across all tables
+  (neg union/[subs[;;0]])@\:(`.u.end;eodval);
+  };
+
+readpage:{[filename]
+  / reads an html file from the configured home directory and returns it as a string
+  / returns a "not found" message string if the file does not exist
+  p:homedir,"/",filename;
+  r:@[read1;`$":",p;""];
+  if[not count r;.z.m.lgwarn[`html;p,": not found"]];
+  :$[count r;"c"$r;p,": not found"];
+  };
+
+readpagereplaceHP:{[filename]
+  / reads a page and replaces MYKDBSERVER and MYKDBPORT tokens with live server values
+  :replace[readpage filename;`MYKDBSERVER`MYKDBPORT!("\"",(string ipa .z.a),"\"";getport[])];
+  };
+
+evaluate:{[inputdict]
+  / safely calls execdict on the input, logging then re-throwing any errors with context
+  :@[execdict;inputdict;{[d;e] m:"failed to execute ",(-3!d)," : ",e;.z.m.lgerr[`html;m];'m}[inputdict]];
+  };
+
+init:{[config;deps]
+  / sets up module state from config and registers websocket handlers
+  / log dep is required; handlers dep is optional
+
+  / log dependency is required - guard that deps is a dict, log key is present and not (::)
+  logdict:$[99h=type deps;$[(`log in key deps)and not(::)~deps`log;deps`log;()!()];()!()];
+  if[not count logdict;
+    '"di.html: log dependency is required; pass `info`warn`error functions - see di.log for a default implementation";
+    ];
+  .z.m.lginfo:logdict`info;
+  .z.m.lgwarn:logdict`warn;
+  .z.m.lgerr:logdict`error;
+
+  / store config
+  .z.m.homedir:config`homedir;
+
+  / register .h content type handlers - protected in case not available in kdb-x
+  @[{.h.tx[`non]:{enlist x};.h.ty[`non]:"text/html"};`;{[e]}];
+
+  / register handlers via dep if provided
+  if[`handlers in key deps;
+    deps[`handlers][`register][`.z.ws;`html.ws;wshandler];
+    deps[`handlers][`register][`.z.wc;`html.close;closehandle];
+    :()];
+
+  / no handlers dep: assign directly, wrapping any existing .z.wc to preserve it
+  .z.ws:wshandler;
+  .z.wc:@[value;`.z.wc;{{}}];
+  .z.wc:{[existing;h] closehandle h; existing h}[.z.wc;];
+
+  .z.m.lginfo[`html;"initialised with homedir: ",homedir];
+  };
--- a/di/html/html.q
+++ b/di/html/html.q
@@ -15,7 +15,16 @@ ipacache:(`int$())!`symbol$();
 / home directory for html files - set by init
 homedir:"";
 
-/ logging functions - required injected deps, wired by init via .z.m
+/ flag so direct .z handler wiring happens only once across repeated init calls
+zwired:0b;
+
+/ logging functions - default to the console, overridable via the log config key on init
+
+/ default console loggers, called as (ctx;msg)
+deflog:`info`warn`error!(
+  {[c;m] -1 "INFO  ",(string c)," ",m;};
+  {[c;m] -1 "WARN  ",(string c)," ",m;};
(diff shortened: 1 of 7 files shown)


di/html/html.q
```diff
@@ -0,0 +1,182 @@
+/ di.html - websocket pub/sub and html page serving
```
DI-Software-Engineering, line 1: The entire file operates in the default (root) namespace — there is no `\d .di.html` / `\d .` pair wrapping the module code. In TorQ convention every module file must open with `\d .namespace` and close with `\d .`, otherwise all symbols (subtables, subs, modifier, zwired, normlog, etc.) are injected directly into the root namespace and will collide with any other module or user code.

DI-Software-Engineering, line 1: Single-slash `/` is used throughout for comments. In q, a `/` at the start of a line opens a multiline comment block when not preceded by a space. These single-slash comments work only because they appear after code or at the start of specific lines, but it is fragile and violates TorQ convention. Use `//` for inline/line comments.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+
+/ list of table names registered for pub/sub
+subtables:`symbol$();
```
DI-Software-Engineering, line 4: All module state and functions are defined in the root namespace (no `\d .di.html` / `\d .` namespace guard). TorQ convention requires every file to open with `\d .ns` and close with `\d .`. Without this, all names pollute the global namespace and can clash with other modules.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+closehandle:{[handle]
+  / removes the given handle from all subscriber lists when a connection closes
+  del[;handle] each subtables;
+  };
+
+execdict:{[inputdict]
+  / extracts the func key and any additional args from a dictionary and calls the function
+  / args are passed to the function in the order the keys appear after func
```
DI-Software-Engineering, line 98: `execdict` calls `f @ 1` when the input dict has only the `func` key (i.e. `1=count key inputdict`). This passes the integer `1` as the sole argument to `f`, not a call with no args. The intention is clearly a no-argument call (`f[]`). The `@ 1` form means "apply f to 1", so `neg` would return `-1` instead of signalling arity; but for any function that does not expect an integer arg this silently produces wrong results or a type error.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  / checks the module funcmap first (module-local functions), then falls back to value for globals
+  if[not `func in key inputdict;'"no func in dictionary"];
+  fname:`$inputdict`func;
```
DI-Software-Engineering, line 101: `wshandler` is defined before `evaluate` and `funcmap`, which it references. In q, function bodies are not closed over at definition time for names resolved via `value`/global lookup — this is fine at runtime. However `funcmap` is also referenced inside `execdict` (line 89), which is defined before `funcmap` on line 106. Since `funcmap` is a global variable assigned later, the first call to `execdict` before `init` could fail if `funcmap` has not yet been assigned. Ensure module load order guarantees `funcmap` is defined before any call path reaches `execdict`.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+
+jsformat:{[tbl]
+  / applies the correct javascript converter to each column of a table that needs it
+  / columns whose type is not in the typemap are passed through unchanged (via ::)
+  coldict:flip 0!tbl;
+  k:key coldict;
+  colvals:value coldict;
+  converters:typemap type each colvals;
+  :flip k!converters@'colvals;
+  };
```
DI-Software-Engineering, line 60: In `del`, `subs[tbl;;0]` projects the list of subscriber pairs to extract handles. If `subs[tbl]` is the empty list `()`, this projection will error (`'type` or `'rank`) rather than returning an empty list. The guard `if[not tbl in subtables;:()]` is absent here; `closehandle` calls `del` for every table on close, so a table with zero subscribers triggers this path.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+
+/ filter applied before sending data to a subscriber - returns full table (no filtering)
+sel:{[tbl;syms] tbl};
+
+del:{[tbl;handle]
+  / removes a handle from the subscriber list for a table
+  / if handle is not found, ? returns count of list and drop has no effect
+  idx:subs[tbl;;0]?handle;
+  .z.m.subs:@[subs;tbl;_;idx];
+  };
```
DI-Software-Engineering, line 76: `add` assigns `subs` via `.z.m.subs` but then reads `value tbl` directly. If `tbl` is not defined in the global namespace (e.g. a table registered via `addtables` before the global variable exists), `value tbl` signals an error and the subscriber has already been added to `subs` — leaving a dangling handle entry.

di/html/html.q
DI-Software-Engineering, line 101: `wshandler` sends `-8!.j.j[evaluate[...]]`. `.j.j` returns a string; `-8!` on a string produces the IPC-serialised bytes for that string, which is the c.js binary protocol. However, if `evaluate` signals an error (the trap in `evaluate` re-throws with `'`), the error will propagate out of `wshandler` uncaught, crashing the `.z.ws` handler and potentially closing the connection without sending an error reply to the client.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  .z.m.modifier:@[modifier;tbl;:;fn];
+  };
+
+evaluate:{[inputdict]
+  / safely calls execdict on the input, logging then re-throwing any errors with context
+  :@[execdict;inputdict;{[d;e] m:"failed to execute ",(-3!d)," : ",e;.z.m.log[`error][`di.html;m];'"di.html: ",m}[inputdict]];
+  };
+
+/ module-local functions callable via websocket evaluate
+/ execdict checks here first before falling back to value for global lookups
```
DI-Software-Engineering, line 155: `.z.wc` and `.z.pc` are assigned directly with `:` instead of using `.dotz.set`. This violates TorQ convention and will silently overwrite any handlers already set by the TorQ framework, even though the code attempts to preserve them by wrapping — the TorQ framework's own chain mechanism is bypassed.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  / sets a custom modifier function for tbl; fn receives (`upd;tbl;data) and must return bytes or a string to send
+  if[not tbl in subtables;'tbl];
+  .z.m.modifier:@[modifier;tbl;:;fn];
+  };
+
+evaluate:{[inputdict]
+  / safely calls execdict on the input, logging then re-throwing any errors with context
+  :@[execdict;inputdict;{[d;e] m:"failed to execute ",(-3!d)," : ",e;.z.m.log[`error][`di.html;m];'"di.html: ",m}[inputdict]];
+  };
+
```
DI-Software-Engineering, line 153: `.z.ws` is assigned directly instead of via `.dotz.set`, violating TorQ convention. Any existing `.z.ws` handler set by the framework is silently overwritten.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  / subscribes the current handle to a table with an optional sym filter
+  / pass backtick as tbl to subscribe to all registered tables
+  / removes any existing subscription for this handle before re-adding
+  if[tbl~`;:sub[;syms] each subtables];
+  if[not tbl in subtables;'tbl];
+  del[tbl;.z.w];
+  :add[tbl;syms];
+  };
+
+setmodifier:{[tbl;fn]
```
DI-Software-Engineering, line 143: `.z.m.log[`info][`di.html;...]` is called but `.z.m.log` is only set inside `init`. If `addtables` is called before `init`, this line will signal `'log` (undefined variable). The log call should be guarded or `log` should be initialised to a default no-op at module load time.

di/html/html.q
DI-Software-Engineering, line 1: The entire file is in the root (default) namespace — no `\d .di.html` / `\d .` wrapping. Per TorQ convention every module file must declare a namespace. All symbols and functions land in the global namespace, creating collision risk and violating the module boundary.

DI-Software-Engineering, line 1: Single-slash comments (`/ ...`) are used throughout. In q a single `/` on a line by itself opens a multi-line comment block; single-slash inline/line comments are a common source of silent comment-blocks when a blank line precedes them. TorQ convention requires `//` for line comments.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  .z.m.subs:$[(count subs tbl)>i;
+    .[subs;(tbl;i;1);union;syms];
+    @[subs;tbl;,;enlist(.z.w;syms)]
+    ];
+  :(tbl;$[99=type v:value tbl;sel[v;syms];@[0#v;`sym;`g#]]);
+  };
+
+closehandle:{[handle]
+  / removes the given handle from all subscriber lists when a connection closes
+  del[;handle] each subtables;
```
DI-Software-Engineering, line 93: `execdict` calls `f @ 1` when the input dict has only the `func` key (`1=count key inputdict`). The intent is to call `f` with no arguments (or call it monadically on the dictionary), but `f @ 1` passes the integer `1` as the argument. This will produce wrong results or a type error for any zero-argument function. It should be `f[]` for a niladic call, or the branch condition is wrong.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  idx:subs[tbl;;0]?handle;
+  .z.m.subs:@[subs;tbl;_;idx];
+  };
+
+add:{[tbl;syms]
+  / adds the current handle to the subscriber list for a table
+  / if already subscribed, updates the sym filter by taking union with new syms
+  / if not subscribed, appends a new (handle; syms) pair to the list
+  / returns (tablename; current data) so the subscriber can initialise their local copy
+  i:subs[tbl;;0]?.z.w;
```
DI-Software-Engineering, line 83: `converters:typemap type each colvals` looks up a type short in `typemap`. For columns whose type is NOT in `typemap`, the lookup returns a null (`::`). Then `converters@'colvals` applies `::` as a function to each element of the column, which projects the identity `::` over lists — this works in practice but is fragile. More critically, `type each colvals` returns a list of shorts (e.g. `11h` for a sym column), but `typemap` keys are `12 13 14 15 16 17 18 19h`. If a column type short is not in the map, `typemap[t]` returns `(::)` which is then applied. This is the intended design, but `colvals` is `value coldict` which is a list of column vectors; `converters@'colvals` applies a potentially mixed list of functions to column vectors — this is fine for mapped types but `(::)@col` returns `col` unchanged. The logic is correct, but `jstsiso8601` etc. are applied to entire vectors via `@'`, meaning each converter must be rank-1 (accept a whole vector). Verify `jstsiso8601` handles the full vector at once rather than scalar — the code does use `where`/bulk indexing, so this is consistent, but the `type each colvals` returns shorts like `-12h` for typed lists vs `12h` for general: for a typed timestamp vector `type` returns `-12h`, NOT `12h`, so the lookup `typemap[-12h]` will miss and the column will NOT be converted. The typemap keys are positive shorts; q typed lists have negative type codes.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  / columns whose type is not in the typemap are passed through unchanged (via ::)
+  coldict:flip 0!tbl;
+  k:key coldict;
+  colvals:value coldict;
+  converters:typemap type each colvals;
+  :flip k!converters@'colvals;
+  };
+
+dataformat:{[msgtype;msgdata]
+  / wraps a message into a name/data dictionary, javascript-formatting each table in msgdata
```
DI-Software-Engineering, line 63: In `del`, `subs[tbl;;0]` attempts to index into the list at column 0 (the handle). If `subs[tbl]` is an empty list `()`, this will signal a rank/type error rather than returning an empty list to search. The guard `if[not tbl in subtables;...]` is not present here; `del` is called from `closehandle` for every table regardless of whether the handle was subscribed, so an empty subscriber list will cause a runtime error.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  if[not `func in key inputdict;'"no func in dictionary"];
+  fname:`$inputdict`func;
+  f:$[fname in key funcmap;
+    funcmap fname;
+    @[value;inputdict`func;{'"unknown function: ",x}]];
+  args:value inputdict _ `func;
+  :$[1=count key inputdict;f @ 1;f . args];
+  };
+
+/ websocket message handler - module-level so it carries the module context for evaluate
```
DI-Software-Engineering, line 109: `wshandler` is defined at load time and captures its closure context. It calls `evaluate` which is defined later in the file. In q, function bodies are not evaluated until called, so forward references to names are resolved at call time — this is fine. However, `wshandler` sends the result of `evaluate` serialised with `-8!.j.j[...]`. If `evaluate` throws (re-throws with `'`), the error propagates uncaught out of `.z.ws`, which kdb+ will send as an error string to the client. This is probably acceptable but the behaviour differs from returning a structured error JSON response. More critically: `.j.j` is called on the raw q result of `evaluate`, which may include kdb+ types (tables, symbols, etc.) that `.j.j` cannot serialise cleanly — there is no error trap around the `.j.j` call itself.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+
+/ module-local functions callable via websocket evaluate
+/ execdict checks here first before falling back to value for global lookups
+funcmap:`sub`addtables`pub`dataformat!(sub;addtables;pub;dataformat);
+
+init:{[deps]
+  / initialise the module; deps must contain a log key
+  / deps: dict with `log key -> dict with at minimum `info!{[c;m]} (dyadic: ctx symbol, msg string)
+  / kx.log instances are normalised automatically via normlog
+  / wires .z.ws/.z.wc/.z.pc handlers once; .h.HOME set from KDBHTML env var (else "html")
```
DI-Software-Engineering, line 162: `.z.wc` and `.z.pc` are assigned directly (`:.z.wc:...` / `.z.pc:...`) instead of using `.dotz.set`. TorQ convention requires `.dotz.set` so that multiple modules can safely compose handlers. Direct assignment silently overwrites any handler set by another module.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+evaluate:{[inputdict]
+  / safely calls execdict on the input, logging then re-throwing any errors with context
+  :@[execdict;inputdict;{[d;e] m:"failed to execute ",(-3!d)," : ",e;.z.m.log[`error][`di.html;m];'"di.html: ",m}[inputdict]];
+  };
+
+/ module-local functions callable via websocket evaluate
+/ execdict checks here first before falling back to value for global lookups
+funcmap:`sub`addtables`pub`dataformat!(sub;addtables;pub;dataformat);
+
+init:{[deps]
```
DI-Software-Engineering, line 158: `.z.ws` is assigned directly instead of using `.dotz.set`, violating TorQ convention and potentially overwriting a handler already registered by another module (e.g. a TorQ WebSocket gateway).

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+pub:{[tbl;data]
+  / publishes data to all current subscribers of a table
+  / applies the table modifier before sending (default modifier json-encodes the data)
+  {[tbl;data;s]
+    if[count data:sel[data;s 1];
+      (neg first s) modifier[tbl]@(`upd;tbl;data)];
+    }[tbl;data] each subs tbl;
+  };
```
DI-Software-Engineering, line 131: In `sub`, when `tbl~\`` the function recurses with `sub[;syms] each subtables` and returns a list of `(tablename; data)` pairs — one per table. The caller in `add` returns a pair `(tbl; data)`. However the `.z.ws` handler does `neg[.z.w] -8!.j.j[evaluate[...]]` — if `sub` is called with `` ` `` the result is a list of pairs, not a single pair, and `.j.j` serialises the outer list. The browser-side `wssubTo` handler does `Array.isArray(r[0]) ? r : [r]` to handle this. This is a fragile protocol — a list-of-pairs vs a single pair relies on the JS-side type check. Not a crash bug, but the ambiguous return type is a correctness concern.

di/html/html.q
```diff
@@ -0,0 +1,182 @@
...
+  r[i]:dd,'("T",/:string `second$x i),\:"Z";
+  :r;
+  };
+
+/ converts a list of dates to javascript epoch milliseconds
+/ kdb dates are days since 2000-01-01; 10957 is the day offset from 1970-01-01 to 2000-01-01
+jstsfromd:{[x] "j"$86400000 * 10957 + `long$x};
+
+/ converts a list of time, second, or minute values to milliseconds since midnight
+jstsfromt:{[x] "j"$"t"$x};
```
DI-Software-Engineering, line 43: `jstsiso8601` uses `count each d:string \`date$x` and then `where 10=count each d` to find non-null entries. A null timestamp cast to date gives `0Nd`; `string 0Nd` is `""` (empty string, count 0), not length 10. Non-null dates stringify to 10 chars (`"2024-01-02"`). So null filtering is correct. However, `dd[;4 7]:" -"` is a single assignment attempting to set positions 4 and 7 of each string to `"-"` — this requires `dd` to be a list of strings of equal length, which is guaranteed by the `where 10=count each` filter. The issue is the assignment `dd[;4 7]:" -"` sets both positions to the same value: q interprets `" -"` as a 2-char string, assigning char `" "` to position 4 and `"-"` to position 7. The original date strings from `string` already contain `"."` at those positions (e.g. `"2024.01.02"`). The intent is to replace `"."` with `"-"`, so both positions should be `"-"`. But `dd[;4 7]:"-"` would assign the single char `"-"` to both positions. `dd[;4 7]:" -"` assigns `" "` to position 4 and `"-"` to position 7, resulting in `"2024 01-02"` — a space instead of a dash at position 4. The correct assignment is `dd[;4 7]:2#"-"`.

di/html/integrationtest.q
```diff
@@ -0,0 +1,58 @@
...
+/ starts a plain-text-mode websocket process for browser testing without c.js
+/ usage: q di/html/integrationtest.q (from kdbx-modules root, after setting QPATH)
+/ pass -p PORT to use a specific port; otherwise the OS assigns a free port
+
+if[0=system"p"; system"p 0"];
+
+html:use`di.html;
+
+logdep:`info`warn`error!(
+  {[c;m] -1 "[INFO] ",string[c]," ",m};
```
DI-Software-Engineering, line 11: Raw output via `-1` and `-2` is used directly instead of routing through the log dependency or `.lg.*`. TorQ convention requires `.lg.o`/`.lg.e` for process output.

di/html/integrationtest.q
```diff
@@ -0,0 +1,58 @@
...
+  syms:n?`AAPL`MSFT`GOOG`AMZN;
+  data:$[t=`trades;
+    ([]time:tm;sym:syms;px:100f+n?1f;sz:n?100);
+    ([]time:tm;sym:syms;bid:99f+n?1f;ask:101f+n?1f)
+    ];
+  msg:.j.j `name`data!("upd";`tablename`tabledata!(t;data));
+  tsubs:(get ` sv hns,`subs) t;
+  if[count tsubs; {[m;s] (neg s 0) m}[msg;] each tsubs];
+  };
+
```
DI-Software-Engineering, line 54: `p:system"p"` returns a long integer; `-1 "...port ",p` will fail with a type error because string concatenation with `,` requires both sides to be strings/char lists and `p` is a long. It should be `-1 "...port ",string p`.

di/html/html.q
DI-Software-Engineering, line 83: `type each colvals` — for a typed vector (e.g. a timestamp column), `type` returns the negative short (e.g. `-12h`), not `12h`. The `typemap` keys are `12 13 14 15 16 17 18 19h` (positive shorts). Therefore `typemap[-12h]` is a null `(::)`, and ALL typed columns will be left unconverted by `jsformat`. Only untyped (mixed/general) lists would match positive type codes, which is never the case for well-formed table columns. The fix is to use `abs type each colvals` when doing the typemap lookup.

DI-Software-Engineering:
## DIReview Summary

**2 critical** | **11 warning(s)** | **0 suggestion(s)**

di/html/html.q
Ruairi-wq2: This is inconsistent with style guide: '- Standalone comments (on their own line, before the code they describe): start with / followed by a space, lowercase first letter.
- End-of-line / trailing comments (appended after code on the same line): use //, not /.'

Ruairi-wq2:
Further to my comment about client side usage of the html doc, as far as I can see, TorQ only uses html common code when serving monitor UI. Heartbeat.q publishes to the monitor process when hb.subenabled is true. So essentially monitor.q is sole loader of code/common/html.q and has a packaged html doc for displaying health of registered processes.

PR #113, opened 2026-07-13
Feature config

ascottDI:
# di.config — TorQ Modularisation PR

## Summary

Extracts TorQ's in-process configuration handling (`torq.q`: `loadf`, `loadconfig`,
`loadaddconfig`, `overrideconfig`) into a standalone kdb-x module: `di.config`. The module
loads a process's q settings files from disk, resolves the override cascade
(`default → proctype → procname`, framework → app), and exposes the resolved config for
querying. `di.torq` will use it to partition config per module and inject each slice into
that module's `init`. It satisfies the di.* contract: a required injected logger, a minimal
exported API, no hard module dependencies, and a full k4unit suite.

> **Status: complete and tested in isolation, not yet deployable in a live stack.** See
> **Caveats** below — di.config needs the (unmerged) `di.log` logger and the (not-yet-built)
> `di.torq` orchestrator before it runs in a real process.

---

## Background

In TorQ, `torq.q` reads a cascade of settings files at start-up and every module pulls its
own values out of global namespaces via `@[value;\`var;default]`. Config resolution was
entangled with the monolithic bootstrap.

Per the modularisation plan, config is now a **hybrid**: `di.config` handles the mechanics
(load files, resolve the cascade, expose a queryable store); `di.torq` handles distribution
(partition by module, inject via each module's `init`). Modules receive a config dictionary
and don't know or care where it came from.

---

### New files

| File | Description |
|---|---|
| `di/config/config.q` | Implementation — `init`, `loadcascade`, `overrideconfig`, `getmodule`, and internal helpers `loadconfig`, `applyoverride`, `raiseerror` |
| `di/config/init.q` | Module entry point — loads `config.q`, declares the export |
| `di/config/test.csv` | k4unit suite (45 assertions) |
| `di/config/config.md` | Module README — API, design notes, and cross-module follow-ups |

### Differences from the TorQ original

| Aspect | TorQ | `di.config` |
|---|---|---|
| Logging | Hard-coded `.lg.o`/`.lg.e` | Injected binary `log` dep, required at `init`, **no fallback and no adaptation** |
| Config values | `@[value;\`var;default]` per module | Loaded into namespaces; read via `getmodule`; distributed by `di.torq` |
| Cascade order | Dir-major (per the actual `torq.q` code) | **Name-major** — specificity wins over directory layer (matches the plan's documented file order) |
| Env / identity | Reads `getenv` + `process.csv` internally | **Env-free** — caller supplies the settings dirs and the name sequence |
| Directory loading | `loaddir` (loads code dirs via `order.txt`) | **Not included** — whole-directory/code loading is `di.torq`'s concern, never used by the config cascade |
| Command-line overrides | `.proc.override` (reads `.z.x`) | `overrideconfig[params]` — caller passes the params; per-variable log-and-skip |
| Module contract | None | kdb-x `use` singleton, `init[deps]`, `export:` list |

---

Logging is an injected dependency, wired via `init` — there is **no default logger** and the
module does **no adaptation** of the logger. `init` **must** be called before any other
function.

- **`init[deps]`** takes a dict carrying the required `log` dependency. It errors immediately
  (plain signal, `di.config:` prefix) if `deps` is not a dict, is missing `` `log ``, `log` is
  not a dict, or `log` lacks any of `` `info`warn`error ``.
- The `log` value must **already** be a binary `` `info`warn`error!{[c;m]} `` dict — each
  function takes a context symbol `c` and a message string `m`. Build it from **`di.log`** (the
  standard logger, which exports binary `info`/`warn`/`error` directly) or hand-roll one:
  ```q
  logger:use`di.log
  config.init[enlist[`log]!enlist `info`warn`error!(logger.info;logger.warn;logger.error)]
  ```
  A logger that is *not* already binary `{[c;m]}` (e.g. a raw monadic `kx.log` instance) must be
  wrapped by the caller first — di.config does not wrap it. *(This is the settled no-`normlog`
  direction: adaptation belongs in `di.log`, not in every consumer.)*
- `init` fans the injected `log` dict out into three module-local functions —
  `.z.m.loginfo`/`.z.m.logwarn`/`.z.m.logerr` — and every call site uses
  `` .z.m.loginfo[`ctx;"msg"] `` (the three-flat-var convention from `consistency.md` and
  `di.compression`; maps 1:1 with TorQ's `` .lg.o[`ctx;"msg"] ``).
- Errors route through an internal `raiseerror[ctx;msg]` that logs via `` .z.m.logerr ``
  **then** signals `'"di.config: ",ctx,": ",msg` — so every failure is observable in the log as
  well as thrown.

---

## Exported API

```q
config:use`di.config

config.init[deps]                      / wire the required log dependency (call first)
config.loadcascade[dirs;names]         / load+resolve the cascade over dirs x names (name-major)
config.overrideconfig[params]          / command-line layer: di.torq parses .Q.opt .z.x and applies it after the cascade
config.getmodule[namespace]            / a whole namespace's config as a dict; index it inline for one value
```

- `dirs` — a settings-directory path or list of paths (low→high priority).
- `names` — a config name or list (least→most specific), e.g. `` `default`proctype`procname ``.
- `namespace` — a **bare** symbol, no leading dot (`getmodule[\`rdb]`); index the returned dict for one value.

Typical caller (di.torq): `config.loadcascade[(kdbconfig;appconfig);\`default,proctype,procname]`
→ `config.overrideconfig[cliparams]` → `config.getmodule[\`rdb]` → `rdb.init[…]`.

---

## Key design decisions

- **Name-major cascade.** For each name (least→most specific) load `dir/{name}.q` from each
  directory in turn, so a more specific name wins over the directory layer, and within a name the
  higher-priority directory overrides. This matches the plan's documented file order and is a
  deliberate departure from the actual `torq.q` code (which is dir-major).
- **Config store = live namespace globals (1a).** Settings files assign globals when executed, so
  those namespaces *are* the store; `getmodule` reads them (and reflects `overrideconfig`
  changes). An explicit store with provenance / cross-source precedence (1b) is deferred until a
  second config *source* (env vars, k8s) exists — the `getmodule` contract is
  storage-independent, so that swap won't touch consumers. Flagged as an `EXTENSION POINT` in code.
- **Env-free by design.** di.config reads no environment variables or process identity — `di.torq`
  resolves `KDBCONFIG`/`KDBAPPCONFIG` → dirs and `proctype`/`procname` → names and passes them
  explicitly. Deliberately no zero-arg `load[]` (config loads once; a blank call would only add
  hidden state and temporal coupling).
- **`overrideconfig` is the command-line layer, not a manual step.** It ports TorQ's
  `overrideconfig`/`override[]` (`torq.q`), which TorQ applies at startup off `.Q.opt .z.x` so
  launch-time flags (`-loglevel`, `-tablelist`, …) beat the settings files. di.torq owns the
  command-line parse and calls this after `loadcascade`; di.config just does the type-aware apply
  (staying env-free). It **improves on TorQ** on error handling: per-variable problems (undefined
  name, non-basic type, value that won't parse) are logged and **skipped** — a single bad override
  never aborts the batch and a null is never written into config.
- **Minimal API.** `loaddir` was dropped (code-directory loading, not config). `loadfile` was merged
  into the internal `loadconfig` (its string-check was dead and its existence check duplicated once
  the loader was internal-only). A single-key `get[namespace;key;default]` getter was also dropped —
  `getmodule` returns the slice and callers index it inline, and every config boots with a default so
  a fallback arg was dead weight (a missing key is now a caller bug, surfaced as a q null). Public
  surface is 4 functions; `loadconfig`/`applyoverride`/`raiseerror` stay internal.
- **KDB-X reserved-word workaround** (an `'assign` trap not listed in `.Q.res`): the cascade driver
  is `loadcascade` (not `load`, which is reserved and throws `'assign` at parse time).

---

## Test coverage — `test.csv` (k4unit), 45 assertions, all green

| Area | Coverage |
|---|---|
| `init` — dependency validation | rejects non-dict `deps`, missing `log`, non-dict `log`, `log` dict missing a key |
| `loadcascade` | dir-name grid loaded; name-major precedence (specific name beats dir layer); returned paths include skipped/missing; missing cascade file skipped at *info*; dedup on re-run; single-string/single-symbol accepted; type-error rejection |
| `overrideconfig` | bool/long/symbol-list parsed from strings and set; returns the vars overridden; undefined / non-basic-type / unparseable logged and skipped; empty params no-op; non-dict and non-symbol-key rejection |
| `getmodule` | whole-namespace dict; slice carries the values; empty dict for unknown namespace; reflects `overrideconfig` changes; non-symbol-namespace rejection |
| logger emission | a real logger (kx.log wrapped caller-side into the binary contract) receives `di.config`'s `init`/cascade/load messages via a file sink — proves the injected-logger path end to end |

### Running the tests

```bash
```

```q
k4unit:use`di.k4unit;
k4unit.moduletest`di.config;   / prints the results table; "All tests passed" on success
```

---

## Caveats — not deployable in a live process until two things land

di.config is **complete and tested as a standalone unit**, and its logging path was verified
end-to-end against the real `di.log` with **no code change**. It cannot yet run inside a live
process because:

1. **`di.log` (the standard logger) is not merged and not final.** di.config requires a conforming
   binary `` `info`warn`error `` logger injected at `init`; `di.log` provides exactly that (verified),
   but it currently lives on the `feature-logging` branch and its contract isn't approved. If that
   contract changes, re-verify — **no di.config code change is expected** (di.config only needs the
   three binary functions). Interim callers can hand-roll a logger.
2. **`di.torq` does not exist yet.** The orchestrator that reads process identity, resolves the
   config dirs, calls `loadcascade`, applies `overrideconfig`, and distributes config via `getmodule`
   hasn't been built. `getmodule` is unit-tested for that role, but the real partition-and-inject flow
   is unproven until `di.torq` lands.

### Deferred follow-ups (documented in `config.md`)

- Add a **committed `di.log` integration test** once `di.log` merges (verified manually for now; a
  `` use`di.log `` test can't be committed on this branch yet).
- Add a **`version` export + `deps.q`** when `di.depcheck` ships (repo-wide versioning rollout — no
  module carries these yet).
- Implement the **1b explicit config store** (provenance / cross-source precedence) if/when a second
  config source (env vars, k8s) is added — behind the unchanged `getmodule` contract.

### Test results screenshot

image


di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+/ public api:
+/   loadcascade    - resolve the full cascade over dirs x names (name-major)
+/   overrideconfig - apply command-line-style overrides after the cascade
+/   getmodule      - query the resolved config store as a per-namespace dict; di.torq
+/     partitions config per module with it and injects each slice via that module's init
+/ internal helper loadconfig handles per-file loading behind loadcascade.
+
+raiseerror:{[ctx;msg]
+  / internal - log an error under ctx then signal it, so failures are observable
+  / in the log as well as thrown to the caller.
```
DI-Software-Engineering: `raiseerror` calls `.z.m.logerr` before `init` has been called — on first invocation (e.g. if `init` itself fails its own type-check on line 32) `.z.m.logerr` is undefined, so the signal inside `raiseerror` is never reached and the process instead throws a `value` error on the undefined name. Use a guard: `@[.z.m.logerr;(ctx;msg);{[x;y]x}[msg]]` or check `@[{.z.m.logerr;1b};::;0b]` before calling it.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+  / it. raw is a string or a list of strings. returns 1b if applied, 0b if the
+  / variable is not a basic type or a value failed to parse (null). the variable
+  / must already exist - its current type drives the parse.
+  t:type value name;
+  if[not (abs t) within (1;-1+count .Q.t);
+    .z.m.logerr[`overrideconfig;"cannot override ",(string name),": not a basic type"];
+    :0b;
+  ];
+  raw:$[10h=type raw;enlist raw;raw];
+  vals:(upper .Q.t abs t)$'raw;
```
DI-Software-Engineering, line 124: `(upper .Q.t abs t)$'raw` uses the character type-letter of the *current* type to cast the string. For a **boolean** (`abs t` = 1, `.Q.t[1]` = `'b'`), `"B"$'` on a string like `"1"` returns a boolean vector correctly. However for **char** (`abs t` = 10, `.Q.t[10]` = `'c'`), `"C"$'` (uppercase) parses the string as a character rather than casting, which is the intended behaviour — but the null check (`any null vals`) that follows will pass on a valid char, giving a silent mis-result when the raw string is the empty string `""`. The bigger practical hazard: for type 10 (char), a config variable that is a char scalar (`t = -10h`) will hit `first vals` and return a char from the first *character* of the raw string rather than the whole string. Add explicit guards for char (`abs t within 10 10`) to skip or handle separately.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+  if[t<0;vals:first vals];
+  if[any null vals;
+    .z.m.logerr[`overrideconfig;"cannot override ",(string name),": value did not parse"];
```
DI-Software-Engineering, line 126: `any null vals` is used to detect parse failures, but `0N=0N` is `0b` in q — `null` correctly uses the `null` built-in, so that part is fine. The actual bug is that for a **boolean** parse failure `"B"$"bad"` returns `0b`, which is **not null** (`null 0b` is `0b`), so a value that cannot be parsed to a boolean silently sets the variable to `0b` instead of being skipped. The guard must be `any null each vals` or use a protected cast with `@["B"$;v;{0Nh}]` per element, then check for the sentinel.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+    raiseerror[`overrideconfig;"params must be a dict keyed by variable name"];
+  ];
+  vars:key params;
+  if[0<count vars;
+    if[not 11h=abs type vars;
+      raiseerror[`overrideconfig;"params keys must be symbols (variable names)"];
+    ];
+  ];
+  defined:vars where {@[{value x;1b};x;0b]} each vars;
+  undefined:vars except defined;
```
DI-Software-Engineering, line 157: The `defined` filter uses `@[{value x;1b};x;0b]` with a literal `value` call. `value` on a symbol returns the value of the global variable, but if the symbol names a **namespace** rather than a variable (e.g. `` `.rdb ``) it returns the namespace dict and evaluates to `1b`, incorrectly treating the namespace itself as a defined variable. This will let `overrideconfig` attempt to override a namespace symbol, which then fails inside `applyoverride` at `type value name` with an unexpected type. Filter more precisely with `@[{-1<type value x;1b};x;0b]` or equivalent to exclude namespaces (type 99h for a namespace dict is not basic).

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+/ additional config sources - environment variables, k8s config maps, external
+/ key-value stores - would plug in by populating the same root namespaces before
+/ the store is queried. keep source reading (loadcascade et al.) separate from
+/ querying (below) so a new source is an additive step, not a change to the
+/ getmodule contract.
+
+getmodule:{[namespace]
+  / return all resolved config for a namespace as a bare-keyed value dict, for
+  / di.torq to partition and pass to a module's init. namespace is a bare symbol
+  / with no leading dot; an unconfigured namespace yields an empty dict.
```
DI-Software-Engineering: `system"v ",x` (called inside `@[…]`) lists variables in the given namespace. The outer `@[…;".",string namespace;`$()]` trap catches failures, but the `vars` result is a symbol list of *short names* (without the leading dot-namespace prefix). The subsequent `value ` sv (ns;v)` joins them back via `` ` sv ``, which produces e.g. `` `.rdb.subscribeto `` — this is correct. However, if any name in the namespace is itself a sub-namespace (its `type` is 99h), `value ` sv (ns;v)` returns a dict, not a scalar/vector, and the result of `vars!{…} each vars` will be a mixed dict containing a namespace dict as a value. `getmodule` then silently returns that polluted dict to the caller. Add a filter: `vars where {-1<type value ` sv (ns;x)} each vars` to exclude sub-namespaces.

di/config/init.q
```diff
@@ -0,0 +1,3 @@
+/ configuration loading and cascade resolution for the modular torq world.
+\l ::config.q
```
DI-Software-Engineering, line 2: `\l ::config.q` uses a hard-coded relative path with the `::` escape. This is a hard-coded path that is not portable across working-directory contexts and should use `.os.pth` or a config-driven path resolution consistent with TorQ's module loader conventions.

DI-Software-Engineering:
## DIReview Summary

**2 critical** | **4 warning(s)** | **0 suggestion(s)**

di/config/config.q
ascottDI: raiseerror did assume .z.m.logerr was wired. Fixed by guarding the log call so it no-ops when unset but still signals:


.[{.z.m.logerr[x;y]};(ctx;msg);{}];

di/config/config.q
ascottDI: Verified against the module: neither hazard reproduces. A multi-char string into a char scalar is rejected by the existing any null vals guard (logged "value did not parse"), not truncated to the first char; an empty string hits the same guard. Nothing is silently mis-set, and no type corruption occurs. The valid case, a single-char string into a char scalar, is applied correctly with type preserved, so a blanket abs t within 10 10 skip would actually regress that.

The genuine limitation is that string-typed (10h) config vars can't be overridden (the null guard rejects them), this fails safe and is already documented as a known edge in config.md. Overriding a string/char config var from the command line is rare enough that I'd leave it; happy to add explicit string support if there's a real use case.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+/ reads no environment variables or process identity itself - the caller (di.torq)
+/ supplies the settings directories, name sequence and any command-line overrides.
+/ public api:
+/   loadcascade    - resolve the full cascade over dirs x names (name-major)
+/   overrideconfig - apply command-line-style overrides after the cascade
+/   getmodule      - query the resolved config store as a per-namespace dict; di.torq
+/     partitions config per module with it and injects each slice via that module's init
+/ internal helper loadconfig handles per-file loading behind loadcascade.
+
+raiseerror:{[ctx;msg]
```
JamesMassey97: raiseerror is getting called within conditionals in most cases, we wouldn't need to create a function for this, you can simplify this by calling it this way: 

if[cond;.z.m.logerr[`module;err:"error message"];'err];

there are similar examples already in compression.q if you need a working example.

di/config/config.q
ascottDI: Confirmed and fixed. Booleans (and bytes) are the only in-scope basic types with no null, so "B"$"bad" -> 0b (and "x" -> 1b) slips past the any null vals check and silently sets the variable, verified a 1b flag getting overwritten to 0b by garbage input.

Note the suggested fixes didn't work for this case: null each doesn't help because booleans/bytes have no null to detect; and the protected cast @["B"$;v;{0Nh}] never fires its handler because "B"$"bad" doesn't throw, it returns 0b. Fixed instead with a type-aware check: boolean raw must be "0"/"1", byte must be even-length hex, all other types keep the null check (which correctly catches them). Bad values are now logged and skipped, never written

di/config/config.q
ascottDI: Checked end-to-end.

It already fails safe, so leaving as-is

di/config/config.q
ascottDI: Confirmed and fixed. \v does list child namespaces, so getmodule was returning e.g. .rdb.sub as a nested 99h dict in the config slice

The suggested -1<type filter is backwards, though: a namespace is 99h (>-1, so it'd be kept), while every basic scalar setting has a negative type (<-1, so it'd be excluded), it would drop .rdb.rows etc. and leave the namespace. Fixed instead by excluding only child namespaces, distinguished from a legitimate dict-valued setting (also 99h) by the null-symbol self-reference key that a namespace carries and a plain dict doesn't. The type check is guarded with a cond rather than and, since and evaluates eagerly and would run key on non-dict values.

di/config/init.q
ascottDI: This is the required module-local load convention, not a portability hazard. The :: prefix tells the KDB-X module loader to resolve the path relative to the module's own directory, independent of the process working directory — it's what makes loading cwd-portable, not what breaks it. Verified: use\di.configloads identically from/, /tmp, /home/ascott, /opt. It's mandated by consistency.md ("Paths") and the module Definition-of-Done ("\lpaths for module-local files must use::"). .os.pthdoesn't apply — that's a TorQ namespace we're extracting away from, and init.q runs before any dependency is injected, so nothing in.os` exists at load time.

di/config/config.q
```diff
@@ -0,0 +1,184 @@
...
+  / internal - log an error under ctx then signal it, so failures are observable.
+  / guard the log call: the logger is unset until init runs, so a public fn misused
+  / before init still signals the real message rather than a name error on logerr.
+  .[{.z.m.logerr[x;y]};(ctx;msg);{}];
+  '"di.config: ",string[ctx],": ",msg;
+  };
+
+init:{[deps]
```
DI-Software-Engineering, line 11: The error handler `{}` silently swallows all failures from the log call, including legitimate errors unrelated to the logger being unset (e.g. wrong argument types, typos in function name). Use a narrower guard — for example catch and re-signal anything that isn't a `'value` error — or check `null .z.m.logerr` / `count key .z.m` directly before calling, rather than swallowing every error unconditionally.

di/config/config.q
```diff
@@ -0,0 +1,184 @@
...
+  if[parsefailed[t;raw;vals];
+    .z.m.logerr[`overrideconfig;"cannot override ",(string name),": value did not parse"];
+    :0b;
+  ];
+  if[t<0;vals:first vals];
+  .z.m.loginfo[`overrideconfig;"setting ",(string name)," to ",-3!vals];
+  set[name;vals];
+  :1b;
+  };
+
```
DI-Software-Engineering, line 134: `vals:first vals` is now executed AFTER `parsefailed`, but the atom-vs-list coercion it performs is used by `set[name;vals]` below. For a scalar type (`t<0`), `vals` is still a 1-element list at the point `parsefailed` runs, so `any null vals` (the fallback branch inside `parsefailed`) works correctly for non-boolean/non-byte types. However, the reordering means the logged value in `-3!vals` will always be the list form for scalar types rather than the scalar — minor cosmetic issue — but more importantly, if a future caller depends on `vals` being an atom at that stage, this ordering is a latent trap. The real correctness concern: for a scalar boolean (`t=-1h`), `parsefailed` receives `vals` as a 1-item boolean list; `not all raw in (enlist"0";enlist"1")` is the guard used, which is correct. No data-corruption bug here, but the moved `if[t<0;vals:first vals]` should be documented or restored to before `parsefailed` to avoid confusion.

di/config/config.q
```diff
@@ -0,0 +1,184 @@
...
+hexchars:"0123456789abcdefABCDEF";
+
+parsefailed:{[t;raw;vals]
+  / internal - true if any raw string failed to parse to type t. boolean (1h) and byte
+  / (4h) are the only in-scope basic types with no null, so a bad parse ("B"$"bad" -> 0b,
+  / "X"$"gg" -> 0x00) slips past a null check and would silently corrupt config - validate
+  / their raw string form explicitly. every other type yields a null on a bad parse.
+  :$[1h=abs t;not all raw in (enlist"0";enlist"1");
+     4h=abs t;not all {(0=count[x] mod 2) and all x in hexchars} each raw;
+     any null vals];
```
DI-Software-Engineering, line 97: `parsefailed` receives `t` as the raw type from `type value name`, which is negative for atoms (e.g. `-1h` for a boolean atom, `-4h` for a byte atom). The check `1h=abs t` correctly uses `abs`, so boolean atoms pass. However, the byte hex-length check `{(0=count[x] mod 2) and all x in hexchars} each raw` validates raw strings that represent a hex byte. `"X"$"ff"` parses correctly to `0xff`, but `"X"$"f"` (odd-length) also yields `0x00` — yet the check rejects it (good). The problem is for a byte *list* (`t=4h`, positive): `"X"$'raw` yields a byte *list*, but each element of `raw` should be a 2-char hex string for a single byte. Multi-byte hex literals like `"0x0102"` are not standard; this may be fine in practice, but the check `count[x] mod 2` on each individual raw element is the correct thing here. No actual bug for the scalar case. — Withdrawing, this is not a clear bug.

di/config/config.q
```diff
@@ -0,0 +1,184 @@
...
+  ns:`$".",string namespace;
+  vars:@[{system"v ",x};".",string namespace;`$()];
+  cfg:vars!{[ns;v] value ` sv (ns;v)}[ns;] each vars;
+  / \v lists child namespaces alongside settings, so drop them - only leaf settings
+  / belong in a module's config slice. a child namespace is a 99h dict carrying a
+  / null-symbol self-reference key; a genuine dict-valued setting has no such key and
+  / is kept. guard the key lookup with a cond (not `and`, which evaluates eagerly and
+  / would run `key` on non-dict values).
+  :(where {$[99h=type x;(`) in key x;0b]} each cfg) _ cfg;
+  };
```
DI-Software-Engineering, line 177: The namespace-detection predicate `(`) in key x` relies on a null-symbol self-reference key being present in every child namespace dict. This is a TorQ/kdb+ convention but is not a language guarantee — a child namespace that has never had the null-symbol key set (e.g. a freshly created namespace with only plain keys) would not be filtered out. Conversely, a genuine dict-valued config variable that happens to contain a null-symbol key would be incorrectly dropped. The filter is fragile; a more robust approach is to compare `type value ` sv (ns;v)` against `-1!` or check `99h=type[v] and (`) in key v` only after confirming the value was produced by `system"v"` listing a sub-namespace (e.g. by re-running `system"v .",…` on it).

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **3 warning(s)** | **0 suggestion(s)**

di/config/config.q
ascottDI: Good point on the over-broad swallow, tightened. The catch-all {} is replaced with an explicit "is the logger wired?" check (if[@[{.z.m.logerr;1b};::;0b];.z.m.logerr[ctx;msg]]), so only the unset-before-init case is tolerated and a wired logger is now called directly, surfacing its own errors like every other log call in the module.

Two notes on the suggestions: re-signalling non-value errors would mask the domain error raiseerror exists to surface (the log failure would replace the actual validation message the caller needs), and null .z.m.logerr throws on the unset name before null runs. The "check before calling" route is the one that works, so that's what I used, and it makes raiseerror consistent with the module's existing unprotected .z.m.loginfo/logwarn call sites.

di/config/config.q
ascottDI: No behavior change needed here, added a clarifying comment. The -3!vals logging concern doesn't actually occur: if[t<0;vals:first vals] runs immediately before both the log line and set, so scalars are logged and stored in scalar form, verified the log reads "setting .t.rows to 500" and the stored value is type -7h. The ordering is deliberate: parsefailed's null check spans the full per-element list of parsed values, so the scalar reduction has to happen after validation, not before. Documented that inline.

di/config/config.q
ascottDI: Automated reviewer realised he was wrong mid message

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+    :file;
+  ];
+  .z.m.loginfo[`loadconfig;"loading ",file];
+  .[system;enlist"l ",file;{[f;e] raiseerror[`loadconfig;"failed to load ",f,": ",e]}[file;]];
+  .z.m.loaded,:enlist file;
+  :file;
+  };
+
+loadcascade:{[dirs;names]
+  / resolve a configuration cascade name-major: for each config name (least to
```
JamesMassey97: These descriptions are quite long on each function, can we make them more concise.

di/config/config.q
```diff
@@ -0,0 +1,184 @@
...
+     4h=abs t;not all {(0=count[x] mod 2) and all x in hexchars} each raw;
+     any null vals];
+  };
+
+applyoverride:{[name;raw]
+  / internal - parse raw command-line value(s) into name's current type and set it. raw
+  / is a string or a list of strings. returns 1b if applied, 0b if the variable is not a
+  / basic type or a value failed to parse. the variable must already exist - its current
+  / type drives the parse.
+  t:type value name;
```
JamesMassey97: Does this line work? I checked it with a symbol and string and got errors, what is the expected input of name?

di/config/config.q
ascottDI: Agreed, the self-key check was fragile, switched to asking kdb+ directly. Verified both edges are reachable: a namespace reassigned to a plain dict is still \v-listed but has no self-key, and a dict setting could coincidentally carry a null-symbol key. getmodule now tests each name with \v on its qualified name (@[{system"v ",x;1b};…;0b]), succeeds for a real namespace, signals for a plain variable, which sidesteps both edges. (Your -1! suggestion I read as a typo, so I went with the \v-requalify route you described alongside it.) Added a test where a dict-valued setting carries a null-symbol key and confirms it's kept, not dropped.

JamesMassey97 (dismissed):

di/config/config.q
ascottDI: Good idea, thanks James I have implemented this now

di/config/config.q
ascottDI: For this one you pass in the name of a variable you want to edit, 
if you call               `value ``rows` 
You will get an error, however if first you define rows such as 
`rows:500`
then when you run the previous line
`value ``rows`
`500`

you will get the output of what the variable is saved as and then we take the type of it

di/config/config.q
```diff
@@ -0,0 +1,163 @@
...
+    '"di.config: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  .z.m.loginfo:deps[`log]`info;
+  .z.m.logwarn:deps[`log]`warn;
+  .z.m.logerr:deps[`log]`error;
+  .z.m.loaded:enlist"";
+  .z.m.loginfo[`init;"di.config initialised"];
+  };
+
+loadconfig:{[dir;name]
+  / internal - load cascade file dir/{name}.q if present, tracking it so it is not
```
DI-Software-Engineering: `.z.m.logerr` is called unconditionally before the `init` function has run and wired the logger. The deleted `raiseerror` helper explicitly guarded against this with `if[@[{.z.m.logerr;1b};::;0b];...]`, but the new inline `logerr` calls carry no such guard. If `loadconfig` (or any other function) is called before `init`, the unset `.z.m.logerr` will cause a name-error crash instead of the cleaner signal that `raiseerror` provided. Restore the guard (e.g. wrap each `logerr` call in `if[@[{.z.m.logerr;1b};::;0b];...]`) or re-introduce a thin helper that replicates the original defensive check.

DI-Software-Engineering:
## DIReview Summary

di/config/config.q
ascottDI: Intentional. Not restoring the guard. This is the accepted trade-off from deliberately dropping the raiseerror helper in favour of di.compression's inline log-then-signal pattern. di.compression is unguarded pre-init in exactly the same way (its checkcsv/loadcsv call .z.m.logerr directly), and the framework contract is "init must be called before any other function," so a pre-init call is a caller bug rather than a supported path. Re-adding the guard would either re-introduce the helper we just removed or spread a per-site if[@[{.z.m.logerr;1b};::;0b];…] across every validation site, both of which diverge from the module this was aligned to. Leaving the inline calls unguarded for consistency.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+  / lazily (only when a .toml file actually appears, so the .q-only path never needs it).
+  / a .q file is plain `name:value` lines: split on the first ":", the RHS run through value
+  / (so `:hdb, `trade`quote, `symbol$() all work); blank and "/"-comment lines are skipped.
+  / a missing file of either extension contributes nothing (empty dict), so the cascade needs
+  / no presence checks. existence is checked BEFORE the .toml delegation so a missing .toml
+  / tier never forces di.toml to load - di.toml is pulled in only when a .toml file actually
+  / exists (di.toml lives only in the PoC today; the .q-only path needs it never).
+  fsym:`$":",path;
+  if[0=count key fsym; :()!()];
+  if[path like "*.toml"; :(use`di.toml)[`parsefile] path];
```
DI-Software-Engineering: The `/`-comment filter `lines like "/*"` matches any line whose first character is `/`, but the glob pattern `"/*"` means 'slash followed by zero or more characters' — in q, `*` in a `like` pattern is a wildcard, so this correctly matches lines starting with `/`. However, the check uses `not lines like "/*"` which also drops lines whose first character is a forward-slash that is part of a valid value (e.g. a path string `` /opt/foo `` on the RHS of a `name:value` pair). More importantly, the filter is applied to the whole raw line, not to lines that lack a `:`, so a value line like `path:/some/dir` is kept (correctly), but a line like `/ comment with : colon` would also be dropped only if it starts with `/` — that part is fine. The real bug is that blank-line and comment-line filtering happens BEFORE the split, but the split lambda `{[ln] i:first where ln=":"; ...}` will signal with a null `i` (and thus `0N#ln`) on any line that survived filtering but contains no `:`. If a settings file has a line that is neither blank, nor a comment, nor a `name:value` pair (e.g. a continuation, a stray word, or a line whose only `:` is inside a string), `first where ln=":"` returns `0N`, and `0N#ln` / `(0N+1)_ln` will error or return unexpected results. Add a guard: `lines:lines where ":" in' lines` after the comment filter to skip non-pair lines safely.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+  / clash if both exist (useful mid-migration). each missing file yields an empty dict.
+  (parsefile base,".q"),parsefile base,".toml"
+  };
+
+cascade:{[builtinroot;approot;proctype;procname]
+  / resolve the settings cascade over the two roots x the (default;proctype;procname) name
+  / sequence and RETURN the merged flat dict. pure - no init, no logger, no globals - so
+  / di.torq can call it directly after use`di.config, before the logger exists. precedence is
+  / name-major: a more specific NAME wins over the root layer (procname beats proctype beats
+  / default), and within a name the later (app) root overrides the builtin root - so a key set
```
DI-Software-Engineering: `cascade` builds `names` as `` `default,proctype,procname `` — if `proctype` and `procname` happen to be the same symbol (e.g. both `` `rdb ``), two identical tier bases are generated and `parsetier` is called twice on the same files, causing those files to be parsed and merged twice. The later (duplicate) merge overwrites the earlier one's keys with identical values, so the result is technically correct, but `parsefile` is called redundantly. More critically, if `proctype~`default` (a caller passes the literal name `` `default `` as proctype), the `default` tier is processed twice with the same precedence slot, which is misleading and could produce unexpected results if `parsetier` or `parsefile` has side-effects in future. Deduplicate `names` after construction: `names:distinct names`.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+  / in both builtin/{proctype} and app/default resolves to the builtin proctype value. within a
+  / single tier, .toml wins over .q (see parsetier). the accumulator is seeded with a sentinel
```
DI-Software-Engineering: `dirs,\:"/",string nm` concatenates each root with `"/"` and the string form of `nm` (e.g. `"rdb"`), but `string` on a symbol strips any leading backtick and returns the bare name. However if `proctype` or `procname` is the null symbol `` ` ``, `string[`]` returns `""`, producing paths like `"/tmp/builtin/"` — a directory, not a settings file base. `parsefile` on `"/tmp/builtin/.q"` would then attempt to read a file named literally `.q` (or `.toml`) in the root, which is almost certainly not intended. Validate that `proctype` and `procname` are non-null symbols before use, or guard in `cascade` with `if[(null proctype)|null procname; ...]`.

di/config/config.q
```diff
@@ -0,0 +1,165 @@
...
+    ];
+  ];
+  defined:vars where vars in key config;
+  undefined:vars except defined;
+  if[count undefined;
+    .z.m.logwarn[`overrideconfig;"skipping unknown setting(s): ",", " sv string undefined];
+  ];
+  / fold each defined override into the config dict; a skipped/rejected one leaves it unchanged
+  :{[config;params;name]
+     res:applyoverride[name;config name;params name];
```
DI-Software-Engineering, line 163: In the `overrideconfig` fold `:{[config;params;name] res:applyoverride[...]; $[res 0; @[config;name;:;res 1]; config]}[;params;]/[config;defined]`, the fold is written as `f/[seed;list]` but the lambda is a 3-argument function projected as `f[;params;]` — making it effectively binary when partially applied. The seed (`config`) is the accumulator and `defined` is the list. This is the correct q `f/[seed;list]` idiom. However, `applyoverride` is called as `applyoverride[name; config name; params name]` where `config` is the *current accumulator* from the fold, which is correct. The real risk is that `defined` is derived from the *original* `config` (`vars where vars in key config`), but if an earlier iteration of the fold changes the keys of config (which `@[config;name;:;res 1]` does not — it only updates a value for an existing key), this is safe. Confirmed no key addition occurs. No bug here — this finding is withdrawn.

di/config/test.csv
```diff
@@ -0,0 +1,62 @@
...
+true,0,0,q,`name`public`descrip`params`return~cols cfg.getapimeta[],1,1,getapimeta rows carry the registry columns
+true,0,0,q,1b~(1!cfg.getapimeta[])[`cascade]`public,1,1,a public API function (cascade) is marked public
+true,0,0,q,0b~(1!cfg.getapimeta[])[`init]`public,1,1,framework plumbing (init) is not public
+comment,,,,,,,logger emission - real logger via a caller-side binary wrapper (must run last; re-inits cfg)
+run,0,0,q,kxlogger:use`kx.log,1,1,load the kx.log module
+run,0,0,q,kxinst:kxlogger.createLog[],1,1,create a real kx.log instance
+run,0,0,q,kxh:hopen `:/tmp/diconfigkxsink.txt,1,1,open a file sink to capture output
+run,0,0,q,kxinst.add[kxh;`info],1,1,route info-level output to the file sink
+run,0,0,q,"kxbinary:`info`warn`error!({[c;m]kxinst[`info][string[c],"": "",m]};{[c;m]kxinst[`warn][string[c],"": "",m]};{[c;m]kxinst[`error][string[c],"": "",m]})",1,1,wrap the monadic logger into the binary {[c;m]} contract
+run,0,0,q,cfg.init[enlist[`log]!enlist kxbinary],1,1,re-init di.config with the conforming binary logger
```
DI-Software-Engineering, line 76: The kx.log emission test seeds the config dict inline: `enlist[`kxrows]!enlist 5` (the first argument to `overrideconfig`). This dict has a single long value `5` for key `` `kxrows ``, and the override applies `enlist"7"` for the same key. The test then asserts `7~kxres`kxrows`. Because `applyoverride` uses `type cur` (which is `-7h` for a long scalar `5`) and `"J"$'enlist"7"` → `enlist 7j`, then `if[t<0; vals:first vals]` reduces to scalar `7`, the assert is correct. However `enlist 5` is a 1-element long list (type `7h`), not a scalar long (type `-7h`): `enlist 5` has type `7h`, so `t:type cur` is `7h` (positive, a list), and `if[t<0; vals:first vals]` is NOT triggered — `vals` stays as `enlist 7j`. The returned dict would have `` `kxrows `` bound to `enlist 7j` (a 1-item list), so `7~kxres`kxrows`` would fail (`` 7~enlist 7j `` is `0b`). Fix the seed to use a scalar: `(enlist`kxrows)!enlist 5j` should be `(enlist`kxrows)!enlist[5j]` ... actually the issue is `enlist 5` produces a list. Change to `(enlist`kxrows)!5j` (scalar value, no `enlist` on the value side) so `t` is `-7h`.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **4 warning(s)** | **0 suggestion(s)**

di/config/config.q
```diff
@@ -0,0 +1,188 @@
...
+  / parse, and any string is valid). this makes .toml-origin string settings overridable, the
+  / same way symbol (.q-origin) settings already are; di.config stays policy-free by preserving
+  / whatever type the setting already had.
+  if[10h=t;
+    vals:first raw;
+    .z.m.loginfo[`overrideconfig;"setting ",(string name)," to ",-3!vals];
+    :(1b;vals);
+  ];
+  vals:(upper .Q.t abs t)$'raw;
+  if[parsefailed[t;raw;vals];
```
DI-Software-Engineering, line 125: The string-override branch takes `first raw` as the override value, but `raw` was already widened to a list on line 121 (`raw:$[10h=type raw;enlist raw;raw]`). If the caller passes a list of strings (e.g. a multi-value command-line flag), `first raw` silently discards all but the first element instead of rejecting or joining them. Fix by either rejecting `1<count raw` for string settings (like other scalar types) or explicitly handling the list case.

DI-Software-Engineering:
## DIReview Summary

di/config/config.q
ascottDI: Correctly identified. Fixed

di/config/config.q
ascottDI: Correctly identified, added distinct keyword to close gap.

di/config/config.q
ascottDI: Correct, added null check

di/config/test.csv
ascottDI: Incorrect assertion, not changes made.

di/config/config.q
ascottDI: Now requires exactly one, instead of accepting many.

di/config/config.q
```diff
@@ -0,0 +1,213 @@
...
+  / setting's current value from the config dict; its type drives the parse. raw is a string or
+  / list of strings. applied is 0b (and newvalue is cur, unchanged) if cur is not a basic type,
+  / a value failed to parse, or a single-valued (scalar/string) setting got other than one value.
+  t:type cur;
+  if[not (abs t) within (1;-1+count .Q.t);
+    .z.m.logerr[`overrideconfig;"cannot override ",(string name),": not a basic type"];
+    :(0b;cur);
+  ];
+  raw:$[10h=type raw;enlist raw;raw];
+  / a scalar-atom (t<0) or string (10h) setting is single-valued: require EXACTLY one override
```
DI-Software-Engineering: The guard `(1<>count raw) and (t<0)|(10h=t)` has an operator-precedence bug. In q, evaluation is right-to-left with no precedence, so this parses as `(1<>count raw) and ((t<0)|(10h=t))` — which is actually correct as written IF the intent is `(count raw <> 1) and (t<0 or t=10h)`. However, `t<0` is itself evaluated right-to-left: `t < (0|(10h=t))`, i.e. `t < (boolean result of 10h=t)`, which is NOT `(t<0)|(10h=t)`. The expression should be written with explicit grouping: `(1<>count raw) and ((t<0)|(10h=t))`, or restructured as `(1<>count raw) and (0>t)|10h=t` with the comparison isolated. Fix by adding explicit parentheses: `if[(1<>count raw) and (0>t)|10h=t;`

DI-Software-Engineering:
## DIReview Summary

di/config/config.q
ascottDI: Not an error, will make grouping explicit anyway.

jprgrant (approved):

PR #114, opened 2026-07-17
Feature handlers

alowrydi:
**Central `.z.*` handler registry, di.handlers**

Extracts `code/handlers/dotz.q`'s handler capture-and-install mechanism from TorQ and packages it as a standalone kdb-x module - the single sanctioned place a process assigns `.z.*` connection-lifecycle callbacks, replacing the per-file, load-order-dependent closure wrapping TorQ uses today, where whichever handler loads last silently overwrites whoever set the event before it.

The other nine files in `code/handlers/` - `writeaccess.q`, `controlaccess.q`, `permissions.q`, `trackclients.q`, `trackservers.q`, `logusage.q`, `zpsignore.q`, `ldap.q`, `apidetails.q` - were read as research evidence during design, not as source material this module extracts or consolidates. Their actual content belongs to other modules per the modularisation plan: `controlaccess.q`/`permissions.q`/`writeaccess.q`/`ldap.q` to `di.permissions`, `trackclients.q`/`trackservers.q` to `di.clienttracking`, `logusage.q` to `di.querylog`, `apidetails.q` to `di.api` (`zpsignore.q` remains unassigned in the plan). Reading all nine directly - not a subset - is what surfaced the real distinction this module is built on: some of TorQ's `.z.*` handlers fan out safely across multiple files, others are hard-coded to a single owner with real mutual-exclusion guards between them, and no single file shows both patterns at once.

Trello ticket - **https://trello.com/c/lMDyVar0/119-kdb-x-handlers-module**

---

**Files created**

| File | Description |
| -- | -- |
| `di/handlers/init.q` | Loads `handlers.q`, defines export of 7 functions |
| `di/handlers/handlers.q` | Full implementation - module state, internal helpers, decider-observer machinery, public API, init |
| `di/handlers/handlers.md` | Full module documentation - see below |
| `di/handlers/test.csv` | 61 k4unit unit test assertions (114 total rows including setup/comment rows) |
| `di/handlers/test_integration.csv` | 3 k4unit integration tests against a real spawned q process, OS-assigned port, self-cleaning on any exit path |

---

**How to test**

Unit tests:
```q
k4unit:use`di.k4unit
k4unit.moduletest`di.handlers
```
```
2026.07.17T10:43:06.123 start
2026.07.17T10:43:06.123 :.../di/handlers/test.csv 114 test(s)
2026.07.17T10:43:06.128 end
...
All tests passed
```
39 `true` + 22 `fail` assertions, 0 failures.

Integration tests (spawns a real second q process on an OS-assigned ephemeral port; skips cleanly if no usable q binary is found, no configuration required). `moduletest` only ever loads `test.csv`, so the integration suite is loaded and run directly:
```q
k4unit:use`di.k4unit
.m.di.0k4unit.KUltf .Q.dd[hsym`$.Q.m.mp`di.handlers;`test_integration.csv]
.m.di.0k4unit.KUrt[]
k4unit.getresults[]
```
```
action ms bytes lang code                             ... ok
true   0  0     q    .it.roundtrip                     ...  1
true   0  0     q    `closed in exec who from .hh.iev  ...  1
true   0  0     q    .it.gone                          ...  1
```
3/3 passing. (The `sh: kill: No such process` line sometimes seen during this run is expected, harmless output from the belt-and-braces forceful cleanup step finding the child already self-exited via its own `.z.pc` backstop - not an error.)

Unit test coverage includes: observer fan-out order and isolation on throw, the captured original always running last, duplicate-name replace-in-place, decider single-owner claim/reclaim/reject, decider `remove` restoring the kdb-x built-in default via `\x`, `.z.ph` rejected outright with a self-contained error, dependency validation (all four failure modes), `.z.pw` password redaction to observers (with the owner still receiving the real password), owner-throw stopping the chain before any observer runs, `register` reclaim after `observe` leaving the dispatcher installed, full state reset when an observed owner is removed, and init idempotency preserving live registrations across a second `init` call.

Integration test coverage confirms a real synchronous IPC round-trip works while the module is loaded, that a registered `.z.pc` observer fires on a genuine connection close (not just a synthetic call), and that the spawned child process is fully cleaned up afterward - proven against both the normal path and a deliberately forced mid-suite failure, since the child is instructed to self-exit whenever its connection to the test process drops, for any reason, rather than relying on the test process shutting down gracefully.

---

**Design decisions**

*1. Observer vs decider split, not a uniform chaining rule* - Not every `.z.*` event behaves the same way when more than one thing wants to hook it. For `.z.pc`, `.z.po`, `.z.exit`, `.z.wo`, `.z.wc`, kdb-x discards the return value, so any number of registrants can safely coexist - `register` fans out to all of them, each individually isolated (a throw is caught, logged at `warn`, and doesn't stop the rest). For `.z.pg`, `.z.ps`, `.z.pi`, `.z.pp`, `.z.pw`, `.z.ws`, kdb-x uses the return value as the actual outcome, so only one thing can coherently own it - `register` enforces single ownership, with same-name reclaim (needed for idempotent re-init) and different-name rejection naming the current owner. This split was derived from reading all of TorQ's `code/handlers/` files directly, not assumed - real TorQ resolves decider conflicts three incompatible ways depending on which files are involved (`permissions.q`'s hard mutual-exclusion exit against `controlaccess.q`, `ldap.q`'s AND-composition, `zpsignore.q`'s conditional bypass), so `di.handlers` doesn't invent a generic composition rule for deciders - it enforces exclusivity instead. Independently corroborated by KX's own shipped Delta Platform product (`.ch.addPC`/`.ch.addPO`/`.ch.addExit`), which formalises multi-handler chaining for exactly the observer-style events and stops there.

*2. `.z.ph` explicitly out of scope, not silently accepted* - On any current kdb-x version, `permissions.q` doesn't gate HTTP GET via `.z.ph` at all - it sets `.h.val:req` instead, a completely different hook outside `.z.*` entirely. Registering against `.z.ph` would silently do nothing on a real deployment, so `register`/`observe` both reject it outright with a self-contained error rather than accepting it as a normal decider.

*3. `observe`/`unobserve` - side-effect-only watchers on a decider's owner* - Single ownership on decider events is correct and unchanged, but real TorQ proves a decider event also needs pure watchers alongside its owner - `logusage.q` wraps every decider event with log-before/after logging that passes the result through untouched; `trackclients.q` wraps `pg`/`ps`/`ws` with byte-counting the same way. `observe[event;name;func]` attaches a watcher to a decider event that already has an owner; it runs after a successful owner call, receives `func[result;args]`, and can never alter what the caller actually receives - the dispatcher's own return statement is the owner's result, computed before any observer runs, so nothing an observer does or returns can reach the client. For `.z.pw` specifically, the owner receives the real password (it has to, to authenticate) but observers only ever receive `(user;"***")` - modelled directly on `logusage.q`'s own real redaction of the same field.

*4. OS-assigned ephemeral port for the integration test, no configuration* - The integration suite needed a real second process on a real port. Rather than a fixed range or an environment variable, the test asks the OS directly (`\p 0W`, confirmed against kdb-x's own documented ephemeral-port mechanism) and reads back whatever port the kernel actually assigned - guaranteed free at the moment of assignment, no configuration step, no bespoke variable. Cleanup is robust to a genuine mid-suite crash, not just the happy path: the spawned child is told to self-exit the instant its connection to the test process drops, for any reason, which was verified empirically to survive an uncaught throw partway through the suite, not just assumed to work.

*5. Export list reviewed against the public API criterion* - All seven exports (`init`, `register`, `remove`, `list`, `version`, `observe`, `unobserve`) are genuine consumer-facing functions. The dispatch machinery, the per-event state, and the decider-observer internals are all private to the module.

---

**Checklist**
- [ ] 61/61 unit test assertions passing (114 total test.csv rows)
- [ ] 3/3 integration tests passing
- [ ] `handlers.md` documents all exported functions, config, usage examples and notes
- [ ] Verified against no TorQ namespaces, env vars, or hard `use` imports anywhere in the module

---

**Documentation**

See `handlers.md` for full reference including the managed-events table, dependency table, exported function documentation with examples, usage example, and notes.


di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+
+dispatchobs:{[event;arg]
+  / fan out to every registered observer in registration order, then run the captured original last, each isolated
+  t:0!.z.m.registry event;
+  runobserver[event;;;arg]'[t`name;t`func];
+  runoriginal[event;arg];
+  };
+
+installobserver:{[event]
+  / on first registration for an event, capture whatever is currently bound (KDB-X default if unset) and install the dispatcher
```
DI-Software-Engineering, line 56: The observer dispatcher calls `runoriginal` unconditionally even when no original handler existed before the first registration. `installobserver` captures `@[value;event;{(::)}]`, so the fallback is the identity `(::)`. Calling `.[::;enlist arg;…]` applies `(::)` to `arg`, which is harmless for most argument types, but it is a spurious call that could produce unexpected behaviour if `arg` is a list (since `(::)` applied via `.` to a list may iterate). The captured value should be checked before calling — skip `runoriginal` when the captured original is `(::)`.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+
+registerobserver:{[event;nm;func]
+  / lazily install on first use, then add or replace this handler in registration order
+  if[not event in key .z.m.original;installobserver event];
+  .z.m.registry[event]:.z.m.registry[event] upsert (nm;func);
+  .z.m.loginfo[`register;"registered observer ",string[nm]," on ",string event];
+  };
+
+registerdecider:{[event;nm;func]
+  / single owner - a different name claiming an owned event is rejected; the same name reclaims (idempotent re-init)
```
DI-Software-Engineering: `installobserver` calls `set[event;dispatchobs[event;]]` to bind the dispatcher, but `dispatchobs[event;]` is evaluated at install time and produces a projection closed over the current state. If the module is re-initialised (`init` called a second time) and `initstate` is NOT re-run (the idempotency guard skips it), the already-installed `.z.*` handlers still hold the old projection, which reads `.z.m.registry` via the old closure. This is likely fine because `.z.m.*` names are used directly, but more critically: `set[event;…]` sets the event to a *value* (a closure), not to a name. `system"x ",string event` in `removedecider` (line 100) resets the event to the KDB-X built-in default, which is correct. However, `installobserver` never has a matching `removeobserver` equivalent that restores the original — `removeobserver` only removes from the registry table, it never restores `.z.event` to the captured original even when the last registered observer is removed. After the last observer is removed the dispatcher is still installed and still calls `runoriginal`, which is the pre-capture value. This is documented as intentional in the notes, so it is a design inconsistency rather than a crash bug, but it means there is no way to fully uninstall an observer-event dispatcher once it has been installed, leaking the dispatcher permanently.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+
+dispatchdeciderunary:{[event;x]
+  / unary decider dispatcher - call the single owner unprotected, then run side-effect observers, return the owner result
+  r:.z.m.ownerfunc[event] x;
+  rundeciderobsall[event;r;enlist x];
+  r
+  };
+
+dispatchdeciderpw:{[u;p]
+  / .z.pw is the only binary decider - the owner sees the real password; observers see it redacted to "***"
```
DI-Software-Engineering: `installdecider` does `value event` to capture the current owner function, but at this point `register` has already called `set[event;func]` (via `registerdecider`, line 83: `set[event;func]`). So `value event` captures the *dispatcher's own owner function* — actually `func` itself — which is correct on the first `observe` after the first `register`. However, if `register` is called again (reclaim, same name), `registerdecider` takes the `event in key .z.m.ownerfunc` branch and does `.z.m.ownerfunc[event]:func` without touching `.z.*`. But if `observe` was never called yet, the `event in key .z.m.ownerfunc` check is false and `set[event;func]` runs, which is correct. The real problem: if `register` (reclaim) is called *after* `observe` has already installed the dispatcher, `registerdecider` correctly updates `.z.m.ownerfunc[event]` in place. But `installdecider` is only called on the *first* `observe`, and it calls `value event` which at that point returns the raw owner `func` (since the dispatcher is not yet installed). This is actually correct. The bug is more subtle: `installdecider` calls `value event` which reads the *current binding* of the event symbol. If between `register` and the first `observe` something else rebinds `.z.pg` (e.g., another library), `installdecider` would capture the wrong function as the owner. This is an unavoidable TOCTOU in the design, but the more actionable issue is: `installdecider` should capture `.z.m.ownerfunc[event]` (which was set by `registerdecider`) rather than `value event` to avoid this race.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+  .z.m.loginfo[`register;"registered decider ",string[nm]," on ",string event];
+  };
+
+removeobserver:{[event;nm]
+  / drop the named handler; the dispatcher and captured original keep firing regardless of remaining count
+  if[not event in key .z.m.registry;:(::)];
+  .z.m.registry[event]:.z.m.registry[event] _ nm;
+  .z.m.loginfo[`remove;"removed observer ",string[nm]," on ",string event];
+  };
+
```
DI-Software-Engineering, line 76: In `registerdecider`, when the event is already in `key .z.m.ownerfunc` (a dispatcher is installed), the reclaim path does `.z.m.ownerfunc[event]:func` and correctly skips `set[event;func]`. However when the event is NOT yet in `key .z.m.ownerfunc` (no observers yet), it calls `set[event;func]`. This means a reclaim before the first `observe` rebinds `.z.*` directly to the new `func`. But at this point the old `func` is still the live `.z.*` binding from the previous `register`. Replacing it with `set` is correct. The problem is that `installdecider` (called from `observe`) subsequently does `value event` to capture the owner — but it should instead read `.z.m.ownerfunc` if already populated, or store the owner in `.z.m.ownerfunc` at `register` time unconditionally, to avoid the indirect read. Recommended fix: in `registerdecider`, always write `.z.m.ownerfunc[event]:func` in addition to (or instead of) `set[event;func]`, so `installdecider` can always read from `.z.m.ownerfunc` rather than `value event`.

di/handlers/test_integration.csv
```diff
@@ -0,0 +1,37 @@
...
+before,0,0,q,qcands:$[count qh:getenv[`QHOME];(raze(qh;"/bin/q");raze(qh;"/";string .z.o;"/q"));()],1,1,candidate q binary locations for kdbx and classic layouts
+before,0,0,q,qbin:$[count ex:qcands where {not ()~key hsym `$x} each qcands;first ex;"q"],1,1,pick the first candidate that exists else fall back to q on PATH
+before,0,0,q,.it.tryopen:{[p] @[{hopen (`$"" sv (":localhost:";string x);500)};p;0Ni]},1,1,attempt one connection returning a null handle on failure
+before,0,0,q,.it.wait:{[p;n] $[n<=0;0Ni;not null hh:.it.tryopen p;hh;[system"sleep 0.3";.it.wait[p;n-1]]]},1,1,poll for the child to accept connections
+before,0,0,q,system"p 0W",1,1,ask the OS for a free ephemeral port on this process
+before,0,0,q,cport:system"p",1,1,read the OS-assigned port
+before,0,0,q,system"p 0",1,1,release the listener so the child can bind that port
+before,0,0,q,cpid:"J"$last system "" sv (qbin;" -q -p ";string cport;" </dev/null >/dev/null 2>&1 & echo $!"),1,1,launch a blank child q on the OS-assigned port and capture its PID
+before,0,0,q,h:.it.wait[cport;20],1,1,connect to the child with a bounded retry
```
DI-Software-Engineering, line 21: The integration test uses `system"p 0W"` to claim an ephemeral port, reads it with `system"p"`, then immediately releases it with `system"p 0"` before launching the child. There is a TOCTOU race: another process can bind that port in the window between `system"p 0"` and the child's `\p <cport>` startup. The test handles a failed connection by killing the child and exiting, so it skips rather than falsely failing, but the race can cause spurious skips in busy CI environments. A more robust approach is to let the child itself pick a free port (`\p 0`) and report it back, or to not release the listener until the child is launched.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+/ ------------------------------------------------------------
+/ decider-observer machinery (side-effect-only watchers on a decider event's owner)
+/ ------------------------------------------------------------
+
+rundeciderobs:{[event;nm;func;result;args]
+  / apply one decider observer under protection - side-effect only, never alters the result; a throw is logged at warn and swallowed
+  .[func;(result;args);{[event;nm;e] .z.m.logwarn[`handlers;"decider observer ",string[nm]," on ",(string event)," failed: ",e]}[event;nm]];
+  };
```
DI-Software-Engineering, line 89: `removedecider` uses `system"x ",string event` to restore the KDB-X built-in default. `string event` on a symbol like `` `.z.pg `` returns `".z.pg"` (with the leading dot). `\x .z.pg` is the correct form, so `string event` produces the right string. However `system` takes a string and runs a shell command — `\x` is a q system command, not a shell command. The correct call is `\x .z.pg` as a q statement, not via `system`. In q, `system"x .z.pg"` does work for `\x`-style commands (they are q interpreter directives accepted via `system`), so this is actually valid q. No change needed — withdrawing.

DI-Software-Engineering:
## DIReview Summary

**2 critical** | **4 warning(s)** | **0 suggestion(s)**

di/handlers/handlers.q
alowrydi: enlist arg passes arg as one positional argument - .[f;enlist arg;g] doesn't iterate over it. With no captured original, the value is (::), and (::)[arg] is just identity, discarded by the caller. No change needed.

di/handlers/handlers.q
alowrydi: Only triggers if something outside di.handlers rebinds the event before the first observe, which would already be a contract violation. The suggested fix doesn't work either, as this line is what creates .z.m.ownerfunc[event], so it can't read it first. Leaving as-is.

di/handlers/handlers.q
alowrydi: This would introduce a bug. `event in key .z.m.ownerfunc` is the "dispatcher installed" check and populating it at register time makes it permanently true, so a reclaim before any observe would skip rebinding the live event, silently breaking reclaim. Also removes the tested zero-overhead path for never-observed deciders. Not applying this.

di/handlers/test_integration.csv
alowrydi: Real race, already an accepted tradeoff. A lost race is a clean skip, not a false failure, since the retry draws a fresh port each attempt. The suggested alternative (don't release until the child launches) isn't possible, as two processes can't hold the same port simultaneously.

di/handlers/handlers.q
alowrydi: Valid - this wasn't actually documented. Adding to Notes in handlers.md stating that an observer-event dispatcher, once installed, stays installed permanently; remove clears its registry entry but the event never reverts to its pre-di.handlers binding, unlike a decider's \x restore. Doc-only change, no logic touched.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+  :$[event in observerevents; $[event in key .z.m.registry;0!.z.m.registry event;0!registryschema];
+     event in deciderevents;  $[event in key .z.m.owner;deciderlist event;([]role:`symbol$();name:`symbol$())];
+     [.z.m.logerr[`list;err:"di.handlers: unsupported event ",string event];'err]];
+  };
+
+observe:{[event;nm;func]
+  / attach a side-effect-only observer to a decider event's owner - it runs after a successful owner call and never alters the result
+  if[not -11h=type event;.z.m.logerr[`observe;err:"di.handlers: event must be a symbol"];'err];
+  if[not -11h=type nm;.z.m.logerr[`observe;err:"di.handlers: name must be a symbol"];'err];
+  if[not (type func) within 100 112h;.z.m.logerr[`observe;err:"di.handlers: func must be a function"];'err];
```
ConorSwainDI: Not breaking or functional but just an aesthetic suggestion. Suggest using `type[func]` instead of `(type func)` to move away from parentheses.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+  .z.m.registry:(`$())!();
+  .z.m.original:(`$())!();
+  .z.m.owner:(`$())!();
+  .z.m.ownerfunc:(`$())!();
+  .z.m.deciderobs:(`$())!();
+  };
+
+runobserver:{[event;nm;func;arg]
+  / apply one observer under protection - a throw is logged at warn and swallowed so the chain continues
+  .[func;enlist arg;{[event;nm;e] .z.m.logwarn[`handlers;"observer ",string[nm]," on ",(string event)," failed: ",e]}[event;nm]];
```
hehirs: either use function[arg] or (function arg) don't mix them please

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+  };
+
+runoriginal:{[event;arg]
+  / apply the captured pre-existing handler last, isolated the same way as registered observers
+  .[.z.m.original event;enlist arg;{[event;e] .z.m.logwarn[`handlers;"original ",(string event)," handler failed: ",e]}[event]];
```
hehirs: I'm 99% sure that `.[f;enlist arg;{err}]` is the same as `@[f;arg;{err}]` could you confirm and, if so, can you change the . to be @

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+    '"di.handlers: log dependency is required; pass `info`warn`error functions keyed on `log"];
+  if[99h<>type deps`log;
+    '"di.handlers: log value must be a dict; pass `info`warn`error functions"];
+  if[not all `info`warn`error in key deps`log;
+    '"di.handlers: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  .z.m.loginfo:(deps`log)`info;
+  .z.m.logwarn:(deps`log)`warn;
+  .z.m.logerr:(deps`log)`error;
+  / initialise the registries only on first init - a direct (module-rewritten) reference detects prior setup
+  if[not @[{.z.m.registry;1b};::;0b];initstate[]];
```
hehirs: Personally not sure if initstate needs to be a separate function. All it does right now is define the registries. I think those definitions could just go in this if statement. This keeps all our definitions inside the init function.

di/handlers/handlers.q
```diff
@@ -0,0 +1,216 @@
...
+  if[not event in key .z.m.ownerfunc;installdecider event];
+  .z.m.deciderobs[event]:.z.m.deciderobs[event] upsert (nm;func);
+  .z.m.loginfo[`observe;"attached observer ",string[nm]," to ",string event];
+  };
+
+unobserve:{[event;nm]
+  / detach an observer from a decider event - the dispatcher stays installed even after the last observer is removed
+  if[not -11h=type event;.z.m.logerr[`unobserve;err:"di.handlers: event must be a symbol"];'err];
+  if[not -11h=type nm;.z.m.logerr[`unobserve;err:"di.handlers: name must be a symbol"];'err];
+  if[not event in key .z.m.deciderobs;:(::)];
```
hehirs: should there be some logging here?

di/handlers/handlers.q
```diff
@@ -0,0 +1,213 @@
...
+registerobserver:{[event;nm;func]
+  / lazily install on first use, then add or replace this handler in registration order
+  if[not event in key .z.m.original;installobserver event];
+  .z.m.registry[event]:.z.m.registry[event] upsert (nm;func);
+  .z.m.loginfo[`register;"registered observer ",string[nm]," on ",string[event]];
+  };
+
+registerdecider:{[event;nm;func]
+  / single owner - a different name claiming an owned event is rejected; the same name reclaims (idempotent re-init)
+  if[event in key .z.m.owner;
```
DI-Software-Engineering: The `removeobserver` function logs 'no observers registered on event' when the event key is missing, but does not check whether the specific name `nm` is actually present in `.z.m.registry[event]`. A caller removing a name that was never attached (but the event does have other observers) will silently succeed and log 'removed observer X on event' even though nothing was removed. After the early-exit guard, add a check: `if[not nm in exec name from .z.m.registry[event];.z.m.loginfo[...];:(::)]` before the drop.

DI-Software-Engineering:
## DIReview Summary

di/handlers/handlers.q
alowrydi: Square brackets style adopted throughout.

di/handlers/handlers.q
alowrydi: Square brackets style adopted throughout.

di/handlers/handlers.q
alowrydi: Agreed, change made.

di/handlers/handlers.q
alowrydi: initstate now incorporated into init.

di/handlers/handlers.q
alowrydi: no-op guard when removing/unobserving a name that is not registered

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  $[phase~`pre;.z.m.pre[event]:upsertphase[.z.m.pre event;nm;pri;func];
+    .z.m.post[event]:upsertphase[.z.m.post event;nm;pri;func]];
+  .z.m.loginfo[`register;"registered ",string[phase]," handler ",string[nm]," on ",string[event]];
+  };
+
+removeprepost:{[event;phase;nm]
+  / drop a pre or post handler; the dispatcher stays installed regardless of remaining count
+  tbl:$[phase~`pre;.z.m.pre;.z.m.post];
+  if[not event in key tbl;.z.m.loginfo[`remove;"no ",string[phase]," handlers on ",string[event],"; nothing to remove for ",string[nm]];:(::)];
+  if[not nm in exec name from tbl event;.z.m.loginfo[`remove;string[phase]," handler ",string[nm]," not registered on ",string[event],"; nothing to remove"];:(::)];
```
DI-Software-Engineering, line 148: `registerprepost` rejects pre/post registration when `event in key .z.m.ownerfunc` is false, but `registerexec` writes to `.z.m.ownerfunc[event]` *before* calling `installphased`, which creates the `.z.m.pre` / `.z.m.post` entries. After the first `registerexec` call, `event in key .z.m.ownerfunc` is true, so the guard works. However, if `registerexec` is called twice for the same event (same-name reclaim) the second call overwrites `.z.m.ownerfunc` again and `installphased` is skipped — that is correct. The real problem: the guard `if[not event in key .z.m.ownerfunc` checks `.z.m.ownerfunc`, which is populated by `registerexec` at line ~143 *before* the `installphased` at line ~145 creates `.z.m.pre`. So the guard condition and the state-creation are in the right order here. **However**, `removeexec` deletes from `.z.m.ownerfunc` (line ~175) and also deletes from `.z.m.pre`/`.z.m.post`. A subsequent `registerexec` for the same event will write to `.z.m.ownerfunc` first and *then* call `installphased` because `event` is no longer in `key .z.m.pre`. But between those two lines, if `registerprepost` were called concurrently the guard would pass with `.z.m.pre` not yet initialised. This is a q single-threaded concern so not a real race — but the more concrete bug is: `installphased` is guarded by `if[not event in key .z.m.pre;installphased event]`, yet `.z.m.ownerfunc[event]:func` is written unconditionally before that guard. On a same-name reclaim where the event IS already in `key .z.m.pre`, `installphased` is skipped (correct), but `set[event;…]` is never called again — the dispatcher already installed still reads `.z.m.ownerfunc` fresh so the reclaim works. This is fine. The actual bug: `registerprepost` checks `not event in key .z.m.ownerfunc`, but `removeexec` deletes the event from `.z.m.ownerfunc` **and** `.z.m.pre`/`.z.m.post`. After removal, both keys are absent. A subsequent `registerexec` writes `.z.m.ownerfunc[event]` first, then (seeing the event absent from `.z.m.pre`) calls `installphased`, which initialises `.z.m.pre` and `.z.m.post` and calls `set`. This path is correct. No bug here — removing finding.

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  runpostall[event;r;enlist req];
+  r
+  };
+
+dispatchpw:{[u;p]
+  / .z.pw is the only binary phased event - the owner sees the real password; post handlers see it redacted to "***"; no pre
+  r:.z.m.ownerfunc[`.z.pw][u;p];
+  runpostall[`.z.pw;r;(u;"***")];
+  r
+  };
```
DI-Software-Engineering, line 111: `dispatchphased` passes `enlist req` as the `args` argument to `runpostall`, where `req` is the *transformed* request after running all `pre` handlers. The `post` handler signature is `func[result;args]` where `args` is documented as the call arguments. For a unary phased event the original caller's argument is `req` before the pre-chain runs, but `post` receives the *post-pre* request. If any `pre` handler reshapes the request, `post` sees the modified version, not what the original caller sent. If `args` is meant to be the original caller's input, capture it before the pre-fold: `orig:req; req:{[acc;f] f acc}/[req;pres]; … runpostall[event;r;enlist orig]`. If `args` is intentionally the transformed request that `exec` received, the documentation should say so explicitly — but the `.z.pw` branch in `dispatchpw` passes the original `(u;p)` pair (line 120), making the two dispatchers inconsistent about what `args` means.

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  / relinquish ownership only if this name owns it, restore the KDB-X default, and clear all phased state for the event
+  if[not event in key .z.m.owner;.z.m.loginfo[`remove;"no exec owner on ",string[event],"; nothing to remove for ",string[nm]];:(::)];
+  if[not nm~.z.m.owner event;
+    .z.m.logerr[`remove;err:"di.handlers: ",string[event]," is owned by ",string[.z.m.owner event],", not ",string[nm]];'err];
+  / .z.ph ships a default handler that \x expunges rather than restores, so put the captured original back; every other phased event \x-restores its built-in default
+  $[(event in key .z.m.original) and not (::)~.z.m.original event;set[event;.z.m.original event];system"x ",string[event]];
+  .z.m.owner:.z.m.owner _ event;
+  .z.m.ownerfunc:.z.m.ownerfunc _ event;
+  .z.m.ownerpri:.z.m.ownerpri _ event;
+  .z.m.original:.z.m.original _ event;
```
DI-Software-Engineering, line 164: `removeprepost` reads the target table via `tbl:$[phase~`pre;.z.m.pre;.z.m.post]` and then checks `if[not event in key tbl;…]`. In q, `$[cond;dict1;dict2]` returns the value of the selected dictionary at evaluation time — `tbl` holds a snapshot of the reference, not a live alias. The subsequent `delete` lines correctly use `.z.m.pre[event]` and `.z.m.post[event]` directly, so the mutation is fine. The check `event in key tbl` is also fine because it reads `tbl` which was assigned the dictionary value. However, `if[not nm in exec name from tbl event;…]` — here `tbl event` selects from the snapshot dictionary, which is correct. No mutation bug. **The real issue**: the `$[phase~`pre;.z.m.pre;.z.m.post]` idiom is used for both the existence check and the name check, but the subsequent mutation uses separate `$[…]` branches. If `phase` is neither `` `pre `` nor `` `post `` (it cannot reach here due to upstream validation in `remove`, but `removeprepost` is also callable internally), the mutation branch silently falls to `.z.m.post`. This is a latent issue rather than a confirmed bug given current call sites — omitting.

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  .z.m.pre:.z.m.pre _ event;
+  .z.m.post:.z.m.post _ event;
+  .z.m.loginfo[`remove;"removed exec owner ",string[nm]," on ",string[event],"; restored default"];
+  };
+
+phasedlist:{[event]
+  / uniform ([]phase;name;priority) view - pre rows (priority order), the exec owner, then post rows (priority order)
```
DI-Software-Engineering, line 171: `removeexec` restores the original handler for `.z.ph` with: `$[(event in key .z.m.original) and not (::)~.z.m.original event; set[event;.z.m.original event]; system"x ",string[event]]`. The `(::)` sentinel was stored in `installphased` via `@[value;event;{(::)}]` — meaning `.z.ph` had no handler at capture time. In that case `(::)~.z.m.original event` is true, and the code falls through to `system"x "…`, which expunges the handler. But the note in both the code and docs says "`\x` expunges rather than restores" `.z.ph`'s shipped default — the entire point of capturing it was to avoid calling `\x`. If `@[value;event;{(::)}]` returned `(::)` it means the shipped default was somehow absent at `installphased` time (unusual), so falling back to `\x` is the only option. This case is correct. **The actual bug**: `installphased` is only called from `registerexec`. `.z.m.original[event]` is populated inside `installphased` only for `.z.ph` (`if[event~`.z.ph;…]`). If the `exec` is removed and re-registered, `installphased` runs again (because `.z.m.pre` was deleted by `removeexec`). At that point `.z.ph` is already bound to the captured original (restored by the first `removeexec`), so the second capture correctly re-captures it. That is fine. However, `removeexec` also does `.z.m.original:.z.m.original _ event` — it clears the captured original. So on re-register after remove, `installphased` captures the (correctly restored) original again. This is correct. No bug here — removing finding.

di/handlers/handlers.q
DI-Software-Engineering, line 111: In `dispatchphased`, the pre-fold `req:{[acc;f] f acc}/[req;pres]` runs every `pre` function unprotected, which is documented and intentional for veto. However, if `pres` is an empty list (no pre handlers registered), q's `/` over an empty list returns the seed unchanged — correct. But `pres` is `exec func from …` which returns a list of functions. When it is a single-element list, `{[acc;f] f acc}/[req;pres]` calls the function once — correct. When it is empty, the seed `req` is returned — correct. No bug.

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  / thread the request through pre (transform, unprotected) -> exec owner (unprotected) -> post (observe, protected)
+  pres:exec func from `priority xasc .z.m.pre event;
+  req:{[acc;f] f acc}/[req;pres];
+  r:.z.m.ownerfunc[event] req;
+  runpostall[event;r;enlist req];
+  r
+  };
+
+dispatchpw:{[u;p]
+  / .z.pw is the only binary phased event - the owner sees the real password; post handlers see it redacted to "***"; no pre
```
DI-Software-Engineering, line 107: `dispatchphased` extracts pre functions as `pres:exec func from …` and folds with `{[acc;f] f acc}/[req;pres]`. The `exec` projection on a column returns a plain list. When the table has exactly one row, `exec func from t` returns a one-element list (not an atom), so the fold works correctly. No bug.

di/handlers/test.csv
```diff
@@ -0,0 +1,189 @@
...
+run,0,0,q,handlers.register[`.z.ph;`post;`log;0;{[r;a] .oo.phran::1b}],1,1,a post watcher on .z.ph
+run,0,0,q,.oo.phran:0b,1,1,reset the .z.ph post flag
+run,0,0,q,.oo.phres:(value `.z.ph)"GET /",1,1,drive the .z.ph dispatcher
+true,0,0,q,.oo.phres~`handled,1,1,the exec owner produced the result
+true,0,0,q,.oo.phran,1,1,the post watcher fired on .z.ph
+true,0,0,q,`web~exec first name from select from handlers.list[`.z.ph] where phase=`exec,1,1,list names the .z.ph exec owner
+run,0,0,q,handlers.remove[`.z.ph;`exec;`web],1,1,remove the .z.ph owner
+true,0,0,q,.oo.phdef~value `.z.ph,1,1,remove restored the exact shipped GET handler not \\x-to-unset
+true,0,0,q,100h=type value `.z.ph,1,1,and .z.ph is a real handler function again not left unset
+
```
DI-Software-Engineering, line 113: The test `true,0,0,q,.oo.res~12` asserts that `pre` ran `add1` (priority 10) then `mul2` (priority 20) on request `5`, giving `(5+1)*2 = 12`. But `dispatchphased` folds left-to-right through `pres` sorted ascending by priority, so priority 10 runs first: `add1(5)=6`, then `mul2(6)=12`. The test assertion `~12` is correct for this ordering. However the comment says "pre ran add1 then mul2 in priority order giving 12 not 11" — if mul2 ran first it would give `2*5=10`, then `add1(10)=11`. The assertion correctly distinguishes these, so the test is a valid correctness check. No bug in the test logic.

di/handlers/handlers.q
```diff
@@ -0,0 +1,248 @@
...
+  r:.z.m.ownerfunc[`.z.pw][u;p];
```
DI-Software-Engineering, line 108: `dispatchphased` calls `runpostall[event;r;enlist req]` where `req` is the transformed request (after pre-chain). For multi-argument phased events there is no issue since `.z.pg`, `.z.ps`, `.z.pi`, `.z.pp`, `.z.ph`, `.z.ws` are all unary. But the `args` passed to `post` is `enlist req` — a one-element list wrapping the (possibly pre-transformed) request. By contrast, `dispatchpw` passes `(u;"***")` — a two-element list with the original arguments. This inconsistency means `post` handlers for unary events receive `enlist transformedRequest` while `.z.pw` post handlers receive the raw original args. If `post` handlers are written expecting consistent `args` semantics (original caller args, not transformed), this is a silent contract violation. Fix: either document that `args` in `post` is the transformed request for phased events (and update `.z.pw` to pass `enlist u` rather than a two-element tuple), or capture the original request before the pre-fold and pass that. Currently the docs say `.z.pw` post sees `(user;"***")` while other `post` handlers see `enlist transformedRequest` — a real asymmetry that will surprise callers.

DI-Software-Engineering:
## DIReview Summary

**4 critical** | **4 warning(s)** | **0 suggestion(s)**

di/handlers/handlers.md
```diff
@@ -0,0 +1,182 @@
...
+
+- **Simple event** (`phase` `` ` ``): `func` is added to the fan-out. Re-registering the same `[event;name]` replaces that entry in place.
+- **`exec`**: `func` becomes the single owner and produces the outcome. Re-registering under the same `name` replaces it; a different `name` on an already-owned event throws. Registering the `exec` is what installs the event's dispatcher.
+- **`pre` / `post`**: added to the pre- or post-chain (priority order). Both are rejected if the event has no `exec` owner yet. A `pre` receives the request and returns the (possibly reshaped) request; a `post` is called `post[result;args]` and its return is discarded. `args` is the argument list `exec` was invoked with, **with one deliberate exception — `.z.pw`'s password is redacted**. For the six unary phased events `args` is `enlist request`, exactly what `exec` ran on: the request *after* any `pre` reshaped it, not the original caller's input. For the binary `.z.pw`, `exec` sees the real `(user;password)` but `post` sees `(user;"***")`, so a `post` handler can never read a real password. `result` is what that request produced.
```
DI-Software-Engineering, line 81: The inline description added to the bullet point on line 81 duplicates — and partially contradicts — the dedicated bullet added at line 175. The inline text says `post[result;args]` while the dedicated bullet says `args` is paired 'with the `result` that request produced', implying the signature is `post[result;args]`. These are consistent in signature but the two descriptions of the `.z.pw` redaction and `enlist request` semantics are now repeated verbatim in two places. More critically, the inline addition on line 81 states 'For the six unary phased events `args` is `enlist request`' — hardcoding the count 'six' — while the dedicated bullet at line 175 omits that count. If the number of supported unary events ever changes, one description will become stale and mislead readers. Remove the inline addition from line 81 and rely solely on the dedicated bullet at line 175, or remove the count from the inline text.

DI-Software-Engineering:
## DIReview Summary

di/handlers/handlers.q
alowrydi: Same finding as the args comment above, see that reply.

di/handlers/handlers.q
alowrydi: args is the argument list exec was invoked with, with one deliberate exception — .z.pw's password is redacted. For the six unary phased events, args is enlist request, exactly what exec ran on (post-pre, not the original caller input). For .z.pw, exec sees the real (user;password) but post sees (user;"***") — that's intentional, it's what lets an audit-log post handler exist at all without ever touching a real password. Not taking either suggested fix: enlist u would delete the redaction and break a passing test; passing the pre-fold original would pair a result with a request that didn't produce it.

di/handlers/handlers.md
alowrydi: This is stylistic redundancy rather than an inaccuracy. Resolving as-is.

jprgrant (approved):

PR #112, opened 2026-06-30
Feature di.kafka

alowrydi:
<html>
<body>
<p class="font-claude-response-body break-words whitespace-normal"><strong>Kafka producer/consumer module, di.kafka</strong></p>
<p class="font-claude-response-body break-words whitespace-normal">Extracts <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">code/common/kafka.q</code> from TorQ and packages it as a standalone kdb-x module. Wraps the <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">kafkaq</code> native shared library, providing consumer and producer lifecycle management and a configurable message callback for kdb-x processes that need to produce or consume Kafka messages.</p>
<p class="font-claude-response-body break-words whitespace-normal">Trello ticket - <a class="underline underline underline-offset-2 decoration-1 decoration-current/40 hover:decoration-current focus:decoration-current" href="https://trello.com/c/eDFHYhSx/103-kdb-x-kafka-module">https://trello.com/c/eDFHYhSx/103-kdb-x-kafka-module</a></p>
<hr class="border-border-200 border-t-0.5 my-3 mx-1.5">
<p class="font-claude-response-body break-words whitespace-normal"><strong>Files created</strong></p>
<div class="overflow-x-auto w-full px-2 mb-6">
**Files created**

| File | Description |
| -- | -- |
| `di/kafka/init.q` | Loads `kafka.q`, defines export of 8 functions |
| `di/kafka/kafka.q` | Full implementation - module state, native function stubs/forwarders, internal helpers, public API, init |
| `di/kafka/kafka.md` | Full module documentation - see below |
| `di/kafka/test.csv` | 34 k4unit unit tests across 9 test groups, runs on any kdb-x machine |
| `di/kafka/test_integration.csv` | 7 k4unit integration tests verifying C library bindings, requires `KDBLIB` |

</div>
<hr class="border-border-200 border-t-0.5 my-3 mx-1.5">
<p class="font-claude-response-body break-words whitespace-normal"><strong>How to test</strong></p>
<p class="font-claude-response-body break-words whitespace-normal">Unit tests:</p>
[...]

PR #117, opened 2026-08-03
Feature depcheck

alowrydi:
**Dependency, version, core-contract, and `.z.ts`-ownership auditing module, di.depcheck**

Extracts and modernises legacy TorQ's `checkdependency`/`runchk`/`checkvers` as a standalone kdb-x module. Audits every already-loaded module's declared dependencies for presence and minimum version, validates the shared core dependency contracts (Logging/Timer/Handlers) the DI pattern relies on, and flags a `.z.ts` binding outside `di.timer`. A per-module manifest replaces the old CSV registry - readable as either `deps.q` (a q dict) or `deps.toml` (a `[dependencies]` table, via `di.toml` - PR #116), merged with `.toml` winning on a clash - and real numeric semver comparison replaces the 5-component digit-walk.

Trello ticket - **https://trello.com/c/PIzyEuoH/125-kdb-x-depcheck-module**

---

**Files created**

| File | Description |
| -- | -- |
| `di/depcheck/init.q` | Loads `depcheck.q`, defines export of 3 functions |
| `di/depcheck/depcheck.q` | Full implementation - session introspection, manifest reading (dual-format), semver comparison, transitive graph walk, core-contract checks, `.z.ts` check, kdb-x version check, public API |
| `di/depcheck/depcheck.md` | Full module documentation - see below |
| `di/depcheck/deps.q` | Empty - di.depcheck has no hard dependencies, dogfooding the convention it introduces |
| `di/depcheck/test.csv` | k4unit unit tests - 96 |
| `di/depcheck/test_integration.csv` | k4unit integration tests against a real spawned child q process - 5 |

---

**Unit tests**

```q
k4unit:use`di.k4unit
k4unit.moduletest`di.depcheck

2026.08.03T12:40:08.215 start
2026.08.03T12:40:08.215 :/home/alowry/bin/local-kdbx-modules/di/depcheck/test.csv 96 test(s)
2026.08.03T12:40:08.291 end
All tests passed

96/96 passing. 
```
(A `tried the following paths... di/log/init.li64.so` message is expected, harmless console output from the `di.log`-absent graceful-skip path being genuinely exercised - `di.log` isn't merged yet - not a failure.)

A number of assertions are conditional on other unmerged PRs happening to be present on QPATH (`di.kafka` #112, `di.handlers` #114, `di.log` #90, `di.toml` #116) - each gracefully no-ops rather than fails when the module it depends on isn't resolvable, so the suite passes standalone against a bare `main` checkout and exercises more real paths the more of those branches happen to be merged locally alongside it.

Coverage includes: `init` dependency validation (all four failure modes), semver comparison and malformed-version handling (a pre-release-tagged found version, a typo'd declared minimum), the kdb-x engine version check, live checks against real shipped/candidate modules rather than only synthetic fixtures (`di.timer` is the permanent live negative control - always loaded, always missing `version` and failing its own Timer contract), a fabricated loaded-but-unreadable-export edge case (regression test for an eager-`or` short-circuit bug caught by manual smoke testing, not by reading the code), non-`di.*` vendor dependency resolution, `checkmoduledeps` against a real on-disk `deps.q` (di.depcheck's own) proving the QPATH file-discovery path actually works, `checkcontract`/`checkcontracts` (including the bare-symbol-atom normalisation on the newly-exported `checkcontract`), `.z.ts` ownership, an end-to-end `init` regression test proving a warning is still logged when it coincides with a failure in the same call (a real bug: warnings used to be silently dropped when the failure branch's signal fired first), the dual-format manifest merge (`deps.q`-only, `deps.toml`-only, both with a clash - `.toml` wins - all via scratch QPATH fixtures created and torn down at runtime, no committed fixtures), a broken `deps.toml` degrading to one aggregated failure line rather than throwing, a `deps.toml` with no `[dependencies]` section, a malformed `[dependencies]` section (a scalar instead of a table), non-string manifest version values from either format (including confirming a legitimate single-character string version is *not* false-positived), and the transitive graph walk itself (an installed-but-unloaded dependency reached at depth ≥ 2, a cycle, and confirmation that a direct dependency of a loaded module is never double-reported between `checkdeps` and `checkgraph`).

---

**Integration tests**

Spawns a real, separate child kdb-x process via its own binary, resolved from `/proc/<pid>/exe` rather than `QHOME` - no configuration required, skips cleanly if no usable binary is found:
```q
k4unit:use`di.k4unit
.m.di.0k4unit.KUltf .Q.dd[hsym`$.Q.m.mp`di.depcheck;`test_integration.csv]
.m.di.0k4unit.KUrt[]
k4unit.getresults[]

2026.08.03T12:40:36.694 start
2026.08.03T12:40:36.694 :/home/alowry/bin/local-kdbx-modules/di/depcheck/test_integration.csv 5 test(s)
2026.08.03T12:40:36.761 end

5/5 passing.
```
Confirms two things the unit suite structurally cannot: a genuine zero-failure success path completing cleanly end to end in a real process (unreachable in `test.csv`, where `di.timer` always fails its own contract), and a real signalled failure actually terminating a real host process with a non-zero exit code and the real error report in captured output.

---

**Design decisions**

*1. Post-load audit, not a pre-load gate* - `di.depcheck` never calls `use` on anything it checks; every check reads another module's state purely by introspecting the session namespace the kdb-x `use` loader already populates (`.m.di.0<shortname>`, each module's `export` dict readable directly underneath). It runs once, after `di.torq` has already loaded every module the process needs. "Not found" in a report means a *declared* dependency was never loaded into this session - not a QPATH filesystem scan.

*2. Dual-format manifest reading, `.toml` winning on a clash* - A module's dependencies are read from both `deps.q` and `deps.toml`, wherever each exists, merged the same way `di.config`'s real `parsetier` does (`(parsefile base,".q"),parsefile base,".toml"` - a right-biased dict union). This lets both formats coexist through a migration rather than forcing an all-or-nothing switch.

*3. `di.toml` is a lazy, file-existence-gated dependency - not a hard one* - `di.toml` is resolved and loaded only once a module being audited is confirmed to actually ship a `deps.toml` file; a module with only `deps.q` never touches it. This mirrors `di.config`'s own `requiretoml` guard exactly in wording and gating - name the file, name the underlying cause, one sentence - but deliberately diverges from it in one respect: `di.config`'s `requiretoml` throws and aborts its entire settings cascade the instant a `.toml` tier can't be read, because it must hand back one complete, correct config. `di.depcheck` catches the equivalent failure, folds one line into its aggregate report, and keeps walking - its whole purpose is surfacing every problem across many modules in one pass, not producing a single value that has to be complete.

*4. Transitive on-disk manifest graph walk* - Beyond each loaded module's own direct dependencies, `checkgraph` walks the manifest graph on disk from every loaded module - reading each peer's manifest whether or not that peer is itself loaded, cycle-guarded via a visited set - and reports a presence failure for anything reached at depth ≥ 2 that resolves nowhere on QPATH. It loads no module code. Direct (depth-1) dependencies keep the stronger existing presence=*loaded*/version=*exported* check and are excluded from the walk's own reporting, so the two never double-report the same gap.

*5. Core dependency contracts - a fixed, named set* - For whichever of `di.log`, `di.timer`, or `di.handlers` happen to be loaded, the export dict is checked against the required keys of the contract it provides. Not a generic self-declaration registry, since no such mechanism exists anywhere in this codebase yet. The single-contract check underneath is also exported directly as `checkcontract[provider;requiredkeys]`, for auditing a contract this module doesn't know by name.

*6. `.z.ts` ownership is a best-effort proxy, not a rebind detector* - `di.timer`'s `enabled` flag is evidence its `init` ran and bound `.z.ts` itself; it cannot prove nothing has rebound `.z.ts` since. Warning-level only, and deliberately scoped to `.z.ts` alone - whether this should ever cover other `.z.*` events is an open question for `di.handlers`' owner.

*7. Malformed input is reported explicitly, never silently mis-compared* - A manifest version that isn't a string (an unquoted `deps.toml` float, a `deps.q` int or symbol) gets its own clear line rather than corrupting a concatenated message or throwing out of semver parsing. A version string that doesn't parse as a clean `major.minor.patch` triple (a pre-release tag, a typo) collapses to an explicit malformed sentinel rather than letting a single bad component silently null itself out of the comparison one piece at a time - which previously could make a genuinely newer version compare as older, or a typo'd minimum compare as satisfied by anything. Both caught by direct/adversarial testing, not by reading the code.

---

**Checklist**
- [ ] 96/96 unit tests passing
- [ ] 5/5 integration tests passing
- [ ] `depcheck.md` documents all exported functions, config, usage examples and notes
- [ ] `di.toml` (PR #116) - soft, lazy dependency only; not required to merge first, relied on only when an audited module ships a `deps.toml`

---

**Documentation**

See `depcheck.md` for full reference including dependency table, configuration options, exported function documentation with examples, usage example, and notes - including the deliberate divergence from `di.config`'s abort-on-broken-TOML behaviour, and the current limitation that transitive *version* checking (as opposed to presence) is deferred until per-module `VERSION` files are a repo-wide convention.


di/depcheck/init.q
```diff
@@ -0,0 +1,9 @@
+/ di.depcheck - dependency presence/version, core-contract, and .z.ts ownership auditing for kdb-x modules
+/ intended to run once, post-load, after a host process (di.torq) has `use`d every module it needs - see depcheck.md
+
+\l ::depcheck.q
+
+/ Module version
```
DI-Software-Engineering, line 6: `read0` on a VERSION file with no trailing newline (the file ends without one, as noted by 'No newline at end of file' in the diff) will return a single-element list containing the version string with no trailing whitespace issues, so `first` is correct there. However, the backtick-colon path `` `:::VERSION `` resolves relative to the current working directory at load time, not relative to the module file. If the process is started from a directory other than `di/depcheck/`, the read will fail or silently pick up a wrong file. Use a module-local path (`` `:::VERSION `` should be `` ::local/VERSION `` or the equivalent CWD-independent form consistent with the repo's path convention) to guarantee the file is always resolved next to `init.q`.

DI-Software-Engineering:
## DIReview Summary

jprgrant (approved):

PR #118, opened 2026-08-06
pass through date param in callendofday

jprgrant:


di/pubsub/pubsub.q
```diff
@@ -75,10 +75,10 @@ closesub:{[h]
 .z.pc:{closesub[x]};
 
 / broadcast to all subscribers upon end of day, client needs to define endofday function
-callendofday:{(neg getallhandles[])@\:`endofday`};
+callendofday:{[d](neg getallhandles[])@\:(`endofday;d)};
 
 / broadcast to all subscribers upon end of period, client needs to define endofperiod function
-callendofperiod:{(neg getallhandles[])@\:`endofperiod`};
+callendofperiod:{(neg getallhandles[])@\:(`endofperiod;x)};
```
DI-Software-Engineering, line 82: `callendofperiod` uses the implicit parameter `x` but the function was not given an explicit parameter list (unlike `callendofday` which correctly declares `[d]`). Without an explicit parameter, `x` is the implicit first argument — but `callendofperiod` is defined as a no-argument lambda `{…}` in the original, so `x` will be the generic null `(::)` when the function is called with no argument, silently broadcasting a wrong value. Fix by adding an explicit parameter: `callendofperiod:{[d](neg getallhandles[])@\:(`endofperiod;d)};`.

DI-Software-Engineering:
## DIReview Summary

jonathonmcmurray (approved):

di/pubsub/pubsub.q
```diff
@@ -75,10 +75,10 @@ closesub:{[h]
...
```
ascottDI, line 81: For the previous function we defined the variable, whereas, for this one we used the implicit. Would it be better to use the same format for both?

PR #120, opened 2026-08-06
Feature server

ascottDI:
# di.servers — TorQ Modularisation PR

## Summary

Extracts TorQ's `.servers` connection management (`code/handlers/trackservers.q` + `servers.q`) into a standalone kdb-x module: `di.servers`. The module dials a static `process.csv` "phone book", maintains a pool of live handles to peer processes, hands them out by proctype via a selection algorithm, and recovers dropped connections. It satisfies the di.* module contract: one-arg `init[deps]` dependency injection, strict validation with no fallback, a conservative exported API, `getapimeta` for `di.api`, and no hard `di.*` dependencies (`log`, `timer` and `handlers` are all injected).

---

## Background

TorQ's `.servers` tracks connected backend processes and lets a gateway (or any process) look up a handle to a peer by type. In TorQ this was entangled with the discovery service, password/access-list files, non-TorQ process tracking, environment reads, and the global process framework.

This PR is part of the broader TorQ → kdb-x modularisation effort. `di.servers` is a **FRAMEWORK-tier** module scoped down for **v1**: no discovery service, no password/access-list files, no non-TorQ process tracking, no FinSpace, and no environment reads. `process.csv` is treated as a static **phone book** (who to *dial*), **not** an identity source — a process's self-identity arrives via config, injected by `di.torq`. This makes connection management independently loadable, testable, and drivable from resolved config rather than TorQ globals.

---

### New files

| File | Description |
|---|---|
| `di/servers/servers.q` | Core implementation — `init`, `startup`, `getservers`, `gethandlebytype`, `waitfortype`, `getapimeta`, plus internal helpers (`retry`, `cleanup`, `opencon`, `formathp`, `readprocesscsv`, `retryrows`, `selector`, `updatestats`, `signalfound`, `raiseerror`) |
| `di/servers/init.q` | Module entry point — loads `servers.q` and declares the export list |
| `di/servers/test.csv` | k4unit test manifest (10 areas; value + error-path asserts + live-peer integration) |
| `di/servers/test.q` | Test helper (spins up / drives real peer processes for the live-handle tests) |
| `di/servers/servers.md` | Module README |

| Aspect | TorQ `.servers` | `di.servers` |
|---|---|---|
| Logging | Hard-coded `.lg.o` / `.lg.e` calls | Injected `log` dependency, three flat vars (`.z.m.loginfo`/`logwarn`/`logerr`), binary `{[c;m]}` |
| Handler registration | `.z.pc` via `.dotz.set` | Injected `handlers` dep; `.z.pc` registered as a **simple/observer** event via `register[event;phase;nm;pri;func]` (fan-out, side-effect only) |
| Retry timer | `.timer` framework | Injected `timer` dep; `` timer[`addjob][`custom] `` 6-arg variant, a **10s** (mode-1, seconds) `serversretry` job |
| Server state | Global `.servers.SERVERS` table | `.z.m.SERVERS` module-local mutable state |
| Self-identity | Derived from `process.csv` / discovery | From config (`proctype`/`procname` in `deps`); `process.csv` is a dial-only phone book |
| Environment | Reads env for paths | Reads **no** env; `process.csv` path arrives via `config`processcsv` (resolved by di.torq) |
| `process.csv` parsing | Full/variable TorQ layout | Strict v1 `host,port,proctype,procname` 4-column layout; header validated and **rejected loudly** (positional read would otherwise misparse silently) |
| Dependency wait | `startupdepcycles` | `waitfortype[proctype;timeoutms;pollms]` — "fail fast, but wait for a hard dependency"; caller decides if timeout is fatal |
| Scope | Discovery, access-lists, non-TorQ tracking, FinSpace | All removed for v1 (no unexercised branches shipped) |
| Module contract | None | kdb-x `use` singleton, strict `init[deps]` (no fallback), `raiseerror` log-then-signal, `export:` list, `getapimeta` |

---

## Exported API

```q
svc:use`di.servers

svc.init[deps]                        / wire injectables + config; required before all other calls. idempotent
svc.startup[]                         / open connections to configured proctypes from process.csv (reads init config)
svc.getservers[proctype]              / live (non-null handle) SERVERS rows for a proctype
svc.gethandlebytype[proctype;sel]     / one live handle via `any/`roundrobin/`last selection; 0Ni if none
svc.waitfortype[proctype;tmoms;polms] / block until a proctype connects or timeout elapses; 1b/0b
svc.getapimeta[]                      / api metadata rows for di.torq to register with di.api
```

`init`/`getapimeta` are exported as **plumbing** (di.torq calls them by convention) but are deliberately **not** listed in `getapimeta[]` — the registry describes the callable API, not plumbing.

---

## `deps` (injectables + config, assembled by di.torq)

`init` takes a single `deps` dict carrying both the injectable dependencies **and** this process's resolved config slice (the one-arg convention shared with `di.config`):

| key | kind | meaning |
|---|---|---|
| `log` | injectable (required) | binary `` `info`warn`error `` `{[c;m]}` logger dict — `di.log` satisfies it directly |
| `timer` | injectable (required) | `di.timer` export; calls `` timer[`addjob][`custom] `` (6-arg variant) |
| `handlers` | injectable (required) | `di.handlers` contract; `register[event;phase;nm;pri;func]` |
| `proctype` / `procname` | config (required) | this process's own identity; used to exclude self from `process.csv` |
| `connections` | config (optional) | proctypes this process should dial (symbols, or strings from a `.toml` cascade — normalised). Default: none |
| `processcsv` | config (optional) | **path** to `process.csv`. Required only once `connections` is non-empty |

`init` wires the deps, records self-identity, and installs two **one-time** process-global side effects — the `.z.pc` cleanup observer and the 10s retry job — guarded by an internal `registered` flag so `init` is **idempotent** (a duplicate `di.timer.addjob` id would otherwise throw). `init` does **not** open connections; that is `startup`'s job.

---

## process.csv format (v1)

Strict 4-column layout — header validated up front and rejected loudly if reordered or wider:

```
host,port,proctype,procname
localhost,5010,tickerplant,tp1
localhost,5011,rdb,rdb1
localhost,5012,hdb,hdb1
```

`startup` reads this phone book, drops this process's own row (matched on `proctype`+`procname`), connects to every row whose proctype is in `connections`, and records each as a `SERVERS` row. A failed connect is **logged, not raised**, and left as `w:0Ni` for `retry` to reopen. `startup` is idempotent — it skips procs already tracked, so a repeat call (or a grown `process.csv`) adds only new rows, never a duplicate or a leaked second handle.

---

## Connection recovery

- **Clean drops** — the injected `.z.pc` observer marks the closed handle's row disconnected (`w:0Ni`, `endp` stamped).
- **Ungraceful drops** — the scheduled `serversretry` job (10s) runs `cleanup` to sweep handles that vanished from `key .z.W` without a clean `.z.pc`, then reopens every dead (null) handle via `retry`.
- `waitfortype` reuses `retry` to block at startup until a hard-dependency proctype comes up (or a timeout elapses).

---

## Test coverage

Tests are in `test.csv` and run via `k4unit` (the live-handle tests spin up real peer processes via `test.q`):

```q
k4unit.moduletest`di.servers
```

### Areas covered

| Area | Coverage |
|---|---|
| `init` — dependency + config validation | Rejects non-dict deps, missing `log`/`timer`/`handlers`, non-dict / partial `log`, bad `timer` shape (missing `addjob` / `custom` variant), non-dict `handlers`, missing/non-symbol `proctype`/`procname` — all with the `"di.servers"` error prefix (plain signal; logger not wired yet) |
| `init` — wiring and idempotency | Deps/config recorded; second `init` refreshes refs without re-registering the `.z.pc` handler or the retry job |
| `startup` — live + dead peer | Connects to a live peer, logs (does not raise) on a dead one, excludes self |
| `gethandlebytype` — live remote handle | Returns a genuinely usable handle to a running peer; stats bumped |
| `cleanup` + `retry` | Recover an ungracefully-killed peer (vanished from `key .z.W`) |
| `waitfortype` | Connected case returns `1b`; timeout case returns `0b` |
| `startup` idempotency | A repeat call does not duplicate rows or leak a second handle |
| `readprocesscsv` fail-loud | A reordered / wider header is rejected instead of silently misparsed |
| Input validation + `getapimeta` | Non-symbol proctype/selection etc. rejected; `getapimeta[]` lists exactly the callable exports (plumbing omitted) with the registry columns |
| Real `di.log` integration | Wired against the merged `di.log` logger itself, not just the recording mock |

All green in local KDB-X runs (live-peer suite).

## Notes

- `di.servers` has **no hard `di.*` module dependencies** — `log`, `timer` and `handlers` are injected and all required; `init` signals immediately (plain `'`, logger not yet wired) if any is missing.
- v1 is **TCP only** (`formathp` builds a `` `:host:port `` handle); a future `SOCKETTYPE` config would reintroduce `tcps`/`unix` when there is a real requirement and a test — no unexercised branches ship.
- `hopen` uses the single 2-item timeout form `hopen (handle;timeoutms)` (the dyadic form throws `'rank`); default `HOPENTIMEOUT` is 2000ms.
- The retry period is **10s** via a mode-1 (seconds) timer job — a bare `10000` would have been ~2.8h, the latent typo that made dead-handle recovery never fire in early POCs.
- Module-local `SERVERS` is mutated as a source-level `.z.m.SERVERS` (catenate-and-reassign) so it picks up the compile-time module-local rewrite; a symbol-based `` `.z.m.SERVERS` `` insert would silently target the wrong table.
- Deferred to later sprints per v1 scope: discovery service, password/access-list handling, non-TorQ process tracking. `di.torq` owns central `di.api` registration of the `getapimeta[]` rows.


di/servers/servers.q
```diff
@@ -0,0 +1,252 @@
...
+    .z.m.registered:1b;
+    ];
+  .z.m.loginfo[`init;"di.servers initialised"];
+  };
+
+formathp:{[host;port]
+  / internal - build the tcp connection-handle symbol from a process.csv row. v1 is tcp only; a
+  / future SOCKETTYPE config would reintroduce tcps/unix handling (and a type arg) when there is a
+  / real requirement and a test - we do not ship unexercised branches.
+  lower `$":",(string host),":",string port
```
DI-Software-Engineering, line 196: `retry` is referenced before it is defined. At line 100, `(.z.m.timer[`addjob][`custom])[`serversretry;retry;();10;1;()!()]` passes `retry` by value to the timer. In q, functions are defined top-to-bottom; at the point `init` is *defined* (at load time), `retry` does not yet exist in the namespace. When `init` is *called* at runtime, q evaluates `retry` in the body — and since q looks up names at evaluation time (not definition time for named functions), `retry` must be defined before `init` is called. If `init` is called before `retry` is defined (e.g. during a partial load or a test that loads only part of the file), this will throw `'retry`. The fix is to move `retry`'s definition above `init`, or reference it as `.z.m.retry` so it is resolved via the module slot at call time rather than looked up as a bare name.

di/servers/servers.q
```diff
@@ -0,0 +1,252 @@
...
+    / catenate+reassign, NOT `tablename insert - a symbol-based insert into `.z.m.SERVERS` misses
+    / the compile-time module-local rewrite a source-level .z.m.SERVERS gets, silently targeting the
+    / wrong (literal) table.
+    newrow:([]procname:enlist row`procname;proctype:enlist row`proctype;hpup:enlist hpup;w:enlist w;hits:enlist 0i;startp:enlist $[null w;0Np;.z.p];lastp:enlist .z.p;endp:enlist 0Np);
+    .z.m.SERVERS:.z.m.SERVERS,newrow;
+    } each 0!procs;
+  };
+
+retryrows:{[rows]
+  / internal - reattempt opencon for the given SERVERS row indices, updating w/lastp (and startp on
```
DI-Software-Engineering, line 269: The `startup` idempotency guard checks `procname` uniqueness in `SERVERS`, but `procname` is not guaranteed to be globally unique across proctypes. Two processes with different `proctype` values but the same `procname` (e.g. `procname` defaulting to a common value) would cause one to be silently skipped. The guard should check `(proctype;procname)` pair membership, not `procname` alone: `where not procname in exec procname from .z.m.SERVERS` should be `where not ([]proctype;procname) in exec proctype,procname from .z.m.SERVERS`.

di/servers/servers.q
```diff
@@ -0,0 +1,252 @@
...
+  / already connected. reconnecting a dropped peer is retry's job, not startup's.
+  procs:select from procs where not procname in exec procname from .z.m.SERVERS;
+  if[0=count procs;.z.m.loginfo[`servers;"no new process.csv rows to connect"];:()];
+  {[row]
+    hpup:formathp[row`host;row`port];
+    w:opencon[hpup];
+    if[not null w;.z.m.loginfo[`servers;"connected to ",(string row`proctype),"/",(string row`procname)," at ",string hpup]];
+    / catenate+reassign, NOT `tablename insert - a symbol-based insert into `.z.m.SERVERS` misses
+    / the compile-time module-local rewrite a source-level .z.m.SERVERS gets, silently targeting the
+    / wrong (literal) table.
```
DI-Software-Engineering, line 262: `readprocesscsv` uses `key` on a file symbol to check existence (`0=count key fsym`). For a plain file (not a directory), `key` returns an empty list when the file does not exist, which is correct. However, if `path` is an empty string (the default value of `.z.m.processcsv`), then `fsym` becomes `` `:  `` (colon only), and `key` on that may return an unexpected result rather than an empty list, potentially bypassing the existence check. `startup` guards against an empty `processcsv` before calling `readprocesscsv`, so this is safe in production — but `readprocesscsv` can also be called directly in tests, where the empty-path edge case is unguarded inside the function itself.

di/servers/servers.q
```diff
@@ -0,0 +1,252 @@
...
+  r:getservers[pt];
+  if[0=count r;:0Ni];
+  wh:(selector[r;selection])`w;
+  updatestats[wh];
+  wh
+  };
+
+signalfound:{[pt]
+  / internal - log and return 1b once a connection to pt exists.
+  .z.m.loginfo[`servers;"connected to ",string pt];
```
DI-Software-Engineering, line 334: `waitfortype` converts `pollms` to seconds with `pollms%1000` and passes the result to `system "sleep "`. For a `pollms` value like `200`, this produces `0.2`, so the command becomes `system "sleep 0.2"` — fractional sleep. On Linux this works, but on macOS the BSD `sleep` does not accept fractional seconds, so the test `0b~svc.waitfortype[\`deadproc;700;200]` would hang or error on macOS. Fix: use `system "sleep 0.2"` only if the platform supports it, or convert to an integer (e.g. `1|`long$pollms%1000`) and document the minimum 1-second granularity.

di/servers/test.q
```diff
@@ -0,0 +1,75 @@
...
+/ --- real peer process fixture ---
+FIXDIR:"/tmp/diserverstest";
+isfree:{[p] not @[{hclose hopen x;1b};(`$":localhost:",string p;100);0b]};
+pickport:{[start] first (start+til 500) where isfree each start+til 500};
+PEERPORT:0N; DEADPORT:0N; PEERPID:0N;
+
+waitlisten:{[port;timeoutms]
+  deadline:.z.p+`timespan$1000000*timeoutms;
+  while[(.z.p<deadline) and isfree port; system "sleep 0.05"];
+  not isfree port};
```
DI-Software-Engineering, line 44: `spawnpeer` uses a global assign (`PEERPID::`) to store the peer's PID. If `waitlisten` fails (the peer never comes up) and the function throws before reaching `PEERPID::h ".z.i"`, the handle `h` is opened but never closed, leaking a connection. The fix is to wrap the `hopen`/`".z.i"`/`hclose` block in a protected evaluation, or close `h` in a trap handler.

di/servers/test.q
```diff
@@ -0,0 +1,75 @@
...
+  hclose h;};
+
+killpeer:{[] if[not null PEERPID;@[system;"kill ",string PEERPID;{}]]; PEERPID::0N; system "sleep 0.3";};
+
+setupfixture:{[]
+  / pick two free ports (peer + a never-listening dead one), then write a header'd process.csv
+  / phone book with self, the peer (otherproc), and a dead proctype.
+  PEERPORT::pickport 20000+`int$.z.i mod 20000;
+  DEADPORT::pickport PEERPORT+1;
+  system "mkdir -p ",FIXDIR;
```
DI-Software-Engineering, line 62: In `setupfixture`, the self-row in the generated `process.csv` uses `PEERPORT-2` as the port for `selfproc/selfinst`. This is an arbitrary offset; if `PEERPORT` happens to be `20000` (the minimum), `PEERPORT-2` is `19998`, which may be in use or reserved. More importantly, `pickport` is only called for `PEERPORT` and `DEADPORT` — the self-row port is never checked for availability and is never actually connected to, but if the test ever evolves to dial self (or if `19998` is already in use and causes a CSV parse issue), this silent assumption will break silently. The self-row port should either be a third `pickport` result or a clearly documented sentinel (e.g. `0`).

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **6 warning(s)** | **0 suggestion(s)**

di/servers/servers.q
ascottDI: Incorrect no change

di/servers/servers.q
ascottDI: Incorrect, no change

di/servers/servers.q
ascottDI: Incorrect, no change

di/servers/servers.q
ascottDI: Incorrect, no change

di/servers/test.q
ascottDI: Incorrect, no change

di/servers/init.q
```diff
@@ -0,0 +1,9 @@
+/ connection management and handle-by-type lookup for the modular torq world.
+\l ::servers.q
+/ module version: fallback default here, exported so di.depcheck can read di.servers' version to
+/ satisfy other modules' declared minimums.
+version:"0.1.0";
+/ version.txt in the module folder is the source of truth and takes priority over the fallback above;
+/ read module-relative at load (`:::` resolves to di/servers). a missing/empty file keeps the fallback.
+version:@[{trim first read0 x};`:::version.txt;{[e]version}];
```
DI-Software-Engineering: The trap handler `{[e]version}` references `version` by name at the time the trap fires, not at parse time. Because this line *is* the assignment to `version`, if the read fails the handler will look up `.version` (or the local `version`) in the enclosing scope. At the top level in q there is no closure over the prior assignment, so `version` in the error branch resolves to whatever `version` is in the global namespace at that moment — which happens to be the string just assigned on line 7, so it works *today*. However, this is fragile: the fallback should capture the value explicitly, e.g. `{[e;v]v}[;version]` projected before the call, or simply inline the literal string: `{[e]"0.1.0"}`; either form makes the fallback independent of the mutable global.

DI-Software-Engineering:
## DIReview Summary

di/servers/test.q
ascottDI: Incorrect, no change

di/servers/init.q
ascottDI: Incorrect, no change

di/servers/init.q
```diff
@@ -0,0 +1,8 @@
+/ connection management and handle-by-type lookup for the modular torq world.
+\l ::servers.q
+/ module version, read from the VERSION file rather than hardcoded, so a release bump touches one
+/ plain-text file. read module-relative at load (`:::` resolves to di/servers), and BEFORE the export
+/ line since export:([...]) evaluates each name. NB `version` stays in the export: di.depcheck
+/ resolves a dependency's version from the export dict
+version:first read0`:::VERSION
+export:([init;startup;getservers;gethandlebytype;waitfortype;getapimeta;version])
```
DI-Software-Engineering: The fallback-on-error protection has been removed. The old code used `@[f;x;handler]` to protect against a missing or unreadable `VERSION` file. The new `version:first read0`:::VERSION` will raise an uncaught signal at load time if the file is absent, empty, or unreadable, breaking module load entirely. Wrap it: `version:@[{first read0 x};`:::VERSION;{"0.0.0"}]` (or whichever fallback is appropriate).

di/servers/servers.q
```diff
@@ -0,0 +1,269 @@
...
+  procs:select from procs where not isme;
+  procs:select from procs where proctype in conns;
+  / SELF-CONNECTION GUARD. the exclusion above is an exact (proctype;procname) match against the
+  / identity from config. if process.csv disagrees - a drifted proctype for this procname - the self
+  / row is not recognised and this process dials ITSELF. procname is process.csv's unique key (the
+  / idempotency filter below relies on that too), so a surviving row carrying our procname IS us:
+  / drop it and say why, rather than opening a self-connection nobody would think to look for.
+  / checked AFTER the connections filter: a row that could never be connected to was never at risk
+  if[count mismatched:select from procs where procname=pn;
+    .z.m.logwarn[`startup;"process.csv lists procname ",(string pn)," as proctype ",
```
DI-Software-Engineering, line 247: `first mismatched\`proctype` is used to build the warning message, but `mismatched` is filtered from `procs`, which has already been narrowed to `proctype in conns`. A drifted self-row whose proctype is NOT in `connections` is silently dropped by the earlier filter and never reaches this guard — the warning comment says "checked AFTER the connections filter" as if that is intentional, but then the comment in the doc says the guard fires for any surviving row carrying our procname. If the misconfigured proctype happens to equal a configured connection type, the guard works; otherwise no warning is issued. The guard should be applied to the full `procs` list before the connections filter (or the comment/doc should be corrected to clearly state the limited scope).

di/servers/test.csv
```diff
@@ -0,0 +1,76 @@
...
+true,0,0,q,1=count svc.getservers[`otherproc],1,1,repeat startup added no duplicate otherproc row (guard skips already-tracked procs)
+comment,,,,,,,readprocesscsv fails loud on a reordered/wider header instead of silently misparsing
+run,0,0,q,"svc.init[`log`timer`handlers`proctype`procname`connections`processcsv!(mocklog;rtmr;mockhandlers;`selfproc;`selfinst;`rdb;writebadcsv[])]",1,1,re-init pointing at a process.csv whose header is reordered
+fail,0,0,q,svc.startup[],1,1,startup signals on the bad header rather than silently misreading columns
+comment,,,,,,,input validation + getapimeta
+fail,0,0,q,"svc.getservers[""nosuch""]",1,1,getservers rejects a non-symbol proctype
+
+comment,,,,,,,"startup must not dial ITSELF when process.csv disagrees with the configured identity"
+run,0,0,q,"svc.init[`log`timer`handlers`proctype`procname`connections`processcsv!(mocklog;rtmr;mockhandlers;`selfproc;`selfinst;`rdb;writedriftcsv[])]",1,1,re-init against a csv listing OUR procname (selfinst) under a different proctype
+run,0,0,q,delete from `logrows,1,1,clear the log capture
```
DI-Software-Engineering, line 118: The warning-match string is `"connecting to myself"` but the actual warning message constructed in `servers.q` contains the phrase `"identity drift, skipping that row rather than connecting to myself"`. Depending on how `warnlogged` does substring matching this may pass, but if it requires an exact match the test will always fail. Verify that `warnlogged` performs a substring search, or align the expected string with the actual message.

DI-Software-Engineering:
## DIReview Summary

di/servers/init.q
ascottDI: Incorrect, no changes

di/servers/servers.q
ascottDI: Incorrect about the code. valid in respect to the documentation

di/servers/test.csv
ascottDI: Incorrect, no change.

di/servers/servers.q
```diff
@@ -0,0 +1,371 @@
...
+  / NB this @[...] IS protected-apply only because init guarantees `deletejobs is present and is a
+  / function. were the key missing, x would be the null-shaped dict a dict-valued dep returns for an
+  / absent key and q would read the line as three-argument amend instead, deleting nothing and
+  / logging nothing while teardown reported success. the init check is what keeps this line honest
+  @[.z.m.timer[`deletejobs];`serversretry;
+    {[e] .z.m.logwarn[`teardown;"could not delete the serversretry timer job: ",e]}];
+  / reset the SIDE-EFFECT guard only, so a later init registers both again. initialised[] probes
+  / .z.m.loginfo, not this, so everything else stays callable after a teardown - see initialised[]
+  .z.m.registered:0b;
+  .z.m.loginfo[`teardown;"di.servers .z.pc registration and serversretry job removed"];
```
DI-Software-Engineering, line 189: `teardown` updates `registered` via `.z.m.registered:0b` but reads `registered` (the module-level variable) in `init`'s idempotency guard. If `.z.m` is the module namespace and these are the same slot this is fine, but if `registered` in `init` resolves as a local/different binding the reset will not be seen and the idempotency guard will never re-allow registration after a teardown. Verify that every read of `registered` inside `init` goes through `.z.m.registered` (or the same slot) so the teardown reset is actually observed.

di/servers/servers.q
```diff
@@ -0,0 +1,371 @@
...
+  if[not all (`info`warn`error) in key deps`log;
+    '"di.servers: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  if[99h<>type deps`timer;
+    '"di.servers: timer value must be a dict (see di.timer)"];
+  if[not `addjob in key deps`timer;
+    '"di.servers: timer dict must expose `addjob (see di.timer)"];
+  if[99h<>type deps[`timer]`addjob;
+    '"di.servers: timer`addjob must be a variant dict (see di.timer addjob.custom/default/simple)"];
+  if[not `custom in key deps[`timer]`addjob;
+    '"di.servers: timer`addjob must expose the `custom variant [id;func;params;period;mode;opts]"];
```
DI-Software-Engineering, line 103: The closing `]` for the `if[not `deletejobs in key deps`timer; ...]` block is missing — the two-line string literal forming the error message is never closed with `]`. This leaves the subsequent `if[99h<>type deps`handlers;...]` parsed as part of the first `if` body, so the handlers type check is only evaluated when `deletejobs` is absent. A timer dep that has `deletejobs` but passes a non-dict `handlers` value will not be caught. Add the missing `]` after the `"di.timer"]` string to close the first `if` before opening the next.

di/servers/test.q
```diff
@@ -0,0 +1,87 @@
...
+/ cleanup, tested in isolation from the auto .z.pc hook (which di.handlers would install for real).
+handlercalls:([]event:`symbol$();name:`symbol$());
+/ remove DELETES the recorded row (it used to be a no-op), so teardown's release is observable the
+/ same way register's is - a no-op remove would let a broken teardown pass every assertion.
+mockhandlers:`register`remove`list!(
+  {[ev;ph;nm;pri;fn]`handlercalls upsert(ev;nm)};
+  {[ev;ph;nm] handlercalls::delete from handlercalls where event=ev,name=nm;};
+  {[ev]});
+
+warnlogged:{[s] any (exec msg from logrows where lvl=`warn) like "*",s,"*"};
```
DI-Software-Engineering, line 30: `mockhandlers`remove` uses a global amend (`handlercalls::delete from handlercalls where event=ev,name=nm`). In a module-namespaced context the bare name `handlercalls` on the right-hand side of `::` resolves to root (`.handlercalls`), but the table was defined at the top of this file without a namespace qualifier. If the test file is loaded inside a namespace (as `.Q.m.mp` / k4unit may do), the global assign will write to a different slot than the one `register` upserts into, so teardown assertions that count rows in `handlercalls` will always see the pre-teardown count and the 'removed' assertion will pass vacuously. Use an explicit path (e.g. `.t.handlercalls`) for both the table definition and all references to it.

DI-Software-Engineering:
## DIReview Summary

PR #121, opened 2026-08-11
Feature clienttracking

ascottDI:
# di.clienttracking

## Summary

Adds `di.clienttracking`, a `di.*` module that tracks the client sessions connected to a KDB-X
process in an in-memory table: who is connected, from where, when each session opened and closed,
and how many requests and result-bytes each client has been served. It is the modular replacement
for TorQ's `code/handlers/trackclients.q` (the `.clients` namespace).

The module registers all of its `.z.*` hooks through an injected `di.handlers` instance and never
assigns `.z.*` directly, so it composes with any other component hooking the same events.

Depends on `di.handlers` and `di.log` (both merged). Framework tier; no hard `use` dependencies —
all providers are injected, so the module loads and is tested standalone.

## Motivation

`trackclients.q` maintains a `CLIENTS` session table by hand-rolling `.z.*` handler chaining via
`.dotz.set`. In the modular framework, ownership of the `.z.*` events belongs to `di.handlers`, so
this extraction re-expresses the tracking logic on top of the `di.handlers` registration contract
while preserving the observable behaviour (a queryable table of current and recent client sessions).

### Event handling

`di.handlers` distinguishes two dispatch models, and this module uses both:

- **Connection lifecycle** — `.z.po`, `.z.pc`, `.z.wo`, `.z.wc` are *simple* events (return value
  discarded), fanned out to every registrant. The module registers monadic observers that open a
  session row on connect and stamp `endp` on disconnect.
- **Usage counting** — `.z.pg`, `.z.ps`, `.z.ws` are *phased* events whose return value is the query
  result. Counting only observes that result, so it is registered as a **`post`** handler.

### Usage-counting activation

`di.handlers` does not accept a `post` registration on a phased event until that event has an `exec`
owner. `init` therefore registers the lifecycle observers unconditionally and then calls
`enableusage[]`, which registers the usage `post` handler on each query event that already has an
owner and skips (with a `warn`) any that do not. `enableusage[]` is exported and idempotent so it can
be called again once the process's query owner (e.g. a gateway or `di.permissions`) is registered.

> **Integration note for `di.torq`:** call `clienttracking.enableusage[]` after wiring the process's
> query `exec` owner. Lifecycle tracking requires no such step.

### Dependencies (injected via `init`)

| Dependency | Key | Required | Description |
|---|---|---|---|
| logger | `` `log `` | yes | dict of binary `{[c;m]}` functions with at least `info`, `warn`, `error` |
| handlers | `` `handlers `` | yes | a `di.handlers` instance's `register` / `remove` / `list` functions |

`init` validates both strictly and signals immediately if either is missing or malformed; there is no
fallback. `di.log`'s `logdict`log` (an `info..fatal` dict) is a conforming superset of the required
log keys and is passed through without adaptation. Optional config keys — `maxidle` (timespan),
`retain` (timespan), `trackusage` (boolean) — are type-checked when present.

Although the dependency tree lists `di.clienttracking → di.handlers`, handler management is an
*injected* dependency per the framework contract, not a hard `use` import; `di.torq` supplies the
shared, initialised `di.handlers` instance to `init`.

## Session table

`getclients[]` returns the table; each row is one session, and an open session has a null `endp`:

| Column | Type | Meaning |
|---|---|---|
| `w` | `` `g#int `` | connection handle (`.z.w` at open) |
| `ipa` | symbol | client IP, dotted-decimal |
| `u` | symbol | client user (`.z.u` at open) |
| `a` | int | client IP, raw int (`.z.a` at open) |
| `startp` | timestamp | connection open time |
| `endp` | timestamp | connection close time; null while open |
| `lastp` | timestamp | time of the last request from this client |
| `hits` | long | requests served |
| `sz` | long | approximate total result-bytes returned |

## Changes from the TorQ implementation

| Change | Rationale |
|---|---|
| **Unkeyed table** (grouped `w`) instead of keyed-on-handle | TorQ keyed `CLIENTS` on the handle and nulled the key on close, which collides on handle reuse; an unkeyed table represents concurrent and historical sessions per handle cleanly (current session = row with that `w` and null `endp`). |
| **`errs` column removed** | A `post` handler runs only after a successful `exec`, so it cannot observe failed queries; the column would always be zero. |
| **INTRUSIVE mode removed** | TorQ optionally sent an async `eval` back to each client to self-report host details. It only works for a cooperating q client and is security-sensitive; removed, along with the columns it populated. |
| **No timer dependency** | Cleanup runs inline on open/close and is exported as `cleanup[]` for a host to schedule via `di.timer`, keeping the injected surface to `log` + `handlers`. |

## Public API

| Function | Signature | Description |
|---|---|---|
| `init` | `[deps]` | Wire dependencies and optional config, create the table, register lifecycle handlers, and enable usage counting where possible. Idempotent. |
| `getclients` | `[]` | Return the session table. |
| `addclient` | `[handle]` | Manually record an int handle as an open session (equivalent to TorQ's `addw`). |
| `cleanup` | `[]` | Run a maintenance sweep: mark departed handles closed, force-close handles idle beyond `maxidle`, and purge closed rows older than `retain`. Also runs automatically on every open/close. |
| `enableusage` | `[]` | (Re)register usage counting on query events that now have an `exec` owner. |
| `version` | — | Module version (`"0.1.0"`). |

`getapimeta[]` exposes the callable API above for central registration with `di.api`; the framework
plumbing (`init`, `getapimeta`, `version`) is intentionally excluded.

## Testing

Run against the real, merged `di.handlers` and `di.log` — no dependencies are mocked.

**Unit suite** — `test.csv`, 41 checks, hermetic (no sockets). Drives `di.handlers`' dispatcher by
invoking the function bound to each `.z.*` event, exercising the full registration → dispatch → update
path. Covers dependency and config validation, the four lifecycle registrations, connect/disconnect
tracking, `addclient` and its input validation, cleanup of departed handles, usage-counting activation
and the deferral path when no owner exists, teardown on `trackusage:0b`, and the API-metadata/version
contract.

```q
k4unit:use`di.k4unit
k4unit.moduletest`di.clienttracking
```

**Integration suite** — `test_integration.csv`, 6 checks. Stands up a child q process, tracks the
handle to it, and exercises the idle-reap path (which requires a genuinely live handle): it verifies
that an idle handle is force-closed and its session stamped closed, and that the child is cleaned up.
Requires a q/kdb-x binary on `QHOME` and skips cleanly if none is available. `moduletest` loads only
`test.csv`, so run the integration suite directly:

```q
k4unit:use`di.k4unit
.m.di.0k4unit.KUltf .Q.dd[hsym`$.Q.m.mp`di.clienttracking;`test_integration.csv]
.m.di.0k4unit.KUrt[]
k4unit.getresults[]
```

Incoming-connection acceptance (`.z.po` firing on a real socket) is not integration-tested: a process
only accepts inbound connections at its top-level event loop, which the test harness does not reach.
That binding is the responsibility of `di.handlers` and is covered by its own integration suite; here
the `.z.po`/`.z.pc` handlers are exercised through the unit suite.

## Limitations

- **Usage counting requires a single-threaded query port.** A `post` handler runs in the query's
  execution context; on a multithreaded (negative `\p`) port that is not the main thread, so its
  update to the session table hits kdb's `'noupdate` restriction. `di.handlers` isolates the handler,
  so the query still succeeds and the update is skipped with a logged warning. Lifecycle tracking is
  unaffected.
- **Usage counting is activation-ordered** — it attaches only to query events that already have an
  `exec` owner; re-run `enableusage[]` after an owner is registered.
- `ipa` formats `.z.a` as a dotted-decimal address without reverse-hostname resolution.

## Files

```
di/clienttracking/init.q                  entry point and export list
di/clienttracking/clienttracking.q        implementation
di/clienttracking/clienttracking.md       module documentation
di/clienttracking/test.csv                unit tests (41)
di/clienttracking/test_integration.csv    integration tests (6)
```


di/clienttracking/clienttracking.q
```diff
@@ -0,0 +1,175 @@
...
+
+/ ============================================================
+/ internal helpers
+/ ============================================================
+
+ipa:{[a]
+  / format a raw .z.a int ip address as a dotted-decimal symbol
+  :`$"." sv string "i"$0x0 vs a;
+  };
+
```
DI-Software-Engineering, line 43: The `ipa` helper casts each byte of the 4-byte big-endian representation with lowercase `"i"$`, which interprets each byte as a signed int8 (range -128..127). For addresses with any octet ≥ 128 (e.g. 192.168.x.x) this produces negative values such as "-64" instead of "192". Use `"i"$"x"$` (`"x"$a` to get 4 bytes then `"i"$` each) or express as `string each 0xff and"j"$0x0 vs a` to get unsigned octets.

di/clienttracking/clienttracking.q
```diff
@@ -0,0 +1,175 @@
...
+  .z.m.logerr[ctx;msg];
+  '"di.clienttracking: ",string[ctx],": ",msg;
+  };
+
+track:{[h]
+  / record a newly-seen client handle - sweep first, then append an open session row for h
+  cleanup[];
+  .z.m.clients:.z.m.clients upsert (h;ipa .z.a;.z.u;.z.a;.z.p;0Np;.z.p;0;0);
+  };
+
```
DI-Software-Engineering, line 55: `track` appends a row with `hits:0` and `sz:0` as int atoms (type -6h), but the `clientschema` declares both columns as `long$()` (type 7h). Appending mismatched types causes a type error at runtime on the `upsert`. Use `0j` (long zero) for both: `(h;ipa .z.a;.z.u;.z.a;.z.p;0Np;.z.p;0j;0j)`.

di/clienttracking/clienttracking.q
```diff
@@ -0,0 +1,175 @@
...
+    '"di.clienttracking: handlers dict must have `register`remove`list keys; got: ",(", " sv string key deps`handlers)];
+  / optional config - validate types up front (nested if, not `and`, to avoid eager-eval on an absent key)
+  if[`maxidle in key deps;
+    if[not -16h=type deps`maxidle;'"di.clienttracking: maxidle must be a timespan"]];
+  if[`retain in key deps;
+    if[not -16h=type deps`retain;'"di.clienttracking: retain must be a timespan"]];
+  if[`trackusage in key deps;
+    if[not -1h=type deps`trackusage;'"di.clienttracking: trackusage must be a boolean"]];
+  .z.m.loginfo:(deps`log)`info;
+  .z.m.logwarn:(deps`log)`warn;
```
DI-Software-Engineering, line 115: The `if[not @[{.z.m.clients;1b};::;0b];...]` idiom used to detect first-init relies on `.z.m.clients` throwing an error if the name is unset. This works but silently suppresses *any* error from evaluating `.z.m.clients` (e.g. a corrupt value), treating it as "not yet initialised" and overwriting the table. A more robust guard is to check `not `.clients in key .z.m` (or whatever the module slot is), so only genuine absence triggers the reset.

di/clienttracking/test_integration.csv
```diff
@@ -0,0 +1,47 @@
...
+before,0,0,q,ct:use`di.clienttracking,1,1,load di.clienttracking
+before,0,0,q,logger:use`di.log,1,1,load the real di.log
+before,0,0,q,logdep:logger.logdict`log,1,1,di.log level dict
+before,0,0,q,handlers.init[enlist[`log]!enlist logdep],1,1,init di.handlers
+before,0,0,q,"hdep:`register`remove`list!(handlers.register;handlers.remove;handlers.list)",1,1,handlers injectable
+before,0,0,q,"ct.init[`log`handlers!(logdep;hdep)]",1,1,init di.clienttracking
+before,0,0,q,qcands:$[count qh:getenv[`QHOME];(raze(qh;"/bin/q");raze(qh;"/";string .z.o;"/q"));()],1,1,candidate q binary locations
+before,0,0,q,qbin:$[count ex:qcands where {not ()~key hsym `$x} each qcands;first ex;"q"],1,1,pick an existing binary else q on PATH
+before,0,0,q,.it.wait:{[p;n] $[n<=0;0Ni;not null h:@[{hopen(`$"" sv (":localhost:";string x);500)};p;0Ni];h;[system"sleep 0.3";.it.wait[p;n-1]]]},1,1,poll for the child to accept
+before,0,0,q,system"p 0W",1,1,ask the OS for a free ephemeral port
```
DI-Software-Engineering: The port-stealing race: `system"p 0W"` asks the OS for a free port, `system"p"` reads it, then `system"p 0"` releases the listener before the child binds it. Between the release and the child's `hopen` another process can claim that port, causing the child to fail to bind (or bind a different port silently). A safer approach is to keep the parent listening (skip `system"p 0"`) and close it only after the child is confirmed up, or to use a pre-chosen fixed test port with a retry loop.

di/clienttracking/clienttracking.q
```diff
@@ -0,0 +1,175 @@
...
+  if[99h<>type deps`log;
+    '"di.clienttracking: log value must be a dict; pass `info`warn`error functions"];
+  if[not all `info`warn`error in key deps`log;
+    '"di.clienttracking: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  if[not `handlers in key deps;
+    '"di.clienttracking: handlers dependency is required; pass di.handlers register/remove/list keyed on `handlers"];
+  if[99h<>type deps`handlers;
+    '"di.clienttracking: handlers value must be a dict; pass register/remove/list functions"];
+  if[not all `register`remove`list in key deps`handlers;
+    '"di.clienttracking: handlers dict must have `register`remove`list keys; got: ",(", " sv string key deps`handlers)];
```
DI-Software-Engineering, line 106: `disableusage` calls `.z.m.handlers[`remove][;`post;`clienttracking] each usageevents` but the `remove` function is invoked with the event as the first argument (via `each`). Verify the `di.handlers.remove` signature matches this 3-argument partial-application pattern; if `remove` expects `[event;phase;name]` as a single call rather than `remove[event][phase;name]`, the projection is mis-formed and will silently no-op or error, leaving stale post handlers registered even after `trackusage:0b` re-init.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **5 warning(s)** | **0 suggestion(s)**

di/clienttracking/clienttracking.q
ascottDI: Incorrect, no change

di/clienttracking/clienttracking.q
ascottDI: Incorrect, no change

di/clienttracking/clienttracking.q
ascottDI: Correct, added guard so this works correctly.

di/clienttracking/test_integration.csv
ascottDI: valid race, fixed with new port set up procedure

di/clienttracking/clienttracking.q
ascottDI: Incorrect, no change

di/clienttracking/test_integration.csv
```diff
@@ -0,0 +1,46 @@
...
+before,0,0,q,qcands:$[count qh:getenv[`QHOME];(raze(qh;"/bin/q");raze(qh;"/";string .z.o;"/q"));()],1,1,candidate q binary locations
+before,0,0,q,qbin:$[count ex:qcands where {not ()~key hsym `$x} each qcands;first ex;"q"],1,1,pick an existing binary else q on PATH
+before,0,0,q,.it.wait:{[p;n] $[n<=0;0Ni;not null h:@[{hopen(`$"" sv (":localhost:";string x);500)};p;0Ni];h;[system"sleep 0.3";.it.wait[p;n-1]]]},1,1,poll for the child to accept
+before,0,0,q,.it.spawn:{[qbin;n] if[n<=0;:(0Ni;0N)];system"p 0W";p:system"p";system"p 0";pid:"J"$last system "" sv (qbin;" -q -p ";string p;" </dev/null >/dev/null 2>&1 & echo $!");h:.it.wait[p;20];ok:$[null h;0b;pid=@[h;".z.i";{0N}]];$[ok;(h;pid);[if[not null h;@[hclose;h;::]];@[{system "" sv ("kill -9 ";string pid)};::;{}];.it.spawn[qbin;n-1]]]},1,1,bind a free port and confirm via PID we reached our own child (retries past a stolen port)
+before,0,0,q,sp:.it.spawn[qbin;5],1,1,spawn the child with up to five attempts
+before,0,0,q,h:sp 0,1,1,the confirmed outgoing handle to the child
+before,0,0,q,cpid:sp 1,1,1,the confirmed child PID
+before,0,0,q,if[null h;exit 0],1,1,skip cleanly if no confirmed child came up
```
DI-Software-Engineering, line 22: The `.it.spawn` retry loop calls `system "p 0W"` / `system "p"` / `system "p 0"` to probe for a free port, but between `system "p 0"` (releasing the listener) and the child process binding that same port, another process on the OS can steal it. If the child ends up on a different port the PID-verification check (`pid=@[h;".z.i";{0N}]`) will catch the mismatch and retry, which is correct. However the child process launched with the stolen-port scenario is left running: `@[hclose;h;::]` closes the handle but there is no kill for the wrongly-spawned child whose port was stolen by a third party (the kill only fires when `ok` is false, and in the stolen-port case `ok` is false so the kill does fire via `@[{system "" sv ("kill -9 ";string pid)};::;{}]`). Actually that path is fine. But if `hopen` connects to a *different* pre-existing server on that port (not the child we launched), `h` is non-null, `pid` won't match `.z.i`, `ok` is false, we hclose that innocent third-party handle, and then kill a PID that is our own child (which may be listening on a completely different port and is still running). The kill is sent to `pid` (the child we launched) which is correct — that child should be killed. The innocent third-party handle is closed, which is a side-effect but not data corruption. The real defect is that on retry, `pid` is reassigned from the new `system` launch but the *previous* child (whose port was stolen) is already killed, so no leak occurs. On closer inspection the logic is consistent. Disregard — no defect found here.

DI-Software-Engineering:
## DIReview Summary

di/clienttracking/test_integration.csv
ascottDI: "no defect found"

alowrydi:
testing approval permissions

alowrydi (approved):
testing permissions

alowrydi (commented):
test

PR #122, opened 2026-08-11
Feature heartbeat

alowrydi:
**Extracts TorQ's `code/common/heartbeat.q` (the `.hb` namespace), di.heartbeat**

Publishes a periodic liveness beat over pub/sub so downstream monitors can detect that a process has stalled or blocked even while its connection is still valid, and carries the monitor role in the same module: subscribing to other processes' beats, tracking the last one seen per process, and raising warning/error transitions when they stop arriving.

Builds on an initial implementation from PR #109 (Olly, June 2026, closed unmerged when he left the project) - the domain model (state machine, tolerance-multiplier grace periods, mode-2 timer scheduling with the catch-up-storm reasoning, single-arg `init`) carried through largely as designed. The integration layer is rewritten against contracts that didn't exist as merged code at the time it was written - `di.log`, `di.handlers`, `di.config`, `di.api` all merged in early August, after #109 was closed - and two further review passes since then found and fixed a blocking-availability hazard and a stale-handle bug in the async rewrite itself, detailed below.

Trello ticket - https://trello.com/c/093fht6Z/95-kdb-x-heartbeat

---

**Files created**

| File | Description |
| -- | -- |
| `di/heartbeat/init.q` | Loads `heartbeat.q`, defines export of 11 functions |
| `di/heartbeat/heartbeat.q` | Full implementation - module state, internal helpers, publisher/monitor logic, init |
| `di/heartbeat/heartbeat.md` | Full module documentation - see below |
| `di/heartbeat/VERSION` | Plain-text version string, read by `init` and resolved by `di.depcheck` |
| `di/heartbeat/deps.q` | Empty - no hard module dependencies, everything injected |
| `di/heartbeat/test.csv` | 469 k4unit unit tests |
| `di/heartbeat/test.q` | Integration-suite fixture - spawns and drains a real child publisher process |
| `di/heartbeat/test_integration.csv` | 50 k4unit integration tests against the spawned process |

---

**How to test**

```q
k4unit:use`di.k4unit
k4unit.moduletest`di.heartbeat
```

469/469 unit tests passing, 0 failures.

The integration suite needs a real second process and is a separate file, so `moduletest` (hardcoded to `test.csv`) doesn't pick it up:

```q
.m.di.0k4unit.KUltf .Q.dd[hsym`$.Q.m.mp`di.heartbeat;`test_integration.csv]
.m.di.0k4unit.KUrt[]
```

50/50 integration tests passing, 0 failures. Exercises what a mock can't reach: the remote-subscribe handshake (`.z.w` resolving correctly inside an inbound async call), the root schema table `di.pubsub` discovers, a published row actually crossing the wire into the monitor's store, a genuinely stalled peer not blocking the sweep, and - wired against real `di.log`/`di.timer`/`di.handlers`/`di.pubsub` rather than mocks - the real `di.timer` scheduler actually *firing* a registered job, not just holding it in a registry.

qlint: 0 error-level findings.

---

**Design decisions**

*1. The remote subscribe is asynchronous, not synchronous like legacy.* A sync call blocks for as long as the peer takes to answer, and a peer that's alive but stalled - a GC pause, a heavy query, exactly the condition this module exists to detect - never answers promptly. Because the subscribe runs inside a `di.timer` job on a single thread, that block stalled `publishheartbeat` and `checkheartbeat` too: the monitor fell silent to its own monitors at precisely the moment a peer misbehaved. Measured on real KDB-X: 10s blocked against a 10s-hung peer, versus 36µs for the async send. `hopen`'s timeout doesn't help here - it bounds connection establishment, not later requests on an already-open handle.

*2. There is no cache of subscribed handles.* An earlier version tracked which handles it had already subscribed, to skip them on later sweeps. Removed rather than repaired: kdb+ reissues the lowest free descriptor immediately (measured: handle `4` → close → reopen → handle `4`), and `hclose` doesn't fire `.z.pc` at all (also measured), so a stale cache entry made the monitor skip a live peer *permanently*, with nothing in any log to explain it. `di.pubsub` already dedupes subscribers by `.z.w`, so the cache bought nothing. Removing it also removed this module's only reason to register `.z.pc`, and with it the `handlers` dependency entirely.

*3. Async can't report a remote failure, so two checks watch the consequence instead.* A subscribed peer that never beats, and a monitor that discovers no usable peer at all. Both warn once at a configurable sweep threshold, distinguish a cold start from genuine peer loss, and close out with a matching-severity recovery line so an alert can never dangle unresolved in a log.

*4. A pre-existing root `heartbeat` table is never adopted, not even a column-identical one.* `init` errors instead. This is a real, reachable collision, not a hypothetical: `di.subscriptions` installs subscribed tickerplant schemas at root, so a monitor watching a TP that carries `heartbeat` already has one. Column compatibility is deliberately not treated as a safety signal - a matching shape isn't evidence a table means the same thing, and adopting on that basis would silently co-mingle this module's liveness rows with whatever the real owner stores there.

*5. Every timer entry point that can throw is isolated against `di.timer`'s `disableonfail:1b` default.* An unprotected throw in a client callback, a transient pub/sub outage, or a dependency error wouldn't just skip one beat - it would permanently disable the job, silently ending the exact monitoring this module exists to provide. `onwarning`/`onerror`/`publish` and the whole discovery sweep run isolated; state is always updated before an isolated callback fires, so a callback failure can never leave the store inconsistent.

*6. Config defaults take the values TorQ actually shipped, not the fallbacks written inside `heartbeat.q` itself.* Legacy carried two disagreeing sets - the in-file `@[value;...]` defaults (`1.5f`, `2f`, `()`) and what `config/settings/default.q` actually set (`2f`, `3f`, `` `ALL ``). This module takes the shipped values, since they reflect what really ran; the in-file `connections:()` in particular could not have been used at all; it makes `` `ALL in () `` false and the entire monitor path silently monitors nothing.

---

**Checklist**
- [ ] 469/469 unit tests passing
- [ ] 50/50 integration tests passing
- [ ] `heartbeat.md` documents all exported functions, config, usage examples and notes
- [ ] No hard dependencies on other `di.*` modules - standalone (`deps.q` empty)

---

**Documentation**

See `heartbeat.md` for full reference including the dependency table, configuration options with shipped-vs-in-file defaults noted per key, exported function documentation with examples, the root-publication and remote-subscribe mechanics, and a dedicated section on deliberate departures from legacy.ection on deliberate departures from legacy.


di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+  / NB no handlers requirement: this module registered a .z.pc observer solely to prune a cache of
+  / subscribed handles, and that cache is gone. a handlers dict passed anyway is silently accepted
+  / (see depkeys) so di.torq's uniform wiring keeps working unchanged
+  if[99h<>type deps`servers;
+    '"di.heartbeat: subenabled is set, so a servers dependency is required (see di.servers)"];
+  if[not `getservers in key deps`servers;
+    '"di.heartbeat: servers dict must expose `getservers (see di.servers)"];
+  };
+
+resolveconfig:{[deps]
```
DI-Software-Engineering, line 198: validhandles drops handles where `null h` OR `0i>=h`, but the condition `0i>=h` is true for ALL non-positive integers including valid negative async handles — yet negative handles are specifically called out as load-bearing to drop (neg of a negative becomes positive). The logic is correct for the intended drop, but for a positive handle `h`, `0i>=h` is false, so it passes. For a null handle `null h` is true. For handle 0i, `0i>=0i` is true. For a negative handle e.g. `-5i`, `0i>=-5i` is true. This all appears correct — BUT the condition is `not (null h) or 0i>=h`, which by q's right-to-left evaluation is `not ((null h) or (0i>=h))`. A null integer `0Ni` satisfies `null h` → true, so `or 0i>=h` → true, `not` → false: dropped correctly. Handle `0i`: `null 0i` → false, `0i>=0i` → true, overall false: dropped correctly. Handle `-5i`: `null -5i` → false, `0i>=-5i` → true, overall false: dropped correctly. However the test at line 562 (`test.csv`) asserts `(enlist 4i)~.m.di.0heartbeat.validhandles -5 4 0Ni,0i` — this passes only if `-5i` is dropped, which requires `0i>=-5i` to be 1b. In q, comparing an int `0i` with a long `-5` involves a type promotion. The input `-5 4 0Ni,0i` is a mixed long/int/null list: `-5` and `4` are longs, `0Ni` is a null int, `0i` is int zero. The `0i>=h` comparison where h contains longs will compare int 0i against long -5, which is a type mismatch (`'type`). This means `validhandles` will throw when given a mixed-type integer list containing longs and ints together.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+  / a peer we have subscribed to that never produces a beat. a null counter means an addprocs seed
+  / rather than an observation - the same discriminator evictstale uses. warned once, at exactly the
+  / threshold sweep, so a permanently broken peer does not log forever.
+  / NB the count is rebuilt from the CURRENT discovered set each sweep, so a peer that drops out of
+  / di.servers' view and comes back starts again from 1. that is deliberate: a connectivity blip is
+  / not the continuously-stuck subscribe this warning exists to catch, and only an uninterrupted run
+  / of pending sweeps should trip it
+  if[not all `proctype`procname in cols rows;:()];
+  / the threshold is hoisted into a LOCAL first: a qsql where clause inside module code cannot resolve
+  / a module-level name and throws 'subscribewarnsweeps. same reason checkheartbeat hoists wp and ep
```
DI-Software-Engineering, line 392: `warningperiod` and `errorperiod` are defined as functions of `processtype` but read the module-level names `warningtolerance`/`errortolerance` and `publishinterval` as bare names. In `checkheartbeat`, these are called as `warningperiod each t\`sym` and `errorperiod each t\`sym`, which iterates over every row. Since the tolerance and interval are uniform (no per-type override is implemented), this computes the same value N times rather than once. More importantly, if the module state for these names is ever read before `init` populates them (e.g. if `checkheartbeat` is somehow invoked after a failed re-init that cleared state), they resolve to the top-level constants rather than `.z.m` state. The functions silently fall back to module-level bare names instead of always reading `.z.m.warningtolerance` etc., which is inconsistent with every other config access in the module.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+
+warnemptysweeps:{[]
+  / internal - count consecutive sweeps that discovered no usable peer at all, and warn ONCE at the
+  / threshold. deliberately cause-AGNOSTIC: it watches the consequence, not any single cause, so it
+  / stays correct whatever the reason - a connections list naming a process type di.servers never
+  / connects to, a servers dependency returning nothing, or peers that are simply all down. that also
+  / means it needs no maintenance as those causes come and go - it survived the di.servers getservers
+  / fix unchanged. reuses subscribewarnsweeps: it is the same "consecutive sweeps" unit as the
+  / never-beaten warning
+  .z.m.emptysweeps:emptysweeps+1;
```
DI-Software-Engineering, line 446: In `emptysweepmsg`, the bare name `subscribewarnsweeps` is read directly. All other config is accessed via `.z.m` at call sites (e.g. `.z.m.logwarn`, `.z.m.emptysweeps`), but `subscribewarnsweeps` here is a bare name. In this module's convention, mutable state lives in `.z.m` and is written as `.z.m.subscribewarnsweeps` by `init`. A bare read of `subscribewarnsweeps` resolves to the `configdefaults` list entry (`3`), not the configured value. If a caller sets `subscribewarnsweeps` to a different value, the message always reports the default `3` instead of the actual threshold. The same bare-name read issue appears in `warnemptysweeps` (line 447) and `trackunseen` hoists it into a local `n` (line 405) as a workaround — but `emptysweepmsg` skips that workaround. Fix: hoist `subscribewarnsweeps` into a local or use `.z.m.subscribewarnsweeps` (if the module loader maps bare names to `.z.m` reads, this may be intentional — but it is inconsistent with the workaround already applied in `trackunseen`).

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+  if[0=count handles;warnemptysweeps[]];
+  / only peers we ACTUALLY attempted are candidates for the never-beaten warning. a row carrying a
+  / null handle (a disconnected server) or handle 0 was deliberately skipped, so warning that it never
+  / heartbeated would be a false alarm about a peer we never asked
+  trackunseen[select from rows where w in handles];
+  };
+
+hbsubscriptions:{[]
+  / timer job - pick up any newly connected publisher of a configured process type. isolated for the
+  / same reason the publish call and the callbacks are: di.timer's addjob.opts defaults
```
DI-Software-Engineering, line 517: `warn` upserts only `sym`, `procname`, and `warning:1b` into `hb`, leaving all other columns (including `time`, `counter`, `error`) to their existing keyed values. This is fine for an update. However `err` (line 516) upserts only `sym`, `procname`, `error:1b` — deliberately not setting `warning`. For a process that jumps past both thresholds in a single check, `checkheartbeat` calls `warn newwarn` then `err newerr`, where `newwarn` contains `status=1` rows and `newerr` contains `status>1,not error` rows. A process at status=2 with `not warning` and `not error` is in `newerr` but NOT in `newwarn` (since `status=1` is false). So `warn` is never called for it, and `error` is set without `warning` ever having been set — which the test suite asserts as correct (the "jumping past both thresholds" test at line 573). This is documented behaviour. No issue.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+  if[debug;logwarnproc each 0!procs];
+  .z.m.hb:hb upsert select sym,procname,warning:1b from procs;
+  safecall[`checkheartbeat;`onwarning;onwarning;procs];
+  };
+
+err:{[procs]
+  / move processes into error state, log the transition and fire the error callback.
+  / NB warning is deliberately NOT cleared here, so an escalated process carries warning:1b AND
+  / error:1b. both flags are literally true - it is past both thresholds - and this matches legacy
+  / (heartbeat.q:71,77 set each flag independently and never clear). only a fresh heartbeat clears
```
DI-Software-Engineering, line 575: `storeheartbeat` updates `hb` with `select by sym,procname from rows`, which keeps only the latest row per `(sym,procname)` pair from the incoming batch. However `update warning:0b,error:0b from select by sym,procname from rows` applies the update before the upsert into `hb`. The upsert `hb upsert ...` upserts the selected-latest rows (with warning/error cleared) into the keyed table. This correctly clears warning/error only for processes that sent a heartbeat in this batch. A process already in `hb` with `warning:1b` that does NOT appear in the new batch keeps its warning state — that is the intended behaviour. No issue here.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,846 @@
...
+  / this is a reachable collision rather than a defensive hypothetical - di.subscriptions installs
+  / subscribed schemas at root, so a monitor watching a tickerplant that carries `heartbeat has one.
+  / there is deliberately no adoption path: the publish table is only ever an empty schema holder
+  / (rows go out over pub/sub, never into it), so a stale one from an earlier load costs nothing to
+  / remove, and "the columns matched" is far too weak a signal to hand over a name on
+  if[not tablename in tables[];
+    set[tablename;schema];
+    .z.m.rootowned:1b;
+    :()];
+  / already there and we created it on an earlier init - keep ownership and leave the table alone
```
DI-Software-Engineering: `validateroot` reads `rootowned` as a bare name via `@[{rootowned};::;0b]`. On a first-ever call before `init` has run, this falls back to `0b` correctly. But `ensureroottable` (called from `installroot`, which is called from `init`) also reads `rootowned` as a bare name (line 319: `if[rootowned;:()]`). After the first `init` sets `.z.m.rootowned:1b`, subsequent calls to `ensureroottable` read `rootowned` as a bare name. Whether this bare-name read resolves to the `.z.m` slot or the module-level name depends on the module loader's name-resolution. The comment in the project context states bare names in module code resolve against `.z.m` — if so, this is correct. If not, `rootowned` always reads the constant (which doesn't exist as a top-level name, so it would error). Given the project convention, this is likely intentional, but the inconsistency between `@[{rootowned};::;0b]` (protected, with fallback) and the bare `if[rootowned;:()]` (unprotected) means the first call before init will throw rather than return `0b` if the module loader does NOT alias bare names to `.z.m`. The `validateroot` protected form at least handles it safely.

di/heartbeat/test.q
```diff
@@ -0,0 +1,94 @@
...
+/   1. the remote-subscribe handshake, which depends on .z.w resolving to the MONITOR's connection
+/      inside an inbound call on the PUBLISHER
+/   2. the root schema table, without which di.pubsub silently refuses the subscription and
+/      discards every published row
+
+/ pick a free port by binding one and immediately releasing it
+freeport:{[]
+  p:0;
+  while[not p;
+    c:10000+rand 40000;
```
DI-Software-Engineering: `freeport` uses `do`/`while` which the project style guide explicitly prohibits. The guide states "Do not use the `do`, `while`, and `for` functions." The function uses `while` to find a free port. This should use a converge/fixed-point iterator instead. (Same issue at line 47 in the same file.)

di/heartbeat/test.q
```diff
@@ -0,0 +1,94 @@
...
+  path:"/tmp/dihbchild",string[port],".q";
+  writechild[hsym `$path;port];
+  system"q ",path," -p ",string[port]," -q </dev/null >/tmp/dihbchild",string[port],".log 2>&1 &";
+  h:0N;
+  i:0;
+  while[(null h) and i<100;
+    h:@[{hopen `$":localhost:",string x};port;0N];
+    i+:1;
+    system"sleep 0.1"];
+  if[null h;'"di.heartbeat test: publisher process failed to start - see /tmp/dihbchild",string[port],".log"];
```
DI-Software-Engineering, line 60: `spawnpublisher` uses two `while` loops (polling for the child process to start and for its init to complete), which the project style guide explicitly prohibits. These should be rewritten using iterators.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **7 warning(s)** | **0 suggestion(s)**

di/heartbeat/heartbeat.q
alowrydi: This is a false positive, on both premises.

The trailing type suffix on a space-separated numeric literal governs the whole vector, not just the last element - so -5 4 0Ni is a uniform int vector, never a mixed long/int list:

q)type -5 4 0Ni        / 6h
q)type -5 4 0Ni,0i     / 6h
q)-5 4 0Ni,0i          / -5 4 0N 0i

And the second premise doesn't hold either: q promotes across numeric types rather than signalling, so even a genuinely mixed list wouldn't throw here.

q)0i>=-5               / 1b
q)h where not (null h) or 0i>=h    / ,4i  - exactly what the test asserts

That row also runs on every invocation of a suite that is green (481/481), which a real 'type at this line couldn't coexist with.

No change.

di/heartbeat/heartbeat.q
alowrydi: Verified on a live session: a bare read inside module code resolves to the current .z.m value, and follows a re-init that changes it.

Initialising with warningtolerance:7f / errortolerance:9f against the 30s publishinterval:

warningperiod[`rdb]  ->  0D00:03:30      (7 × 30s)
errorperiod[`rdb]    ->  0D00:04:30      (9 × 30s)

There is also no stale value available to fall back to: configdefaults (heartbeat.q:47) is a dictionary, not a set of top-level names, so warningtolerance has no module-level constant behind it.

And init itself bare-reads publishroot at heartbeat.q:817, two lines after writing .z.m.publishroot - if bare reads didn't alias .z.m, every init call would throw.

This is now written down once at heartbeat.q:64-69 and asserted in test.csv (re-init changes the threshold; the read follows).

On the recomputation point: warningperiod each t\symis deliberate. The unusedprocesstype parameter is the per-type extension point described in the function comment, and the store holds one row per monitored process - a handful, not a table worth optimising.

di/heartbeat/heartbeat.q
alowrydi: Same mechanism as the warningperiod thread - bare reads track .z.m live.

Re-initialising with subscribewarnsweeps:9 makes emptysweepmsg[] report 9, never the shipped 3; both are asserted in test.csv now.

The key correction is the trackunseen precedent this cites. That hoist is not a .z.m workaround - its own comment says so (heartbeat.q:391-392), as does checkheartbeat's (heartbeat.q:608-609): a qsql select/where/by clause cannot resolve a module-level name at all.

Line 393 (n:subscribewarnsweeps) is itself a bare read that works fine; the local exists only so the later where sweeps=n resolves.

So the two cases are unrelated, and emptysweepmsg needs no hoist. Also worth noting configdefaults is a dict - the suggested failure mode ("resolves to the configdefaults list entry (3)") has no mechanism.

I swept every qsql statement in the module for an un-hoisted module name, which is the real bug class here, and found none.

di/heartbeat/heartbeat.q
alowrydi: Good catch on the asymmetry, and the premise is right - an unset bare module name genuinely throws rather than returning a null:

ensureroottable[] before init, foreign `heartbeat table at root  ->  'rootowned

It is unreachable through any supported path, though: init calls validateroot at heartbeat.q:775 before any state is written, which raises this module's own error for the foreign-table case and aborts; the first init writes .z.m.rootowned:0b at :788, before installroot runs at :817; and ensureroottable is internal, not exported.

Hardened anyway to match validateroot, since it costs one line and strictly improves the only path that can reach it - a direct call now gets di.heartbeat: installroot: a table named heartbeat already exists at root… instead of a bare 'rootowned.

Covered by a new test.csv row.

di/heartbeat/test.q
alowrydi: Accepted - rewritten as a while-iterator:

freeport:{[]
  p:{0=x}{@[{[c] system"p ",string c;c};10000+rand 40000;0]}/0;
  system"p 0";
  :p;
  };

I checked for precedent before changing it rather than assuming either way: main has zero while, and the only do[ is inside di/k4unit/k4unit.q, which is vendored KX k4unit (tab-indented, not written to this guide) - not precedent for new code.

di/servers does use while for this same OS-polling shape, in its fixture and in servers.q:251, but that's #120 and unmerged, so it isn't precedent either. With style.md:124 unqualified, the rule stands.

qlint also flagged that the original outer [x] was declared, unused, and shadowed by the inner lambda's implicit x - the same hazard fixed in subscribeone on #109 - so the inner parameter is now explicitly named c.

di/heartbeat/test.q
alowrydi: Accepted - both polls are now named helpers using the same while-iterator form, with the 100 × 0.1s caps unchanged:

awaithandle:{[port]
  s:{(null x 1) and x[0]<100}{[port;x]
    system"sleep 0.1";
    (x[0]+1;@[{hopen `$":localhost:",string x};port;0Ni])}[port]/(0;0Ni);
  :s 1;
  };

Pulled out of spawnpublisher rather than inlined, since style.md prefers named functions over longer anonymous lambdas.

The state is (attempts;handle), seeded 0Ni rather than 0N so it doesn't change type on the first iteration (hopen returns an int).

Verified the timeout branch still terminates rather than spinning: against a dead port it returns 0Ni after ~10.7s, so spawnpublisher's existing null-handle throw still fires.

Integration suite green at 70/70.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,858 @@
...
+  pid:`int$();
+  host:`symbol$();
+  port:`int$()
+  );
+
+/ keyed store of the latest received heartbeat per process, with warning / error state
+storeschema:update warning:0b,error:0b from `sym`procname xkey schema;
+
+/ keyed count of consecutive sweeps a discovered peer has gone without producing a heartbeat
+unseenschema:2!([]sym:`symbol$();procname:`symbol$();sweeps:`long$());
```
ConorSwainDI, line 32: Purely aesthetic - any particular reason why `2!` is used, and not surrounding the `sym` and `procname` columns with the square brackets to key it?

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,858 @@
...
+  if[not initialised[];
+    '"di.heartbeat: ",string[ctx],": init must be called before any other function"];
+  };
+
+tosecs:{[span]
+  / di.timer mode-2 periods are in whole seconds; legacy expressed these intervals as timespans
+  :`int$span%0D00:00:01;
+  };
+
+warningperiod:{[processtype]
```
ConorSwainDI, line 101: For `warningperiod` and `errorperiod` functions, how is the `processtype` used in the function? It doesn't look like it is being used at all in the logic.

di/heartbeat/heartbeat.q
```diff
@@ -0,0 +1,858 @@
...
+  if[99h<>type deps`pubsub;
+    '"di.heartbeat: pubsub value must be a dict (see di.pubsub)"];
+  if[not all `publish`subscribe in key deps`pubsub;
+    '"di.heartbeat: pubsub dict must expose `publish and `subscribe (see di.pubsub)"];
+  if[not all `proctype`procname in key deps;
+    '"di.heartbeat: proctype and procname (self-identity) are required in deps - they are published ",
+      "as the heartbeat's sym and procname and have no sensible default"];
+  if[not all -11h=type each deps`proctype`procname;
+    '"di.heartbeat: proctype and procname must be symbols"];
+  if[any null deps`proctype`procname;
```
ConorSwainDI, line 180: Trim this error message

PR #123, opened 2026-08-13
Feature subscriptions

alowrydi:
**Subscription management module, di.subscriptions**

Extracts `code/common/subscriptions.q` from TorQ and packages it as a standalone kdb-x module. Manages tickerplant subscriptions for RDB/WDB/chained-TP consumers - subscribe with optional exactly-once log replay, resubscribe after a dropped connection, unsubscribe, and a live registry of what's currently held.

The modularisation plan places `di.subscriptions` in the FRAMEWORK tier with hard dependencies on `di.servers` and `di.pubsub`. `getsubscriptionhandles` (the legacy handle-resolution function TorQ consumers call before subscribing) is ported here rather than duplicated across every future `di.proc.*` module, and a new opt-in `republish` capability lets a chained/segmented tickerplant consumer hand its subscribed tables to `di.pubsub` for downstream fan-out - the actual job a chained TP exists to do. Closing the `di.pubsub` edge safely required three small, necessary changes to that module, included here rather than as a follow-up: its `.z.pc` was overwriting rather than chaining, meaning depending on it as shipped would silently destroy any other module's observer; it shipped no `VERSION`, so a version pin against it would have been unenforceable; and it had no read counterpart to `setsubtables`, which the additive handoff needs. A fourth, unrelated `di.pubsub` fix (a broken result-shape guard in `subscribestr`/`subscribestrfilter`) is bundled in as well - found during this due-diligence read, not required by anything here, worth knowing that distinction if this gets split later.

Trello ticket - https://trello.com/c/aUtlpkfA/133-kdb-x-subscriptions-module

---

**Files created / modified**

| File | Status | Description |
| --- | --- | --- |
| `di/subscriptions/init.q` | new | Loads di.servers/di.pubsub deps, loads subscriptions.q, defines export of 9 functions |
| `di/subscriptions/subscriptions.q` | new | Full implementation - module state, internal helpers, public API, init |
| `di/subscriptions/subscriptions.md` | new | Full module documentation |
| `di/subscriptions/deps.q` | new | Declares di.servers >=0.1.0 and di.pubsub >=0.2.0 as hard dependencies |
| `di/subscriptions/test.csv` | new | 534 k4unit unit assertions |
| `di/subscriptions/test_integration.csv` | new | 131 k4unit integration assertions, including the three-real-process republish path |
| `di/pubsub/pubsub.q` | modified | `.z.pc` now chains via di.handlers-style capture instead of overwriting; getsubtables added; subscribestr/subscribestrfilter's result-shape guard fixed |
| `di/pubsub/init.q` | modified | Exports getsubtables, version |
| `di/pubsub/VERSION` | new | 0.2.0 |
| `di/pubsub/test.csv` | modified | 119 assertions; portable - no hardcoded port or path |

---

**How to test**

```q
k4unit:use`di.k4unit
k4unit.moduletest`di.subscriptions
```
```
===== di.subscriptions : test.csv =====
2026.08.13T00:53:08.477 start
2026.08.13T00:53:08.477 :.../di/subscriptions/test.csv 534 test(s)
2026.08.13T00:53:08.781 end
assertions : 534
failures   : 0
by action  : `s#`fail`run`true!70 250 214

===== di.subscriptions : test_integration.csv =====
2026.08.13T00:53:08.818 start
2026.08.13T00:53:08.818 :.../di/subscriptions/test_integration.csv 131 test(s)
2026.08.13T00:53:14.390 end
assertions : 131
failures   : 0
by action  : `s#`run`true!70 61

===== di.pubsub : test.csv =====
2026.08.13T00:53:14.418 start
2026.08.13T00:53:14.418 :.../di/pubsub/test.csv 119 test(s)
2026.08.13T00:53:15.550 end
assertions : 119
failures   : 0
by action  : `s#`fail`run`true!2 62 55
```
784 suite assertions, 0 failures. Plus 317 smoke/adversarial-harness assertions across 12 scratchpad suites (`smoke.q`/`smoke2.q`/`smoke3.q`, `adversarial.q`, `iter3.q`, `probe7`/`probe8`/`probe10`/`probe11.q`, `gsh.q`, `repub.q`, `pssmoke.q`) - 1101 assertions total, 0 failures, 0 flags.

The integration suite runs three real processes for the republish path specifically - an upstream tickerplant, a middle process using this module with `republish` on, and a downstream subscriber - since a subscriber that serves its own subscribers can't be tested in-process. Its tickerplant peer runs the real `di.pubsub`, and assertions read that module's own registry directly, so the coupling can't regress into a stand-in.

The handoff-failure warn-not-throw path was proven non-vacuous, not just exercised: the same suite run against a copy with the fix reverted to `raiseerror` produces 7 failures (the module throws, the completed subscription is lost, nothing is logged at `warn`); with the fix in place, 0.

`di.depcheck`:
```
2026.08.12D23:50:20.082403699 INFO [depcheck] dependency check complete: 0 failure(s), 0 warning(s)
depcheck: PASS
```
Resolved against `deps.q = `di.servers`di.pubsub!("0.1.0";"0.2.0")`, with `di.servers` at 0.1.0 and `di.pubsub` at 0.2.0 on disk.

---

**Design decisions**

*Protocol correctness (subscribe/replay against a real segmented tickerplant)*

*1. Guards that don't need the tickerplant's reply run before it* - `fetchdetails`'s underlying `subdetails` call is `.u.sub`, which registers the handle for live delivery as a side effect of just fetching schemas. Every check not needing that reply (root-`upd` presence, the already-held-table guard, the setschema-implies-exists guard) now runs first; only the log-integrity preflight is irreducible, since log filenames only exist in the reply itself.

*2. `` ` `` (all tables) is resolved via a `tablelist` round trip before `subdetails`* - a segmented tickerplant's `subdetails` can't accept the bare sentinel directly (throws `'rank`); legacy always resolved a concrete list first. Falls back to sending `` ` `` directly if no usable `tablelist` is offered.

*3. A shared log (segmented `singular`/`periodic` multilog mode) collapses to one whole-file replay* - the same physical file is reported once per table with different counts; replaying each entry separately re-applies the file's head. Collapsing to one whole-file replay trades a bounded, diagnosable trailing duplicate for never silently under-replaying, which a naive per-table sum could do.

*4. A whole-file replay is narrowed to the offered tables; a genuine per-table log is not* - a segmented tickerplant logs more tables than it publishes, so a whole-file read (from the shared-log collapse or the `0Wj` sentinel) can carry a table the caller never subscribed to. Narrowed via the existing table filter; the common per-table-log fast path is unaffected.

*5. `0Wj` means "replay everything", not "more than expected"* - resolved to the log's own readable total rather than forwarded to `-11!`, which would otherwise replay a corrupt log's good prefix and only then throw - the exact half-populated state the preflight exists to prevent.

*Framework-tier dependencies and republish*

*6. `getsubscriptionhandles` ported rather than left to each caller* - real TorQ callers across `rdb.q`, `wdb.q`, `chainedtp.q`, `sctp.q` all resolve a handle the same way before subscribing; centralising it here avoids four future `di.proc.*` modules each reimplementing legacy's `` ` ``/`()`/intersection-vs-union selection semantics independently.

*7. `republish` is opt-in, default off* - realises the chained/segmented-tickerplant role (subscribe upstream, serve the same tables downstream) that `chainedtp.q`/`sctp.q` split across themselves in TorQ. Off by default: a plain RDB must not silently become a publisher.

*8. The handoff to `di.pubsub` is additive* - unions with whatever the publisher already serves rather than replacing it, because `setsubtables` replaces its list and `unsubscribe` deletes registry rows; a registry-only union would drop an unsubscribed table at the next unrelated subscribe. Never shrinks toward empty either, since an empty list makes `di.pubsub` publish every root table.

*9. A handoff failure warns, doesn't throw* - it's the last step of `subscribe`, by which point the subscription has already fully succeeded (schemas defined, log replayed, registry committed). Throwing would discard completed work over an optional secondary step.

*10. `di.pubsub`'s `.z.pc` now chains* - a bare `.z.pc:{closesub[x]}` at load silently destroyed any observer another module had already registered, with `di.handlers` continuing to report it as live. Depending on `di.pubsub` as a hard dependency without fixing this would import that risk.

---

**Checklist**
- [ ] 534/534 unit + 131/131 integration assertions passing (di.subscriptions)
- [ ] 119/119 assertions passing (di.pubsub)
- [ ] di.depcheck clean
- [ ] Follows consistency.md and style.md
- [ ] subscriptions.md documents all exported functions, config, usage examples and notes
- [ ] di.servers PR #120 and di.permissions-equivalent version conventions confirmed compatible

---

**Documentation**

See `subscriptions.md` for full reference, including the `subdetails`/`tablelist` protocol notes, the shared-log and whole-file replay behaviour, `republish` configuration, and known non-blocking limitations (live-feed filtering delegation is unverified end-to-end pending `di.tickerplant`; a shared log with exactly-equal per-table counts is indistinguishable from an ordinary reply and silently under-replays - both carried forward as requirements for `di.tickerplant` to address upstream).


di/pubsub/pubsub.q
```diff
@@ -71,8 +71,19 @@ closesub:{[h]
...
+/ define .z.pc, add bespoke actions as needed.
+/ CHAINS onto whatever already owns .z.pc rather than replacing it. a bare .z.pc:{closesub[x]} here
+/ silently destroyed every observer another module had already registered - measured: with a
+/ di.handlers registration in place first, loading this module stopped it firing while di.handlers
+/ went on listing it as registered, so the failure was invisible from the registry.
+/ this stays a raw assignment rather than a di.handlers registration because the modularisation
+/ plan classifies di.pubsub as STANDALONE - it takes no injected dependencies, so it cannot reach
+/ di.handlers without contradicting its own tier
+priorpc:@[value;`.z.pc;{[e] (::)}];
+.z.pc:{[w]
```
DI-Software-Engineering, line 78: `getsubtables` always throws and never returns the table list: the trap `@[{[x] t};::;{[e] `symbol$()}]` passes `::` as the argument `x` to the lambda `{[x] t}`, but `t` is a bare name that — inside module code — is rewritten to `.z.m.t` at load. The argument `x` is never used, so the lambda always tries to read `.z.m.t`. If `.z.m.t` exists this works, but if the module is loaded in a context where the compile-time rewrite resolves differently, or if `t` has not been set yet, the outer lambda throws and the trap always returns `\`symbol$()`. More critically, the argument to `@[f;x;g]` is `::`, so `f` is called as `f[::]` — which is `{[x] t}[::]` — not as `{[x] t}[]`. For a zero-argument read the correct form is `@[{[] .z.m.t};::;{[e] \`symbol$()}]` (or simply reference `.z.m.t` directly inside the trap).

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,998 @@
...
+  / is `, so this costs the whole-file path a table-membership test per message and nothing else
+  raw:alltabs and (allsyms syms) and not wholefile;
+  n:$[raw;replayall[nmsg;lf];replaynarrowed[nmsg;lf;wanted;syms]];
+  .z.m.loginfo[`replay;"replayed ",(string n)," message(s) from ",string lf];
+  };
+
+replaylogs:{[entries;wanted;syms;alltabs]
+  / replay every pre-subscription log the tickerplant reported. entries are the PREFLIGHTED triples
+  / from preflightlogs - shared logs already collapsed to one replay each, any 0W already resolved
+  if[0=count entries;
```
DI-Software-Engineering, line 597: `resubscribe` removes dead rows only for the tables it successfully re-established (`done`), but the deletion predicate is `not active, f'[tabs]` — i.e. it deletes every inactive row whose full `tabs` list is a subset of `done`. A subscription to multiple tables (e.g. `` `trade`quote ``) is only deleted when *both* are in `done`; if one of them failed and only the other was re-established, the row is kept permanently, and every subsequent call to `resubscribe` will re-attempt the failed table and warn about it — exactly the timer-noise the comment above says was fixed. The deletion should match on individual table membership or on the handle, not on the full `tabs` list.

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,998 @@
...
+    raiseerror[`replay;"cannot read log ",(string lf),": ",(last r),registerednote]];
+  i:last r;
+  :$[1<count i;(first i;1b);(i;0b)];
+  };
+
+tryreplay:{[nmsg;lf]
+  / -11!(n;logfile) under protected apply, returning (1b;count) or (0b;error). n is the message count
+  / the tickerplant had logged when we subscribed, so messages that arrive after that - which also
+  / come down the live feed - are not replayed as well
+  :@[{(1b;-11!(x 0;x 1))};(nmsg;lf);{[e] (0b;e)}];
```
DI-Software-Engineering, line 432: `requiretablesexist` checks `wanted where not wanted in tables[\`.]`, but at the call site on line ~565 (`if[not alltabs; if[not setschema; requiretablesexist[requested]]]`) it is called with `requested`, which is the tablelist-resolved list — potentially larger than what `schemalist` will actually return. The comment on the surrounding code explicitly acknowledges this for the `setschema:0b` post-reply call against `wanted` (line ~575), but the pre-reply call still uses `requested`. For an explicit request this means a table in `tablelist` but absent from `schemalist` (a valid case the code later handles with `warnmissing`) will cause a spurious early failure even though `setschema:0b` replay would have succeeded. The pre-reply call should use `tabs` (the caller's explicit list, before narrowing), not `requested`.

di/subscriptions/test_integration.csv
```diff
@@ -0,0 +1,230 @@
...
+run,0,0,q,"(`$"":"",BASE,""/process.csv"") 0: (""host,port,proctype,procname"";""localhost,"",string[PEERPORT],"",tickerplant,peertp"")",1,1,write the strict 4-column phone book di.servers reads pointing at this run's live peer
+run,0,0,q,"svc.init[`log`timer`handlers`proctype`procname`connections`processcsv!(caplog;mocktimer;handlerdep[];`rdb;`rdb1;enlist`tickerplant;BASE,""/process.csv"")]",1,1,init di.servers as an rdb configured to connect to tickerplants
+run,0,0,q,svc.startup[],1,1,dial the peer over real IPC
+true,0,0,q,1=count svc.getservers[`tickerplant],1,1,di.servers holds a live connection to the peer
+run,0,0,q,sub.init[deps[]],1,1,di.subscriptions needs no servers wiring - di.servers is a hard dependency it imports itself; this re-init just refreshes log and handlers
+run,0,0,q,SH:sub.getsubscriptionhandles[`tickerplant;()],1,1,the rdb.q:163 call shape against a genuinely populated registry
+true,0,0,q,1=count SH,1,1,exactly one tickerplant resolved
+true,0,0,q,(enlist`peertp)~exec procname from SH,1,1,and it is the peer this suite spawned
+true,0,0,q,`procname`proctype`w~cols SH,1,1,projected to the three columns a subscribe caller needs
```
DI-Software-Engineering, line 196: The middle-process readiness check `1b~MH"MIDREADY"` races with the upstream `sub.subscribe` inside the middle process. The middle script reports its OS port and *then* runs `sub.subscribe`, which opens a socket to the peer and replays the log. `readport` returns as soon as the port file exists — before `MIDREADY` is set — so `midhandle[]` and the `MIDREADY` check can run while the subscribe is still in flight. If the peer or the log is slow, `MIDREADY` will be `0b` or undefined and the assertion fails spuriously. A `while` poll on `MIDREADY` (like `readport` uses for the port file) would close the window.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **4 warning(s)** | **0 suggestion(s)**

di/pubsub/pubsub.q
alowrydi: The bug claim doesn't hold. getsubtables[] returns `symbol$() before init and `quote`trade after. t is a bare read of .z.m.t, which init sets at pubsub.q:158 - so the trap's fallback is the pre-init case, not the normal one, and the empty return there is intentional and documented in the comment directly above.

One correction though:
@[{[] .z.m.t};::;g] works fine - a niladic lambda has rank 1, so it applies under @ without a rank error. The current form wasn't avoiding one.

Taking the explicitness half of the suggestion regardless: switched to an explicit .z.m.t read. qlint agrees - the old {[x] t} raised both UNUSED_PARAM and UNDECLARED_VAR, the new form raises neither, and it matches the convention already noted at subscriptions.q:694.

di/subscriptions/subscriptions.q
alowrydi: The per-row wrapper is all-or-nothing: subscribe is a single call with the row's full table list, so under the old code done got either all of it or none of it. An intra-row partial failure can't arise that way.

What can happen is narrowtabs (:383) succeeding with fewer tables - the tickerplant has retired a table. The row then looked fully successful and the row was deleted - silently losing the drop.

Registry-visible: subscribe stores wanted, not the request (:819).

The cross-row version of this symptom is reachable though (row `trade`quote fails, row `trade succeeds, first row retried and changed).

Fix: done now comes from subscribe's actual subtables (:924, :928), and dead removed, row dropped only when empty (:945).

Keeping the row whole, as you suggest, the call just created. Scoped by registry index to the rows the call actually attempted, f'[tabs].

New tests cover the full life cycle; they produce 9 failures against the pre-fix code.

di/subscriptions/subscriptions.q
alowrydi: The call site is guarded by if[not alltabs (:781 pre-fix), and on that branch requested is the tablelist-resolved list only. The premise that requested is the tablelist-resolved list only holds on the all-tables path, which that guard excludes.

The real problem is the early check being strictly stricter than the late one. It ran against the caller's raw list while the late copy run and schemalist, so a table narrowtabs was about to drop could throw early under `set subscription`.

Fix: the check now runs after narrowtabs, against sendtabs (:794), still ahead of fetchd.

Narrowing is pure and registers nothing, so the registration window stays closed. The underlying point about tablelist ⊃ schemalist is valid; the copy at :822 still covers. sendtabs is the closest approximation to wanted obtainable before the reply, not an equal one.

Test added: [test description was truncated in the source] and unpublished - throws pre-fix, passes post-fix.

di/subscriptions/test_integration.csv
alowrydi: Confirmed and fixed. Added a waitready poll fixture mirroring readport and replaced the one-shot check (:209).

Worth noting the failure mode was worse than a spurious 0b: MIDREADY is undefined until the subscribe returns, so the sync call errors rather than returning false - the fixture traps that.

Verified in isolation: returns on the 3rd poll in ~107ms when readiness is delayed, respects its timeout on a handle that never becomes ready, and the suite ran clean 3× in a row.

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,1026 @@
...
+  / never needed to hold. this runs AFTER narrowtabs but still BEFORE fetchdetails, which is what
+  / matters: the tablelist round trip behind `published` is pure and registers nothing, whereas
+  / fetchdetails registers this handle for live delivery as a side effect
+  / sendtabs is the closest approximation to `wanted` obtainable before the reply, not an equal one -
+  / it can still be a superset when schemalist omits a table the tablelist advertised, which is the
+  / gap the post-reply copy below continues to cover
+  if[replay;
+    if[not alltabs;
+      if[not setschema;requiretablesexist[sendtabs]]]];
+  details:fetchdetails[tph;sendtabs;syms];
```
DI-Software-Engineering, line 716: `sendtabs` is now computed before the `if[replay; if[not alltabs; if[not setschema; requiretablesexist[sendtabs]]]]` guard, but it is also used further down as the argument to `fetchdetails`. The old code computed `sendtabs` *after* the guard, meaning a non-replay, non-alltabs, non-setschema path that failed `requiretablesexist` would throw before `sendtabs` was ever used. That ordering is preserved by the restructuring only because `requiretablesexist` still throws. However, `narrowtabs` (called inside the `sendtabs` assignment) calls `fetchpublished` / uses `published`, which was obtained from the tickerplant via a `tablelist` round-trip. If `narrowtabs` itself can throw (e.g. the tickerplant is unreachable at that point), the error is now raised *before* the `requirerootupd` and `guardduplicate` guards that previously ran first. Those guards are pure/local and were intentionally placed ahead of any tickerplant contact. Moving `sendtabs` above them changes the order of side-effects on failure: a missing root `upd` now surfaces as a network error rather than a clear `requirerootupd` error. This is a latent ordering regression on the error path.

DI-Software-Engineering:
## DIReview Summary

di/subscriptions/subscriptions.q
alowrydi: sendtabs hasn't moved above those guards - it moved past requiretablesexist only. requirerootupd (:771) and guardduplicate (:772) run before it in both versions; sendtabs went from :785 to :775, and requiretablesexist from :782 to :794.

narrowtabs is also pure - except/inter, a warn, and a raiseerror. There's no fetchpublished in the module. It receives published as a value; the round trip that produced it is publishedtabs at :757, which already precedes both guards, deliberately and unchanged (see the comment on that line). So those guards were never ahead of tickerplant contact - the tablelist call is pure and registers nothing, which is exactly why it's safe there.

The described symptom can't arise for a second reason: publishedtabs traps and returns `, and sendtabs's $[alltabs or published~\;…] then skips the narrowtabs branch, so an unreachable tickerplant never reaches narrowtabs at all.

Measured on the exact case - dead handle, replay:1b, no root upd - the error is `replay was requested but no upd is defined at root`, not a connection error.

di/pubsub/test.csv
```diff
@@ -113,5 +147,22 @@ run,0,0,q,res:h"select from .m.di.0pubsub.reqfilteredtbl where table=`quote",1,,
...
+run,0,0,q,.test.EOPRAN:0b,1,1,did the subscriber body actually run
+run,0,0,q,.test.EOPGOT:`seeded,1,1,and what did it receive
+run,0,0,q,endofperiod:{[c;n;dat] .test.EOPRAN:1b; .test.EOPGOT:(c;n;dat); },1,1,a TERNARY subscriber - the shape TorQ's rdb and wdb both use
+run,0,0,q,h(`.m.di.0pubsub.callendofperiod;09:30:00;10:00:00;`a`b!1 2),1,1,broadcast a period roll from the server
+run,0,0,q,h"1+1",1,1,round trip so the async broadcast is processed before we assert
+true,0,0,q,.test.EOPRAN,1,1,the subscriber BODY ran - this is the assertion that fails on the unary version
+true,0,0,q,.test.EOPGOT~(09:30:00;10:00:00;`a`b!1 2),1,1,all three arguments arrived intact
+run,0,0,q,.test.EOPRAN:0b,1,1,reset before the one-argument case
+run,0,0,q,h(`.m.di.0pubsub.callendofperiod;09:30:00),1,1,the old one-argument calling shape
+run,0,0,q,h"1+1",1,1,round trip
```
DI-Software-Engineering, line 180: The test asserts that a one-argument call to a ternary `callendofperiod` "broadcasts nothing" and `not .test.EOPRAN` is true. But `callendofperiod:{[currentperiod;nextperiod;data]...}` called with one argument returns a projection — on the server side. That projection is then passed to `(neg getallhandles[])@\:`, which broadcasts the projection value (a lambda/projection) to all subscribers. The subscriber `endofperiod` receives this projected function, not a `(`endofperiod;...)` list, so it does not match the dispatch pattern and `.test.EOPRAN` stays false. The test assertion is correct by accident: the body does not run, but only because the broadcast payload is a projection object rather than a `(`endofperiod;c;n;d)` list, not because "nothing is broadcast". The comment "broadcasts nothing" is misleading and could cause future misdiagnosis. More importantly, if `.z.m.callendofperiod` or any wrapper validates arity before broadcasting, this contract changes silently. The test should assert the intended behaviour explicitly, or the comment should be corrected to describe what actually happens.

DI-Software-Engineering:
## DIReview Summary

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,1034 @@
...
+    raiseerror[`getsubscriptionhandles;"a general-list argument must be empty: use () to match nothing, ` to match everything"]];
+  / hoisted OUT of the where clause below: q-sql resolves procname to the COLUMN, so comparing
+  / against the parameter of the same name inside the select would silently compare the column
+  / with itself and match every row
+  pn:(),procname;
+  / di.servers refuses every accessor until its own init has run. that is deliberate on its side - a
+  / pre-init getservers used to return an empty table, indistinguishable from "nothing is connected" -
+  / but the raw signal names only di.servers and bypasses THIS module's log, so it is caught and
+  / re-raised through raiseerror: the caller learns which of the two modules is unwired, and the
+  / failure is observable in the log like every other domain error here. an empty result is NOT an
```
DI-Software-Engineering, line 883: The trap `@[{[] :servers.getservers[`]};::;handler]` catches ALL errors from `servers.getservers[\`]`, not just the "init not yet called" signal. Any transient or unrelated error from inside `getservers` (e.g. a type error, a connection failure within that function) will be caught, wrapped with the "di.servers must be initialised" message, and re-raised — misleading the caller about the true cause. Use a more targeted check, for example trap the error string and only rewrite it when it matches the known init-guard message, re-raising the original signal otherwise.

DI-Software-Engineering:
## DIReview Summary

di/subscriptions/VERSION
```diff
@@ -0,0 +1 @@
+0.1.0
```
DI-Software-Engineering, line 1: The version has been decremented from 0.2.0 to 0.1.0, which is a version regression. If this module has already been published or consumed at 0.2.0, downgrading the version number will break dependency resolution and version tracking for consumers. Restore to 0.2.0 (or increment to 0.3.0 / 1.0.0 etc.) rather than rolling back the version string.

DI-Software-Engineering:
## DIReview Summary

di/subscriptions/VERSION
alowrydi: Has never been merged to main so version remains 0.1.0

di/pubsub/pubsub.q
```diff
@@ -103,25 +125,79 @@ pubclear:{[t]
...
+  / stayed correctly chained, but pubsub's own hook stayed invisible to the registry itself.
+  / the raw chaining block keeps running regardless of whether handlers is passed here - it fires at
+  / module LOAD time, before init is ever called, so it cannot be conditionally skipped from inside
+  / init. that is harmless, not a double-registration bug: closesub is idempotent under a repeat call
+  / (delhandle/delhandlef are both remove-if-present), so a disconnect invoking it via both the raw
+  / chain and a handlers dispatch in the same tick is a no-op on the second call
+  if[not (::)~deps;
+    if[99h<>type deps;'"di.pubsub: deps, if given, must be a dict"];
+    if[`handlers in key deps;
+      if[not iscallable deps[`handlers]`register;
```
DI-Software-Engineering, line 168: The `registered` guard prevents re-registering with `di.handlers` on repeated `init` calls, but `.z.m.registered` is set to `1b` inside the `if[not registered;…]` block after a successful registration. On a second `init` call with a different `handlers` dep, the guard fires correctly. However, if the first `init` call provides a `handlers` dep and a second call omits it (i.e. `deps` is `(::)`), `registered` remains `1b` from the first call — there is no deregistration path. The module then silently retains an association with the first `handlers` instance and has no way to update or remove it. This is an unhandled lifecycle case: if `init` is documented as re-callable (the original `init[]` was), callers who reinitialize without `handlers` will be surprised to find the old registration persists with no way to clear it.

di/pubsub/pubsub.q
```diff
@@ -103,25 +125,79 @@ pubclear:{[t]
...
 setsubtables`;
 
 initialized:0b;
+registered:0b;
+
+iscallable:{[x]
+  / internal - is x a genuinely callable value? 100 112h spans every callable form, but 101h - the
+  / generic null :: - sits INSIDE that range while being callable in no useful sense, and :: is
+  / exactly what a dep dict hands back for a missing key. see di.servers' identical helper
+  t:type x;
```
DI-Software-Engineering, line 150: `iscallable` checks `type x` but `x` here is `deps[\`handlers]\`register` — a two-step index into a dict of dicts. If `deps[\`handlers]` is itself `(::)` (i.e. the `handlers` key exists in `deps` but maps to the generic null), the second index `(::)\`register` will signal a type error rather than returning `(::)`, so the `iscallable` guard is never reached and the error message from the type check is less helpful than the explicit message the guard would produce. Guard the outer value first: check that `deps[\`handlers]` is itself a dict (type 99h) before indexing into it for `register`.

di/pubsub/test.csv
```diff
@@ -51,9 +58,45 @@ run,0,0,q,res:h"select from .m.di.0pubsub.reqfilteredtbl where table in `trade`q
...
+run,0,0,q,"regres:system (getenv[`QHOME]),""/bin/q "",regdir,""/probe2.q -q""",1,1,run it in a genuinely separate process
```
DI-Software-Engineering, line 75: The probe script seeds a fake handle by setting `` `.m.di.0pubsub.reqalldict `` directly — a hard-coded internal implementation path that will silently break if the variable name or namespace ever changes, causing `BEFORE` to be `0b` and making the `closesub`-ran assertion (`before=1`) permanently fail without any diagnostic. The test should instead call the public subscribe API to create a real entry, so it tests the actual observable behaviour rather than an internal namespace.

DI-Software-Engineering:
## DIReview Summary

di/pubsub/test.csv
```diff
@@ -51,9 +58,53 @@ run,0,0,q,res:h"select from .m.di.0pubsub.reqfilteredtbl where table in `trade`q
...
+comment,,,,,,,di.pubsub instance loaded directly in this runner process (not the spawned server's copy),
+comment,,,,,,,so re-initialising it here cannot disturb any row above or below
+run,0,0,q,localps:use`di.pubsub,1,1,a local di.pubsub instance for testing init's re-registration behaviour in isolation
+true,0,0,q,"cnt:0; mockh:enlist[`register]!enlist {[e;p;n;pri;f] cnt+:1;}; localps.init[enlist[`handlers]!enlist mockh]; localps.init[enlist[`handlers]!enlist mockh]; cnt~2",1,1,"register is invoked on BOTH init calls, not just the first - di.handlers.register is itself idempotent under a repeat call with the same name, so there is no reason to latch this locally"
+
+comment,,,,,,,the STRING entry points must SIGNAL when a subscribe matched nothing. they exist for non-kdb+ clients which cannot inspect a q result shape, so an error has to arrive as an error
+comment,,,,,,,the guard that used to do this could never fire: errmsg is built with `$ so it is a symbol, and last of either success shape is the schema list - never the 10h string it tested for
+fail,0,0,q,"h(`pubsub.subscribestr;""nosuchtable"";"""")",1,1,subscribestr signals on a table that does not exist
```
DI-Software-Engineering, line 89: The mock handler `{[e;p;n;pri;f] cnt+:1;}` uses `+:` which is a local increment — it increments a local copy of `cnt`, not the outer `cnt` variable. In q, `cnt+:1` inside a lambda is equivalent to `cnt:(cnt:0)+1` scoped to that lambda's local environment, not an update to the enclosing scope's `cnt`. The register-call counter will remain 0 after both `init` calls, so `cnt~2` will never be true and the test will always fail (or, if the test is currently passing, it is passing for the wrong reason). Fix by using a global: change `cnt:0` to `cnt::0` and `cnt+:1` to `cnt+::1`, or accumulate calls differently (e.g. capture args in a list with `cnt,:enlist(e;p;n;pri;f)` using global amend).

DI-Software-Engineering:
## DIReview Summary

di/pubsub/pubsub.q
alowrydi: Tested this directly against the real build: 
pubsub.init[enlist[\handlers]!enlist (::)]does not throw a type error - (::)acts as the identity function under indexing, so(::)`registerreturns the symbol ``register`` rather than signalling, and iscallable correctly rejects that symbol. The intended message fires as designed: "di.pubsub: handlers\register must be a function [...] - see di.handlers"`. Adding the outer dict-type guard would be defensive code for a failure mode that doesn't occur in this q build. No change made.

di/pubsub/test.csv
alowrydi: Tested directly: cnt reaches 2 after both init calls, and cnt~2 evaluates true. No change made.

di/pubsub/init.q
```diff
@@ -1,3 +1,13 @@
...
+/ which tables (optionally sym- or condition-filtered), and the fan-out that publishes to them
+
 \l ::pubsub.q
 
-export:([subscribe;subscribestr;subscribestrfilter;publish;setsubtables;callendofperiod;callendofday;closesub;pubclear;init])
+/ module version, read from the VERSION file rather than hardcoded, so a release bump touches one
+/ plain-text file. read module-relative at load (`:::` resolves to di/pubsub) and BEFORE the export
+/ line, since export:([...]) evaluates each name. NB `version` must STAY in the export: di.depcheck
+/ resolves a dependency's minimum version from the export dict, and reports "exports no version" -
+/ failing the dependency check - for any module that omits it
```
ConorSwainDI: Purely cosmetic but perhaps reduce the rambling comment to something like
`/ set version string from VERSION file - fail is missing or empty`

di/pubsub/pubsub.q
```diff
@@ -71,14 +71,36 @@ closesub:{[h]
   delete from .z.M.reqfilteredtbl where handle=h;
   };
 
-/ define .z.pc, add bespoke actions as needed
-.z.pc:{closesub[x]};
+/ define .z.pc, add bespoke actions as needed.
```
ConorSwainDI: Does this comment need to be so wordy?

di/pubsub/pubsub.q
```diff
@@ -71,14 +71,36 @@ closesub:{[h]
...
 
 / broadcast to all subscribers upon end of day, client needs to define endofday function
 callendofday:{[d](neg getallhandles[])@\:(`endofday;d)};
 
-/ broadcast to all subscribers upon end of period, client needs to define endofperiod function
-callendofperiod:{(neg getallhandles[])@\:(`endofperiod;x)};
+/ broadcast to all subscribers upon end of period, client needs to define endofperiod function.
+/ TERNARY, matching legacy: TorQ's code/common/pubsub.q:19 broadcasts (`endofperiod;x;y;z) and both
+/ of its subscribers (code/rdb/endofperiod.q, code/wdb/writedown.q:52) are {[currp;nextp;data]}.
+/ it was unary, which failed two ways at once (both measured): callendofperiod[c;n;d] threw 'rank, so
```
ConorSwainDI: Again comment seems too wordy. In my view, it should just be explaining what the function does

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,1034 @@
...
+  / not, and the registry would then report a dead tickerplant as live for as long as .z.W held the
+  / handle number - and indefinitely once kdb+ reissued that number to another connection. that is
+  / precisely the failure the handlers dependency exists to prevent (see deps.q), so taking a new
+  / subscription in that state is refused rather than silently degraded to the .z.W-only mode this
+  / module documents as insufficient. reading and releasing stay available, for shutdown paths
+  / read explicitly, not as a bare name. a bare read does resolve to .z.m, but every other state
+  / access in this module is explicit, and the bare form is the one thing qlint flags as an
+  / undeclared global - a warning a reader has to dismiss by hand every time
+  if[not .z.m.observing;
+    raiseerror[ctx;"the .z.pc observer is not installed - teardown removed it, or init did not ",
```
ConorSwainDI: Seems like too much of an error message, can this be shortened at all?

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,1034 @@
...
+  / fails on a name it does not publish: standard SIGNALS 'x (u.q sub), while segmented and chained
+  / answer the pair (name;"Table ... not in list of stp pub/sub tables") which the schema guard then
+  / rejects. either way ONE bad name sinks the WHOLE call - and because the each-left runs left to
+  / right, every valid table ahead of it has already been registered by suball before the failure.
+  / a typo or a table retired at the tickerplant would otherwise take down the entire subscription
+  / and leave a partial one live at the tickerplant with nothing on this side recording it
+  dropped:(),tabs except published;
+  if[0=count dropped; :tabs];
+  .z.m.logwarn[`subscribe;"tickerplant does not publish ",(", " sv string dropped)," - dropping ",
+    $[1=count dropped;"it";"them"]," from the request, because asking for a table a tickerplant ",
```
ConorSwainDI: Another message that could maybe do with being shortened

di/subscriptions/subscriptions.q
```diff
@@ -0,0 +1,1034 @@
...
+  / the rest. a silently missed row is invisible and permanent; a duplicated row is visible and
+  / diagnosable, and this module rejects rather than silently narrows everywhere else.
+  / NB a file that appears ONCE keeps its own count - only a shared one is marked
+  if[0=count entries; :entries];
+  fs:entries[;1];
+  if[(count distinct fs)=count fs; :entries];
+  g:group fs;
+  .z.m.logwarn[`replay;"tickerplant reported ",(", " sv string where 1<count each g)," more than ",
+    "once - a shared-log (singular or periodic multilog) tickerplant. replaying each in full; ",
+    "messages logged between the subdetails call and the replay may arrive again on the live feed"];
```
ConorSwainDI: Can this message be split into smaller entries?

ConorSwainDI:
Functionally looks okay, but I do think some of the log messages and indeed comments need to be slimmed down a bit.

Could do with moving comments into the function definition, as per convention

ConorSwainDI:
Happy with changes. Resolve conflicts and I'll approve

PR #125, opened 2026-08-18
Feature tickerplant

ascottDI:
# di.tickerplant

> Ships together with `di.tplog`. This module depends on the new `di.tplog` (injected-log `init` and
> 1-arg `check`); its test suite exercises the two against each other, so the two modules should be
> reviewed and merged as a pair.

## Summary

Adds `di.tickerplant`, the core tick-capture process for the modular TorQ world: it receives updates
from feeds, stamps them, writes them to a tickerplant log for recovery, and publishes them to
subscribers, rolling the log at end of day. It is the modular replacement for TorQ's
`code/processes/tickerplant.q`.

It orchestrates three hard dependencies — `di.pubsub` (subscribe/publish), `di.eodtime` (roll timing)
and `di.tplog` (log check/repair) — with an injected logger and timer.

## Motivation

TorQ's `tickerplant.q` hand-rolls the whole capture loop — `.z.*` handling, the `.u` publish/subscribe
machinery, log writing and rolling, and end-of-day scheduling — in one process file. The modular
framework already provides those pieces as standalone modules (`di.pubsub`, `di.eodtime`, `di.tplog`,
`di.timer`), so this extraction re-expresses the tickerplant as an orchestration layer over them:
capture, stamp, log, publish, roll — delegating pub/sub, roll timing and log recovery to the modules
that own them.

### Root tables and the upd contract

The captured tables live at **root**, not in `.z.m`: a tickerplant owns its tables, feeds insert into
them, and `di.pubsub` reads them by name, so they cannot be module-local. This is the one deliberate
root-state exception; all other mutable state is module-local. `di.torq` wires the process's root
`upd` to `tickerplant.upd` so feeds publish to it.

### Capture modes

- **Batch** (default) — `upd` inserts into the root table; a timer job publishes the accumulated rows
  every `batchperiod` and clears them.
- **Zero-latency** — `upd` publishes each update immediately and does not buffer.

Both modes stamp the update (prepending a data timestamp unless one is already present) and log every
message.

### End of day

The roll fires when the current time passes `di.eodtime`'s next roll timestamp, checked on every `upd`
and every timer tick. `endofday` flushes the buffer, notifies subscribers, rolls the log to the next
day, and refreshes the roll time and data-timestamp offset from `di.eodtime`.

### Use of `di.tplog`

`di.tplog` is used only for check/repair on recovery: when `openlog` finds a pre-existing log it runs
it through `tplog.check`, repairing a corrupt one. The tickerplant opens for **append** and rolls the
log itself, rather than using `di.tplog`'s `open`/`roll`, because those replay the log through `upd` —
which a tickerplant must not do to its own log. (This module was updated to the new `di.tplog`
contract: it now calls `tplog.init` during its own `init`, and `tplog.check` is 1-arg.)

### No `di.handlers` dependency

Subscriber-disconnect cleanup is handled by `di.pubsub`'s own `.z.pc`, so this module takes no handler
dependency. `di.pubsub` should migrate to `di.handlers` so `.z.*` is not assigned outside the central
registry — tracked separately, out of scope here.

### Dependencies

| Dependency | Kind | Description |
|---|---|---|
| `di.pubsub` | hard (`use`) | subscribe / publish / roll notifications |
| `di.eodtime` | hard (`use`) | trading date, roll scheduling, data-timestamp offset |
| `di.tplog` | hard (`use`) | log check/repair on recovery |
| `log` | injected | `` `info`warn`error `` dict of `{[ctx;msg]}` functions |
| `timer` | injected | `di.timer`'s exports (must expose `addjob`) |

Hard deps are declared in `deps.q`; the injected `log` and `timer` are validated by `di.depcheck`'s
core-contract check, not declared there. `init` validates every dependency strictly and signals
immediately if any is missing or malformed.

## Public API

| Function | Signature | Description |
|---|---|---|
| `init` | `[deps]` | Wire log + timer, init the dep modules, materialise the schemas as root tables (`` `g# `` on `sym`), open today's log, and schedule the batch/roll timer job. Idempotent. |
| `upd` | `[table;data]` | Feed entry point: stamp, then buffer+log (batch) or publish+log (zero-latency). |
| `subscribe` | `[tables;filters]` | Register a subscriber (delegates to `di.pubsub`). |
| `endofday` | `[]` | Flush, notify subscribers, roll the log, advance the end-of-day state. |
| `getcounts` | `[]` | `` `i`j`d `` — messages published, messages logged, trading date. |
| `gettables` | `[]` | The tables this tickerplant captures. |
| `version` | — | Module version (`"0.1.0"`). |

`upd` validates its arguments and routes failures through a log-then-signal helper. `getapimeta[]`
exposes the callable API for central registration with `di.api`; the framework plumbing (`init`,
`getapimeta`, `version`) is intentionally excluded.

`init` also accepts optional config keys: `batch` (`1b`), `batchperiod` (timespan), `logdir` (string,
`""` disables logging), `logname` (prefix, default `"tp"`), `subtables` (symbol list), and the
`di.eodtime` keys (`rolltimezone` / `datatimezone` / `rolltimeoffset`) forwarded verbatim.

## Testing

`test.csv` / `test.q` (k4unit), 18 checks, run against the **real** `di.pubsub`, `di.eodtime`,
`di.tplog` and `di.timer` — no dependencies are mocked. The timer is used without `init` (so no live
`.z.ts`; its job is exercised through `endofday`), and a capturing logger is shared across the modules
so their output is assertable.

Coverage: the metadata/version contract; strict `init` dependency validation (a `fail` row per guard);
init materialising the root tables and scheduling the timer job; batch and zero-latency `upd`;
`endofday` flushing and rolling; `upd` input validation; and the two `di.tplog` integration points — a
tickerplant-written log replaying through `di.tplog`, and rolling into a corrupt log repairing it via
`tplog.check`.

```q
k4unit:use`di.k4unit
k4unit.moduletest`di.tickerplant
```

Because two tests replay/repair through `di.tplog`, the suite passes once `di.tplog` and
`di.tickerplant` are present together (as they merge) — it is not runnable against an older `di.tplog`.

## Files

```
di/tickerplant/init.q          entry point, version read, export list
di/tickerplant/tickerplant.q   implementation
di/tickerplant/tickerplant.md  module documentation
di/tickerplant/VERSION         module version (single source of truth)
di/tickerplant/deps.q          dependency manifest (di.pubsub, di.eodtime, di.tplog)
di/tickerplant/test.csv        k4unit tests (18)
di/tickerplant/test.q          test fixtures and helpers
```

The module version lives in a plain-text `VERSION` file, read in `init.q` (`version:trim first
read0\`:::VERSION`) and exported for `di.depcheck`, matching the convention used by `di.servers`,
`di.clienttracking` and `di.tplog`.


di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,198 @@
...
+  };
+
+stamp:{[x]
+  / prepend a data-timestamp column unless the update already carries one (first value is a timestamp)
+  if[-12h=type first first x;:x];
+  a:.z.p+eodtime.getdailyadj[];
+  :$[0>type first x;a,x;(enlist(count first x)#a),x];
+  };
+
+openlog:{[date]
```
DI-Software-Engineering, line 69: `stamp` checks `if[-12h=type first first x; :x]` to detect an already-stamped update. `first x` on a list of column vectors gives the first column (the time vector), and `first` of that gives its first element. But if the update is an atom row (0>type first x on the next line treats first-element type to detect an atom), `first x` is a scalar, and `first` of a scalar atom is the atom itself — so the check degenerates to `if[-12h=type x[0]; :x]` where x[0] is the first field value. If the feed sends a single-row atom update whose first field is a non-timestamp scalar, this guard still works correctly. However if x is an empty list (zero-row update), `first first x` is a null generic and type is 0h, not -12h, so the guard passes and stamp tries to prepend a timestamp to an empty update. Feeding an empty update is likely invalid input, but the `upd` function does not guard against it, so this path is reachable and will produce a malformed row inserted into the table.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,198 @@
...
+  };
+
+upd:{[t;x]
+  / feed entry point: stamp the update, then buffer+log (batch) or publish+log (zero-latency).
+  / t is the table name, x the column data. wired to root `upd` by di.torq so feeds can call it.
+  if[not -11h=type t;raiseerror[`upd;"table must be a symbol"]];
+  if[not t in .z.m.tabs;raiseerror[`upd;"unknown table ",string t]];
+  rollcheck .z.p;
+  x:stamp x;
+  if[.z.m.batch;t insert x;writelog[t;x]];
```
DI-Software-Engineering, line 322: `endofday` increments `.z.m.d` with `+:1` (adding the integer 1 to a date atom), then calls `rolllog[]` which calls `openlog .z.m.d` with the already-incremented date. The log filename therefore uses the new date. However `testendofday` in test.q asserts `not ()~key hsym`$dd,"/tp",string oldd+1` — i.e. it expects a file for `oldd+1`. This is consistent. But the sequence also means the log opened after roll is for `oldd+1`, while `eodtime.setnextroll` and `eodtime.setdailyadj` are called *after* `rolllog`. If `openlog` internally calls `tplog.check` (which uses `eodtime.getdailyadj` or roll state), it would see stale eodtime state. Whether `tplog.check` or `openlog` reads eodtime state cannot be confirmed from this diff alone, but the ordering — open log before refreshing eodtime state — is a latent sequencing risk.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,198 @@
...
+  .z.m.loginfo:(deps`log)`info;
+  .z.m.logwarn:(deps`log)`warn;
+  .z.m.logerr:(deps`log)`error;
+  .z.m.timer:deps`timer;
+  / optional config with defaults
+  .z.m.batch:$[`batch in key deps;deps`batch;1b];
+  .z.m.batchperiod:$[`batchperiod in key deps;deps`batchperiod;defaultbatchperiod];
+  .z.m.logdir:$[`logdir in key deps;deps`logdir;""];
+  .z.m.logname:$[`logname in key deps;deps`logname;"tp"];
+  / materialise the schemas as root tables (see the header note on root state), applying `g# to sym
```
DI-Software-Engineering: `init` guards against duplicate timer registration with `if[not `scheduled in key .z.m; ...]` and sets `.z.m.scheduled:1b`. However `.z.m.batch`, `.z.m.batchperiod`, and the timer job itself are set unconditionally on every `init` call. On a re-init the timer job is not re-added (correct), but `.z.m.logfile` is re-opened without closing the previous handle (line 127: `.z.m.logfile:openlog .z.m.d`). If `init` is called a second time the old file handle is orphaned — it is overwritten in `.z.m.logfile` with no `hclose`, leaking the OS file descriptor. `rolllog` closes `.z.m.logfile` before reopening, but `init` does not call `rolllog` or close the existing handle before calling `openlog`.

DI-Software-Engineering:
## DIReview Summary

di/tickerplant/tickerplant.q
ascottDI: added count=0 check

di/tickerplant/tickerplant.q
ascottDI: not a bug, no change

di/tickerplant/tickerplant.q
ascottDI: Fixed, close handle before opening new.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,201 @@
...
+    .z.m.timer[`addjob][`custom][`tickerplant;tick;();`int$.z.m.batchperiod%0D00:00:01;1h;()!()];
+    .z.m.scheduled:1b];
+  .z.m.loginfo[`init;"di.tickerplant initialised (",$[.z.m.batch;"batch";"zero-latency"]," mode)"];
+  };
+
+upd:{[t;x]
+  / feed entry point: stamp the update, then buffer+log (batch) or publish+log (zero-latency).
+  / t is the table name, x the column data. wired to root `upd` by di.torq so feeds can call it.
+  if[not -11h=type t;raiseerror[`upd;"table must be a symbol"]];
+  if[not t in .z.m.tabs;raiseerror[`upd;"unknown table ",string t]];
```
DI-Software-Engineering, line 317: The early-return `if[not count x;:()]` skips `rollcheck .z.p`, so a roll that is due will not be triggered if the first update of a new day happens to be empty. Move the empty-check after `rollcheck .z.p`, or document that empty updates are never expected when a roll is overdue.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,201 @@
...
+  if[not count x;:()];
+  rollcheck .z.p;
+  x:stamp x;
```
DI-Software-Engineering, line 321: The `hclose` guard checks `.z.m.logfile>0i` (integer comparison), but `hclose` expects a handle of type -7h (long) or -6h (int) depending on version, and `openlog` may return a long. If the handle is a long atom, `>0i` performs a cross-type comparison that always returns false on some q versions, meaning the old handle is never closed. Compare with `0` (long zero) or use `not null .z.m.logfile` combined with a type check, whichever matches the actual return type of `openlog`.

di/tickerplant/test.q
```diff
@@ -0,0 +1,123 @@
...
+/ re-init is safe: it closes the previous log handle instead of leaking the descriptor, and logging
+/ keeps working. fd count (linux /proc, as the suite is already unix-coupled) must not grow.
+testreinitnoleak:{[]
+  fddir:"/proc/",(string .z.i),"/fd";
+  freshinit["reinit";1b];
+  b:"J"$first system"ls ",fddir," | wc -l";
+  freshinit["reinit";1b]; freshinit["reinit";1b]; freshinit["reinit";1b];
+  a:"J"$first system"ls ",fddir," | wc -l";
+  tp[`upd][`trade;row`AAPL];
+  (a=b) and 1=tp[`getcounts][]`j};
```
DI-Software-Engineering: In `testreinitnoleak`, the fd count `b` is captured before the first `freshinit` but the comment says it is the baseline after one init. The baseline should be taken after `freshinit["reinit";1b]` (line 81) and before the three additional inits, so `b` already includes one open log handle. As written, `b` counts fds from whatever state preceded the test, and the comparison `a=b` will spuriously pass even if one extra fd leaks (because the baseline itself has one fewer handle open than the post-loop state). Move the `b` capture to after the first `freshinit` call.

DI-Software-Engineering:
## DIReview Summary

di/tickerplant/tickerplant.q
ascottDI: Correctly identified bug, moved guard to after `rollcheck`

di/tickerplant/tickerplant.q
ascottDI: Not a bug, no change warranted

di/tickerplant/test.q
ascottDI: Incorrect, no change

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,410 @@
...
+  / throws on a duplicate id, and teardown has deleted it. tick reads .z.m.batch live, so the one job
+  / serves both modes across re-inits
+  if[not .z.m.scheduled;
+    .z.m.timer[`addjob][`custom][`tickerplant;tick;();`int$.z.m.batchperiod%0D00:00:01;1h;()!()];
+    .z.m.scheduled:1b];
+  .z.m.loginfo[`init;"di.tickerplant initialised (",$[.z.m.batch;"batch";"zero-latency"]," mode)"];
+  };
+
+teardown:{[]
+  / release everything init installed process-wide: the root subscription protocol and the timer job.
```
DI-Software-Engineering, line 294: `pubsub.subscribe` returns a bare error symbol when no tables match, but the partial-match branch checks `type first r` — when `r` is a symbol atom, `first` of a symbol atom is the atom itself (not its first element), so `-11h=type first r` is `1b` for BOTH the all-fail case and a partial-match case where the first element of the returned pair happens to be a symbol. The all-fail guard (`-11h=type r`) is evaluated first, so the all-fail path is handled correctly, but if `pubsub.subscribe` ever returns a pair whose first element is a symbol atom (e.g. an error symbol paired with partial results), the partial-match branch would silently log and continue rather than raise. More concretely: after the all-fail guard the only shapes remaining are `(tables;schemas)` and `(errmsg;(tables;schemas))`. In the partial case `first r` is an error symbol atom, so `-11h=type first r` is correct — but the code relies on `first` of a 2-element list returning the first element, which is fine. The real bug is that `pairs:flip $[partial;last r;r]` uses `last r` when partial, but `pubsub.subscribe` on a partial match returns `(errmsg;(tables;schemas))` — so `last r` is `(tables;schemas)`, and `flip` of that is `(tables list; schemas list)`, which is correct. However when all tables match, `r` is `(tables;schemas)` directly, and `flip r` yields the same shape. This path is actually consistent. The real issue: in the all-none case the guard correctly signals, but for ZERO matched tables where `pubsub.subscribe` instead returned a `(tables;schemas)` pair with empty lists, `pairs[;0]` would be an empty list and `nms!0^.z.m.rowcounts nms` would be an empty dict — silently returning a dict that looks like success with no tables. Verify that `pubsub.subscribe` always returns a bare symbol (not an empty pair) when nothing matches.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,410 @@
...
+  .z.m.rowcounts:@[.z.m.rowcounts;t;+;count d];
+  .z.m.i:.z.m.j;
+  };
+
+writelog:{[t;x]
+  / append the update to the tp log (if enabled) and bump the total message count
+  if[.z.m.logfile>0i;.z.m.logfile enlist (`upd;t;x);.z.m.j+:1];
+  };
+
+logfilelist:{[]
```
DI-Software-Engineering, line 111: `logfilelist` checks `if[null .z.m.logpath;:()]` but `openlog` sets `.z.m.logpath:`` (the null symbol) when logging is disabled (`0=count .z.m.logdir`). The `null` check on a symbol atom is correct per q semantics. However on the very first call to `subdetails` before any `init` (i.e. before `openlog` has run), `.z.m.logpath` does not exist yet and the protected-apply in `initialised[]` guards against that for the public entry points — but `logfilelist` is an internal helper called from `subdetails` which itself calls `requireinit`. So the ordering is safe only because `requireinit` runs first. This is fine as written, but if `logfilelist` is ever called from a path that skips `requireinit`, it will throw `'.m.di.0tickerplant.logpath`. No fix needed unless the call graph changes, but worth noting the implicit coupling.

di/tickerplant/test.q
```diff
@@ -0,0 +1,264 @@
...
+/ init deliberately seeds runtime state (date, counts, open log) only on the first call - see
+/ tickerplant.md - so a suite that wants a genuinely clean tickerplant per test has to clear the name
+/ initialised[] probes. done here rather than by exporting a reset verb: a state-wiping function that
+/ exists only for tests does not belong in the public api. also drops the self-subscription a
+/ subdetails test leaves behind on handle 0, so a later flush does not publish into it
+resetmodule:{[]
+  if[`schemas in key `.m.di.0tickerplant;
+    tp[`teardown][];
+    @[{if[.m.di.0tickerplant.logfile>0i;hclose .m.di.0tickerplant.logfile]};::;{[e] :(::)}];
+    ![`.m.di.0tickerplant;();0b;`schemas`scheduled]];
```
DI-Software-Engineering, line 35: `resetmodule` calls `tp[\`teardown][]` and then directly manipulates `.m.di.0tickerplant` state with `@[{if[.m.di.0tickerplant.logfile>0i;hclose .m.di.0tickerplant.logfile]};::;{[e] :(::)}]` followed by `![\`.m.di.0tickerplant;();0b;\`schemas\`scheduled]`. But `teardown` already sets `.z.m.scheduled:0b` (which is `.m.di.0tickerplant.scheduled`), so the manual delete of `scheduled` is redundant. More critically, `teardown` does NOT close the log handle — it deliberately leaves module state intact per the design. The explicit `hclose` immediately after `teardown` is therefore the only place the log is closed during reset. If `teardown` throws (e.g. because it internally errors before the `scheduled:0b` write), the `hclose` is still attempted via the protected apply, which is correct. However, deleting `schemas` from `.m.di.0tickerplant` after a `teardown` that has already returned means `initialised[]` will correctly probe false — this is the intended mechanism. The sequence is fragile: if `tickerplant.q` ever adds a new name that `initialised[]` probes, `resetmodule` silently stops working. The test comment acknowledges this (`initialised[] probes`), but the fix — exporting a test-only reset or making `initialised[]` probe a dedicated flag — is documented as a deliberate decision. Not blocking, but the brittleness is real: adding a second probe variable to `initialised[]` in `tickerplant.q` would silently break the test suite.

di/tickerplant/tickerplant.q
```diff
@@ -0,0 +1,410 @@
...
+  if[0=count .z.m.logdir;.z.m.logpath:`;:0i];
+  l:hsym `$ .z.m.logdir,"/",.z.m.logname,string date;
+  if[count key l;l:tplog.check l];
+  if[not count key l;l set ()];
+  .z.m.logpath:l;
+  h:hopen l;
+  .z.m.loginfo[`openlog;"logging to ",string l];
+  :h;
+  };
+
```
DI-Software-Engineering, line 84: `publishbuffer` computes `.z.m.rowcounts:.z.m.rowcounts+.z.m.tabs!count each `. .z.m.tabs` before calling `pubsub.pubclear`, which publishes and then empties each table. If `pubsub.pubclear` throws (e.g. a subscriber handle is broken and the publish errors out internally), the row counts have already been incremented but the rows may not have been cleared from the buffer or fully published. On the next tick, `pubclear` would attempt to publish the same rows again and `rowcounts` would be incremented a second time, producing double-counted `rowcounts` and potentially duplicate delivery. The safer order is to increment `rowcounts` after `pubclear` succeeds. Consider wrapping the whole sequence or at minimum moving the `rowcounts` update after `pubclear`.

di/tickerplant/test_integration.csv
```diff
@@ -0,0 +1,137 @@
...
+before,0,0,q,"H1:spawnpeer[""tp1"";`tp1tok;1b]",1,1,"spawn a batch-mode tickerplant. this ONE handle carries the control calls, the subscription and the live feed - exactly as a real subscriber's does"
+before,0,0,q,"PID1:H1"".z.i""",1,1,"its pid, for the leak check"
+before,0,0,q,H1 (`feed;til 4),1,1,"four updates: logged, and buffered in the peer's root table"
+before,0,0,q,"PRE:H1""counts[]""",1,1,capture: counts before anything is published
+before,0,0,q,"H1""flushtp[]""",1,1,publish the buffer to nobody - this is what moves i up to j
+before,0,0,q,"MID:H1""counts[]""",1,1,capture: counts after the flush
+before,0,0,q,H1 (`feed;4 5),1,1,"two more updates: logged, still buffered, NOT yet published"
+before,0,0,q,"ATSUB:H1""counts[]""",1,1,capture: the counts a subscriber arriving now is answered from
+before,0,0,q,TPD:ATSUB`d,1,1,capture: the trading date the tickerplant reports
+before,0,0,q,R1:sub.subscribe[H1;`;`;1b;1b],1,1,"THE CALL UNDER TEST - ` for all tables, so tablelist is exercised too; define the schemas at root and replay the log"
```
DI-Software-Engineering, line 61: The `readport` helper uses a `while` loop with `system"sleep 0.05"` — a busy-wait that the project style guide explicitly prohibits (`do not use the do, while, and for functions`). While this is test/integration code rather than production code, it violates the documented project convention. Replace with a recursive converge or a timed `\` iteration, or at minimum acknowledge the exception in a comment.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **5 warning(s)** | **0 suggestion(s)**

di/tickerplant/test.q
```diff
@@ -0,0 +1,276 @@
...
+    and all `subdetails`tablelist in key `.};
+
+/ batch mode: upd stamps, buffers into the root table and logs (bumping j). i is the PUBLISHED
+/ watermark and must NOT move while the rows are still buffered - it catches up at the flush
+/ init's OPTIONAL `handlers dep is forwarded verbatim to di.pubsub.init - see the design note in
+/ tickerplant.md. a fresh init WITH a handlers dep must make di.pubsub's own .z.pc hook visible in
+/ di.handlers' own registry (di.handlers.list[`.z.pc] naming `pubsub), which omitting the key (every
+/ other test here) never triggers
+testhandlersforwarded:{[]
+  resetmodule[];
```
DI-Software-Engineering, line 84: The test function `testhandlersforwarded` never resets the module state it leaves behind (a `handlers.init` + `tp.init` with a `handlers` dep injected). All other test functions in this file call `freshinit` or `resetmodule[]` at entry AND rely on a clean state set up by the `before` rows. Because the test does not call `resetmodule[]` or `freshinit` at the end, the live `handlers` dep injected into `tp`/`pubsub` during this test will persist into subsequent tests, which were never designed to run with that dep present and may behave differently (e.g. any test that triggers `.z.pc` or calls `pubsub.init[]` without a `handlers` key will now reach a `pubsub.init` that was last called with one). Add a `resetmodule[]` / `freshinit` call after the assertion, or use a `.[;; resetmodule]` trap pattern to guarantee cleanup even on failure.

DI-Software-Engineering:
## DIReview Summary

PR #126, opened 2026-08-19
Feature parquet extract module

ConorSwainDI:
Converts an in-memory kdb+ table into one or more `.parquet` files via `kx.arrow`. Rows are grouped
by instrument and packed into files close to a configurable target size, splitting any single
oversized instrument across multiple files where required. A manifest recording what was written
(file, instruments, row count, time range, on-disk size) is accumulated in the module's private
`manifest` table


di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+
+  sizeondisk:hcount hsym `$testloc;
+  newratio:sizeondisk % -22!t;
+
+  / clean test file
+  .z.m.loginfo[`pqx;"Cleaning up test file"];
+  system "rm ",testloc;
+
+  / return new ratio
+  .z.m.loginfo[`pqx;"Returning calibrated compression ratio"];
```
DI-Software-Engineering, line 99: `system "rm ",testloc` deletes the calibration test file using an unsanitised string built from `o\`outdir` (user-supplied option). If `outdir` contains spaces or shell metacharacters the `rm` command will behave incorrectly or dangerously. Use a quoted path: `system "rm -- \"",testloc,"\""` or use `hdel` on the hsym instead of `system "rm"`.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+  / takes a kdb+ table of data, its name, the date and any option overrides and saves down to parquet format
+  .z.m.loginfo[`pqx;"Extracting ",string[tname]," data for ",string dt];
+  / override default opts with o where applicable
+  opts:default,o;
+
+  / check for instrument and time cols, error out if not
+  if[not opts[`symcol] in cols[t];
+    .z.m.logerr[`pqx;err:"di.pqx: no symcol found in table"];
+    'err
+  ];
```
DI-Software-Engineering, line 359: `system "mkdir -p ", writedir` passes an unsanitised user-supplied path directly to the shell. If `opts[\`outdir]`, `tname`, or `dt` contain spaces or shell metacharacters the command will misbehave. Quote the path: `system "mkdir -p \"",writedir,"\""`.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+ };
+
+writefile:{[t;o;writeopt;writedir;map]
+  / writes data to disk in Parquet format
+  / returns stats to be inserted into the manifest
+  syms:map[`syms];
+  seqno:map[`seqno];
+  estbytes:map[`estbytes];
+  symcol:o[`symcol];
+
```
DI-Software-Engineering, line 211: In `writefile`, the manifest row uses a hardcoded `time` column name in `first exec min time from t[i]` and `first exec max time from t[i]` instead of `o[\`timecol]`. If the caller overrides `timecol` (e.g. to `` `ts ``), the `mintime`/`maxtime` manifest columns will either error (column not found) or silently capture the wrong column. Use `o[\`timecol]` as the column reference.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+    oversized:select from t where islargerthantargetsize;
+    t:t except oversized;
+    oversized:update numfiles:ceiling[estbyt%maxsize] from oversized;
+    plans,:enlist each raze {[t;c] t[`numfiles]#enlist t[c]}[;o`symcol] each oversized
+  ];
+
+  / next fit function for packing instruments into buckets if they conform to the max size
+  if[count t;
+    .z.m.loginfo[`pqx;"Bucketing small instruments"];
+    tabs:t[o[`symcol]];
```
DI-Software-Engineering: `calcsize` builds a functional-form select with a literal backtick-quoted `` `syms `` symbol in `enlist(in;symcol;\`syms)` — this tests whether the `symcol` column is `in` the atom `` `syms ``, not in the list `syms` (the local argument). The intended filter is `where symcol in syms` (the argument), but the functional form encodes the symbol `` `syms `` literally. The correct form passes the value: `enlist(in;symcol;enlist syms)`. As written, the `?[tbl;…]` will match rows where the instrument column equals the symbol `` `syms `` rather than the instruments in the plan, so `estbyt` will always be 0 unless a real instrument happens to be named `` `syms ``.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+    plans,:value[tabs @ group bins]
+  ];
+  plans:(1 + til count plans)!plans;
+
+  / attach estbytes to plan's output
+  :update estbytes:.z.m.calcsize[symstats;o`symcol;;]'[syms;seqno] from {`syms`seqno!/: flip (key[x];value[x])} group plans
+ };
+
+datalookup:{[t;symcol;syms;cnt]
+  / get lists of indices by file
```
DI-Software-Engineering, line 172: `step` accumulates `(bucketIndex; runningSize)` using `scan` (`\`). The scan accumulates state across all `i` correctly, but `bins` is built as `(step\[(0;0);til n])[;0]` — indexing column 0 of the scan result. Because `step` returns a 2-element list `(bucketIndex; size)`, indexing `[;0]` on the scan result gives the bucket indices only when the scan result is a matrix. However, when `n` is 1, `step\[(0;0);enlist 0]` returns a single-element list of a 2-element list, not a 2-column matrix, so `[;0]` returns the whole inner list `(0;sz)` rather than just `0`. This causes `group bins` to fail or produce wrong output for a single non-oversized instrument. Fix: use `{x[;0]}` guarded with a type check, or reshape: `bins:(step\[(0;0);til n])[;0]` should be `first each (step\[(0;0);til n])`.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+
+calibrateratio:{[t;o;writeopt]
+  / writes a sample of data to disk and reads its size on disk
+  / calcaultes the compression ratio and returns if a new ratio was successfully calculated, otherwise old ratio is maintained
+
+  / remove leading : from outdir
+  testloc:$[":" ~ first string[o`outdir];
+    1_string[o`outdir],"/testWrite.parquet";
+    string[o`outdir],"/testWrite.parquet"];
+
```
DI-Software-Engineering, line 77: `medsym` is derived as the row of `cnts` closest to the median row count, but when `cnts` is empty (zero-row input table) `cnts[o[\`symcol]]` returns an empty vector, `med` of an empty list is null (`0n`), and `bytesperrow` becomes `%[…;0]` (division by zero → `0w`). The downstream `estimate` returns `0w` `compressionratio`-scaled estimates, causing `plan` and `writefile` to behave unexpectedly. The test suite marks a zero-row `extract` as `fail`, but the failure is expected to come from `kx.arrow` during the calibration write, not from a clean early error — if `calibrate:0b` is set, the zero-row case silently produces infinite estimates. Add an explicit early guard: `if[0=count t;'"di.pqx: cannot extract empty table"]` at the top of `extract`.

di/pqx/pqx.q
```diff
@@ -0,0 +1,283 @@
...
+  if[not `log in key deps;
+    '"di.pqx: log dependency is required; pass `info`warn`error functions keyed on `log"];
+  if[99h<>type deps`log;
+    '"di.pqx: log value must be a dict of `info`warn`error functions"];
+  if[not all (`info`warn`error) in key deps`log;
+    '"di.pqx: log dict must have `info`warn`error keys; got: ",(", " sv string key deps`log)];
+  .z.m.loginfo:deps[`log]`info;
+  .z.m.logwarn:deps[`log]`warn;
+  .z.m.logerr:deps[`log]`error;
+ };
```
DI-Software-Engineering: The file is missing a trailing newline. Many q loaders and POSIX tools expect text files to end with a newline; absence can cause the last line to be silently ignored or cause parse errors depending on the loader.

di/pqx/pqx.q
DI-Software-Engineering: In `calcsize`, the functional-form query `?[tbl;enlist(in;symcol;\`syms);0b;()]` passes the literal symbol `` `syms `` as the right-hand operand of `in` rather than the local variable `syms` (the list of instruments). This means the filter matches rows where the instrument column equals the atom `` `syms `` — an instrument name that will never exist — so `sum[…\`estbyt]` is always `0`, making every planned file's `estbytes:0`. Fix: pass the value of `syms` by wrapping it: `enlist(in;symcol;enlist syms)` (or use `enlist syms` as a constant in the parse tree).

DI-Software-Engineering:
## DIReview Summary

**3 critical** | **5 warning(s)** | **0 suggestion(s)**

di/pqx/pqx.q
```diff
@@ -0,0 +1,282 @@
...
+      .z.m.logwarn[`pqx;"seqNo ",string[seqno]," write unsuccessful. Error - ",last res]
+    ];
+    :`file`seq`syms`nsyms`rows`mintime`maxtime`estbytes`bytes`split`status!/: flip (
+      hsym `$path;
+      seqno;
+      enlist distinct[t[i][o`symcol]];
+      count[distinct[t[i][o`symcol]]];
+      count[i];
+      first ?[t[i];();();(min;o`timecol)];
+      first ?[t[i];();();(max;o`timecol)];
```
DI-Software-Engineering: The functional select `?[t[i];();();(min;o`timecol)]` applies a single aggregation and returns a dictionary keyed by the column name, not a scalar. `first` on that dictionary returns the value for the first key (which happens to be the aggregated value), but only because there is exactly one aggregation here. More critically, if `o`timecol` is not `time` (e.g. a differently named timestamp column), the previous hard-coded `exec min time from t[i]` would have always read the column named `time` regardless. The new code is correct in intent, but `first` on the result of a no-by functional select returns the first element of the result vector, not a scalar atom — confirm that the downstream consumer of `mintime`/`maxtime` in the result table handles a list value correctly if `t[i]` is a sub-table with multiple rows, because `?[t[i];();();(min;o`timecol)]` returns a single-element list `(min_val)`, not an atom, and the earlier `first exec …` was doing the same unwrapping — verify this is still correct after the change.

di/pqx/pqx.q
```diff
@@ -0,0 +1,282 @@
...
+  .z.m.loginfo[`pqx;"Cleaning up test file"];
+  system "rm ",testloc;
+
+  / return new ratio
+  .z.m.loginfo[`pqx;"Returning calibrated compression ratio"];
+  :newratio
+ };
+
+calcsize:{[tbl;symcol;syms;seqno]
+  / find the estimated size in bytes for each instrument per file to be saved down
```
DI-Software-Engineering: The fix wraps `syms` in `enlist` inside the functional where clause: `enlist(in;symcol;enlist syms)`. In a functional select, the `in` operator requires its right argument to be a list of values to test membership against. If `syms` is already a list (the typical case — a list of symbols), `enlist syms` turns it into a nested list `(syms;)`, which will cause a type error at runtime. The original `\`syms` (backtick-name reference) was likely intended to resolve the local variable; if `syms` is already a symbol list, the correct form is `enlist(in;symcol;syms)` without the extra `enlist`.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **2 warning(s)** | **0 suggestion(s)**

di/pqx/pqx.q
ConorSwainDI: This was indeed an oversight of mine, this has now been resolved

di/pqx/pqx.q
ConorSwainDI: This is correct, must have missed this first time around. This has been resolved

ConorSwainDI: Similar to above, resolved

di/pqx/pqx.q
ConorSwainDI: Having tested this out, I believe enlist syms is the correct logic here, as in my use cases removing this will throw an error (error looks like the name(s) of instruments).

di/pqx/pqx.q
ConorSwainDI: This has now been implemented

di/pqx/pqx.q
```diff
@@ -0,0 +1,288 @@
...
+      enlist distinct[t[i][o`symcol]];
+      count[distinct[t[i][o`symcol]]];
+      count[i];
+      ?[t[i];();();(min;o`timecol)];
+      ?[t[i];();();(max;o`timecol)];
+      estbytes;
+      @[hcount;hsym `$path;0];
+      split;
+      `error`ok[first res]
+    )
```
DI-Software-Engineering, line 236: Removing `first` from `?[t[i];();();(min;o`timecol)]` changes the return type from a scalar atom to a single-element list (the aggregation result of a functional select with no `by` clause returns a table-like structure or list). Downstream code that expected a scalar (e.g. for comparison, arithmetic, or writing metadata) will now receive a list, causing a type error or silent misbehaviour. Restore `first` unless callers have been updated to handle the list form.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
ConorSwainDI: Removed unnecessary logic

di/pqx/pqx.q
ConorSwainDI: Fixed to protect against this

di/pqx/pqx.q
ConorSwainDI: Resolved

di/pqx/pqx.q
ConorSwainDI: Resolved to leverage hdel hsym testloc

di/pqx/pqx.q
ConorSwainDI: I've tested this case out of a single non-oversized table and have not encountered the failure anticipated here

```
q)tabs:t[o[`symcol]];
q)sizes:t`estbyt;
q)n:count tabs;
q)n
1
q)step:{[maxsize;sizes;state;i] sz:sizes i; tot:state 1; $[(tot+sz)>maxsize; (1+state 0; sz); (state 0; tot+sz)]}[maxsize;sizes];
// below returns as normal
q)(step\[(0;0);til n])[;0]
,0
// repeating by explicitly stating enlist 0, again same return
q)step\[(0;0);enlist 0][;0]
,0
q)bins:(step\[(0;0);til n])[;0]
q)value[tabs @ group bins]
DOW
q) plans,:value[tabs @ group bins]
q) plans:(1 + til count plans)!plans;
q)plans
1| DOW
```

I'm planning to leave this as is, unless there is another reason why I shouldn't keep this.

di/pqx/pqx.q
```diff
@@ -0,0 +1,317 @@
...
+
+extract:{[t;tname;dt;o]
+  / main data extract function
+  / takes a kdb+ table of data, its name, the date and any option overrides and saves down to parquet format
+
+  / check for bad input keys
+  if[any not key[o] in key[default];
+    .z.m.logerr[`pqx;err:"di.pqx: input keys not recognised - ", "," sv string key[o] where not key[o] in key default];
+    'err 
+  ];
```
DI-Software-Engineering, line 335: The bad-key check reads `key[o]` (the raw user-supplied overrides), but `o` may be an empty dict `()!()` or even a general dict where `key` returns a symbol list. That is fine. However, the check `any not key[o] in key[default]` will throw a type error if `o` is passed as a non-dict (e.g. a general list), crashing before the user-friendly error message is produced. More importantly, if `o` is the null/empty dict the `any` result is `0b` and it passes silently — which is correct. The real issue is that `key[default]` is recomputed on every `extract` call from the module-level `default` dict; if `default` has been extended elsewhere before this check, a key that was not in the *original* `default` will not be caught. This is a fragile dependency on the mutable module-level `default`. Consider capturing `key[default]` at init time.

di/pqx/pqx.q
```diff
@@ -0,0 +1,317 @@
...
+
+  .z.m.loginfo[`pqx;"Extracting table: ",string[tname]," data for date: ",string dt];
+  / override default opts with o where applicable
+  opts:default,o;
+
+  / if one sym per file requested, turn off splitoversized
+  if[`onesymperfile in key o;
+    if[o`onesymperfile;
+      .z.m.loginfo[`pqx;"onesymperfile requested, turning off splitoversized"];
```
DI-Software-Engineering: `onesymperfile` override of `splitoversized` is applied to `opts` (the merged dict) only when `onesymperfile` is present in the *raw* `o` dict. If a caller passes `onesymperfile` as a default by having it set to `1b` in `default` (which it is not currently, but could be changed) and does *not* pass it in `o`, the guard `if[`onesymperfile in key o` ]` would not fire and `splitoversized` would never be forced off. The intent is clearly to force it off whenever `opts[`onesymperfile]` is true, regardless of whether it came from `o` or `default`. The condition should be `if[opts`onesymperfile; ...]` rather than checking `key o`.

di/pqx/pqx.q
```diff
@@ -0,0 +1,317 @@
...
+  res:raze parallelfn [.z.m.writefile[t;opts;writeopt;writedir]; plans];
+  .Q.gc[];
+
+  / write down manifest to partition - best-effort, does not abort the extract call if it fails
+  .z.m.loginfo[`pqx;"Writing down manifest file for table: ",string[tname],"; date: ",string dt];
+  manwrite:.z.m.tryfn[set;(hsym `$writedir,"manifest";res)];
+  if[not first manwrite;
+    .z.m.logwarn[`pqx;"di.pqx: error writing manifest to disk: ",last manwrite]
+  ];
+
```
DI-Software-Engineering, line 431: `tryfn` is called as `.z.m.tryfn[set;(hsym `$writedir,"manifest";res)]`. Looking at `tryfn`'s definition it wraps `.[f;x;{...}]`, so `x` must be the argument list for `f`. `set` is a binary operator taking `(path;data)`, so the argument list should be passed as a 2-element list. Here `(hsym `$writedir,"manifest";res)` is indeed a 2-element list, which is correct for `.[set;(path;data);handler]`. However `writedir` is built elsewhere as a string ending in `/`, so `writedir,"manifest"` concatenates to e.g. `"pqxout10/pqxtrade/date=2025.07.25/manifest"` — a relative path string cast to symbol with `` hsym `$ ``. If `extract` is called from a different working directory than the one `writedir` was constructed relative to, the sidecar will be written to the wrong location while the data files (which use the same `writedir`) will also be wrong. This is a pre-existing issue with path construction, but the new manifest write makes it observable in a new failure mode.

di/pqx/pqx.q
```diff
@@ -0,0 +1,317 @@
...
+        $[(tot+sz)>maxsize; (1+state 0; sz); (state 0; tot+sz)]
+      }[maxsize;sizes];
+      bins: (step\[(0;0);til n])[;0];
+
+      plans,:value[tabs @ group bins]
+      ]
+    ]
+  ];
+  plans:flip `seqno`syms!((1 + til count plans);plans);
+
```
DI-Software-Engineering, line 166: In the new `plan` function, the `onesymperfile` branch does `plans,:enlist each t[o[`symcol]]` which extracts the sym-column values from the `symstats` table and enlists each one. This produces a list of singleton sym lists (one symbol per bucket) — matching the shape the packing branch produces. However `symstats` at this point still holds the pre-filtered table (it may contain symbols that appear more than once in the sym column if `symstats` was not de-duped). If two rows share the same symbol, you get two separate buckets for the same symbol, producing two output files for one instrument. Verify that `symstats` is guaranteed to have exactly one row per symbol before reaching `plan`, or add a distinct/dedup here.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **4 warning(s)** | **0 suggestion(s)**

di/pqx/pqx.q
```diff
@@ -0,0 +1,414 @@
...
+buildvirtualtable:{[hdbdir;tname;datecol;dictcols]
+  / builds a queryable virtual table over parquet files scattered under hdbdir/tname, without reading any
+  / of their data - partition values (the date segment and any dictcols segments) are reconstructed from
+  / each file's hive-style path, mirroring how writefile strips those same columns from the on-disk data
+  path:` sv hdbdir,tname;
+  files:([] file:system"find \"",(1 _ string path),"\" -name \"*.parquet\"");
+  files:update split:"/" vs/:file from files;
+  lv:1+count where "/"=string path;
+  levels:(),datecol,dictcols;
```
DI-Software-Engineering: The `find` command is built by string-concatenating `1 _ string path` directly into a shell command without any escaping. If `hdbdir` or `tname` contains shell metacharacters (spaces, quotes, `$`, backticks, etc.), this will silently produce wrong results or execute unintended shell commands. Use a safe path construction instead, or at minimum document that path components must not contain shell metacharacters.

di/pqx/pqx.q
```diff
@@ -0,0 +1,414 @@
...
+  levelcols:{[datecol;x] (castvirtualcol[datecol;x;];`split)}[datecol] each lv+til count levels;
+  files:![files;();0b;(`file,levels)!enlist[({hsym `$x};`file)],levelcols];
+  pqt.mkP (levels#files)!pq.pq each exec file from files
```
DI-Software-Engineering: `lv` counts the number of `/` characters in `string path` (e.g. `string `:pqxout1/pqxtrade` → ":pqxout1/pqxtrade"), but `find` returns absolute paths while `path` is a relative hsym. The leading `:` (stripped by `1 _` in `find`) means the split-segment offsets computed from `path` do not match the actual depth of the `find` output paths, so `files[;lv+n]` will index the wrong path component and `castvirtualcol` will silently reconstruct the wrong values. The level-offset must be computed against the absolute path returned by `find`, not against the module-local relative path symbol.

di/pqx/pqx.q
```diff
@@ -0,0 +1,414 @@
...
+ };
+
+castvirtualcol:{[datecol;x;y]
+  / casts the xth hive-style path segment (a "key=value" string) of every row in y's split column to its
+  / reconstructed value - a date if it's the datecol level, else a symbol
+  :$[datecol = first `$distinct first each "=" vs' v:y[;x];
+    "D"$last each "=" vs' v;
```
DI-Software-Engineering: `castvirtualcol` inspects only the first row's key name (`first `$distinct first each "=" vs' v`) to decide whether the level is the datecol. If the directory happens to contain no files (empty `files` table), `v` is empty and `first` returns null, causing the comparison `datecol = first null` to produce a type error or wrong result. More critically, it assumes all files at that path depth share the same key name — if they don't (mixed directory layouts), only the first file's key name is consulted, silently miscasting the rest.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
ConorSwainDI: Path will be passed in as a full path, so I'm comfortable leaving this as is.

di/pqx/pqx.q
ConorSwainDI: symstats will not contain duped syms, so I don't believe this is an issue

di/pqx/pqx.q
ConorSwainDI: Path will be passed in as a full path, so it shouldn't be an issue

di/pqx/pqx.q
```diff
@@ -0,0 +1,415 @@
...
+    .z.m.logerr[`pqx;err:"di.pqx: input keys not recognised - ", "," sv string key[o] where not key[o] in key default];
+    'err 
+  ];
+
+  .z.m.loginfo[`pqx;"Extracting table: ",string[tname]," data for date: ",string dt];
+  / override default opts with o where applicable
+  opts:default,o;
+
+  / if one sym per file requested, turn off splitoversized
+  if[o`onesymperfile;
```
DI-Software-Engineering: Unguarded access `o\`onesymperfile` will throw a key error if `onesymperfile` is absent from `o`. The previous code guarded with `if[\`onesymperfile in key o;` before indexing. Removing that guard means any caller that omits `onesymperfile` from their options dict will get a runtime error. Fix: restore the outer guard, or use `$[\`onesymperfile in key o; o\`onesymperfile; 0b]` as the condition.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
```diff
@@ -0,0 +1,419 @@
...
+      .z.m.logwarn[`pqx;"seqNo ",string[map`seqno]," write unsuccessful. Error - ",last res]
+    ];
+    :`file`seq`virtualcols`syms`nsyms`rows`mintime`maxtime`estbytes`bytes`split`status!/: flip (
+      hsym `$path;
+      map`seqno;
+      enlist o`virtualcols;
+      enlist[syms];
+      count[syms];
+      count[i];
+      ?[t[i];();();(min;o`timecol)];
```
DI-Software-Engineering, line 230: The `virtualcols` value is wrapped with `enlist` but `syms` (which previously occupied this position) is also wrapped with `enlist`. The `!/: flip (...)` pattern requires all rows to have the same length. `enlist o\`virtualcols` produces a 1-element list containing the virtualcols value, while `enlist[syms]` produces a 1-element list containing the syms list — this pairing is consistent only if every other scalar in the column list is also a 1-element list. Verify that `o\`virtualcols` is already a list (not an atom) and that `enlist` is the correct wrapper here; if `virtualcols` is itself a list (e.g. a list of column names), `enlist` is correct. But if it is already a nested list, `enlist` adds an extra layer inconsistent with how `syms` is stored in the same column, breaking the manifest schema where both `virtualcols` and `syms` are typed as `()` (generic list columns expected to hold lists-of-lists per row).

di/pqx/pqx.q
```diff
@@ -0,0 +1,419 @@
...
+  / check parallel flag to pick iterator function
+  parallelfn:(each;peach)[opts[`parallel]];
+
+  .z.m.loginfo[`pqx;"Writing down ",string[tname]," Parquet files to disk for ",string dt];
+  res:raze parallelfn [.z.m.writefile[t;opts;writeopt;writedir]; plans];
+  .Q.gc[];
+
+  / write down manifest to partition - best-effort, does not abort the extract call if it fails
+  .z.m.loginfo[`pqx;"Writing down manifest file for table: ",string[tname],"; date: ",string dt];
+  manwrite:.z.m.tryfn[0:;(hsym `$writedir,"manifest.json";enlist .j.j res)];
```
DI-Software-Engineering, line 427: The manifest is now written as JSON (`manifest.json` via `.j.j res`) but any existing code that reads the manifest back (e.g. a load path using `get hsym \`$writedir,"manifest"`) will look for the old binary-format file at the old path and silently find nothing or error. If there is a reader elsewhere that still uses the old path/format, it will break. Ensure all readers are updated in the same change.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **1 warning(s)** | **0 suggestion(s)**

di/pqx/pqx.q
ConorSwainDI: No issue with out of date readers

di/pqx/pqx.q
ConorSwainDI: Resolved

di/pqx/pqx.q
ConorSwainDI: Resolved

di/pqx/pqx.q
ConorSwainDI: Default should not be extended, and has indeed been hidden from export

di/pqx/pqx.q
ConorSwainDI: virtualcols is already a list, so the enlist will not be an issue

di/pqx/test.csv
```diff
@@ -0,0 +1,193 @@
...
+true,0,0,q,(`$first exec syms from pqxsidecar10)~first exec syms from pqxret10,1,,syms round-trips through JSON once each instrument name is cast back from string to symbol
+true,0,0,q,0=count first exec virtualcols from pqxsidecar10,1,,virtualcols is empty in the sidecar for a call made without virtualcols set
+
+run,0,0,q,pqxret11a:pqx.extract[pqxbasic;`pqxtrade;2025.07.27;enlist[`outdir]!enlist `:pqxout11/],1,,First extract call into a partition that will be written to twice
+run,0,0,q,pqxret11b:pqx.extract[pqxsingle;`pqxtrade;2025.07.27;enlist[`outdir]!enlist `:pqxout11/],1,,Second extract call into the same partition with different data
+run,0,0,q,pqxsidecar11:.j.k first read0 hsym `$"pqxout11/pqxtrade/date=2025.07.27/manifest.json",1,,Parse back the JSON manifest sidecar after the second call into the same partition
+true,0,0,q,("j"$exec rows from pqxsidecar11)~exec rows from pqxret11b,1,,The manifest.json sidecar reflects only the most recent call's stats
+true,0,0,q,not ("j"$exec rows from pqxsidecar11)~exec rows from pqxret11a,1,,Repeat extract calls into the same partition overwrite the sidecar rather than appending to it
```
DI-Software-Engineering, line 101: The `syms` round-trip test uses `first exec syms from pqxsidecar10` to get a single string, then casts it with `` `$ ``. But `syms` in the JSON sidecar is a list of instrument name strings (e.g. `["AAPL","MSFT"]`), so `exec syms from pqxsidecar10` returns a list of lists (one inner list of strings per row). `first` of that gives the inner list `("AAPL";"MSFT")`, and `` `$first exec syms from pqxsidecar10 `` correctly produces `` `AAPL`MSFT ``. However, the right-hand side `first exec syms from pqxret10` returns `first` of a column of symbol lists — a symbol list `` `AAPL`MSFT `` — so the comparison is `` `AAPL`MSFT ~ `AAPL`MSFT `` which is fine. This is actually correct for the single-row case, but if `pqxbasic` ever produces multiple rows the test would silently only compare the first row's syms. More critically, the test does NOT cover the case where `syms` in the JSON is an array of strings: `.j.k` will decode `["AAPL","MSFT"]` as a list of strings, so `` `$first exec syms from pqxsidecar10 `` works. This is fine as written for a single-row result, but the test description says 'each instrument name is cast back' yet only checks the first row's syms — fix by asserting on all rows or documenting the single-row assumption.

DI-Software-Engineering:
## DIReview Summary

di/pqx/test.csv
ConorSwainDI: The first is unwrapping the nested list within a single row, this has been documented in later change

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
...
+  pqt.mkP (levels#files)!pq.pq each exec file from files
+ };
+
+checkvirtuallevels:{[levels;lv;splits]
+  / validates that datecol/virtualcols (levels) match, in both count and name, the hive-style
+  / key=value directory segments actually present under every discovered file's path - so a
+  / declared datecol/virtualcols combination that doesn't match what's really on disk (e.g. a
+  / datecol directory that exists but wasn't declared, a missing/extra virtualcols level, or the
+  / wrong order) fails loudly here rather than silently mislabeling or dropping partition columns
+  depths:count each splits;
```
DI-Software-Engineering, line 296: `onDiskKeys` uses camelCase, violating the repo style guide which mandates all-lowercase names for all functions and variables. Rename to `ondiskkeys`.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **1 warning(s)** | **1 suggestion(s)**

### Suggestions
- **di/pqx/pqx.q:284** — The variable `onDiskKeys` is built with `distinct` over per-file extracted key lists, but when files agree on keys the result is a 1-element list whose sole element is itself a list of symbols. `first onDiskKeys` is then that symbol list, and the comparison `not levels~first onDiskKeys` works. However, when `splits` is a single-element list, `{[idx;x] \`$first each "=" vs' x idx}[idx] each splits` returns a list containing one symbol list, so `distinct` on a 1-item list returns that same 1-item list — this part is fine. The real problem is the earlier `depths:count each splits` line: `splits` is passed as `exec split from files`, which is a list of lists (each element is the split path segments for one file). `count each splits` gives the number of segments per file — correct. But `(first[depths]-1)-lv` computes the number of hive levels as `(segments_in_first_file - 1) - lv`. The `-1` is intended to exclude the filename itself, but `lv` is already `1 + count where "/"=string path`, where `path` is the base hive root (e.g. `` `:pqxout14/ ``). The split segments for a file path like `pqxout14/pqxtrade/date=2025.08.01/exch=N/f.parquet` will have 5 elements; `lv` for a single-level base path is 1, so `depth = (5-1)-1 = 3`, but only 2 levels (`date`, `exch`) are present — off by one. Fix: use `(first[depths]-1)-lv` only if `lv` counts exactly the non-level prefix segments, or rewrite as `(first[depths]-1) - lv` and verify `lv` is computed consistently with the number of prefix segments in `splits`.

di/pqx/pqx.q
ConorSwainDI: Resolved

di/pqx/init.q
```diff
@@ -0,0 +1,20 @@
...
+/ evaluates each name; version stays in the export so di.depcheck reads it from the export dict.
+/ trim, and fail LOUD on a missing/unreadable/empty VERSION, rather than a bare `first read0`:
+/ a raw OS error names no module, and an empty or whitespace-padded value is worse than an error -
+/ di.depcheck compares versions as STRINGS, so padding silently breaks the comparison and an empty
+/ value reads as "exports no version", failing every dependent module's check for a reason that
+/ points nowhere near the real cause.
+version:@[{trim first read0 x};`:::VERSION;{'"di.pqx: VERSION file missing or unreadable"}];
+if[0=count version;'"di.pqx: VERSION file is empty"];
+
+export:([init;extract;getmanifest;checkandconvertcols;estimate;plan;writefile;readfile;tryfn;buildvirtualtable;checkvirtuallevels;version])
```
DI-Software-Engineering: `castvirtualcol` is removed from the export list but no corresponding removal of its definition appears in the diff. If `castvirtualcol` is still defined in the module, callers who previously discovered it via `key pqx` will silently lose access to it without any deprecation notice or error. If it is genuinely being made internal-only, verify no external callers depend on the exported symbol.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
ConorSwainDI: The expected directory structure is stated in the documentation. Partitioning on anything other than date first is a quite extreme edge case, so I think it is fine to assume that this will always be the first level.

di/pqx/init.q
ConorSwainDI: No dependency on external calls

di/pqx/init.q
```diff
@@ -0,0 +1,20 @@
+/ KDB-X Parquet extract module to save kdb+ data to parquet storage convention
+
+arrow:use`kx.arrow
+pq:use`kx.pq
+pqt:use`kx.pq.t
+
+\l ::pqx.q
+
+/ module version, read from the VERSION file (one plain-text file to bump per release). read
```
jprgrant: I've had this issue with claude - it goes into storytelling module in code comments, far too verbose. Considering these lines will be repeated in every module, I think one line is enough here: "set version string from VERSION file - fail is missing or empty" or something like that

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **0 warning(s)** | **2 suggestion(s)**

### Suggestions
- **di/pqx/pqx.q:281** — The `checkvirtualtypes` function never returns a value on the success path. In q, a function with no explicit return falls off the end returning the last expression's value — here that is the result of `if[count bad;...]`, which on the success path is a generic null `(::)`. This is benign if callers ignore the return, but it is inconsistent with the repo style guide's instruction to use explicit return (`:`) and may surprise a caller who checks the result. Add `: ` (or `: ::`) at the end of the success path to make the intent explicit.
- **di/pqx/pqx.q:271** — `levelcols` is built with `.' flip (types;lv+til count names)`, which uses the Apply-Each (`.`) iterator over pairs of `(typ; index)`. But `castvirtuallevel` takes THREE arguments `[typ;x;y]` — the third argument `y` (the `files` table row) is not passed here; it is meant to be a projection (partial application) that will be applied later in the functional-update `![files;...]`. The projection `castvirtuallevel[typ;x;]` leaving `y` open is what `castvirtualcol` previously used. The new `.' flip (types;lv+til count names)` will call `castvirtuallevel[typ;x]` with only two args, producing a projection, not an error — so `levelcols` becomes a list of projections, which is structurally the same as before. However, verify that `.' flip (types;lv+til count names)` actually produces `castvirtuallevel[types[i];(lv+i);]` projections for each i, because `. ` (Apply) on a list of 2 elements calls `castvirtuallevel` with those 2 elements as positional args (i.e. `castvirtuallevel[typ;x]`), which IS a valid 2-arg partial application leaving `y` open. This appears correct — but the old code used `{[typ;x] (castvirtuallevel[typ;x;];`split)}` as a lambda explicitly forming the projection and pairing it with `` `split ``. The new code does `.' flip (types;lv+til count names)` which only produces the `castvirtuallevel[typ;x]` projections WITHOUT pairing them with `` `split ``. The `` `split `` column reference needed for the functional-update is missing from `levelcols`. In the functional update `![files;();0b;(`file,names)!enlist[...],levelcols]`, each element of `levelcols` should be a PARSE-TREE expression like `(castvirtuallevel[typ;x;];`split)`, not a bare projection. Bare projections are not valid parse-tree column expressions for `!` functional update — they must be wrapped as `(f;col)`. The old lambda `{[datecol;x] (castvirtualcol[datecol;x;];`split)}[datecol] each lv+til count levels` explicitly built `(projection;`split)` pairs; the new code drops the `` `split `` wrapping, breaking the functional update.

di/pqx/pqx.q
```diff
@@ -0,0 +1,462 @@
...
+  / parse-cast type chars (see castvirtuallevel)
+  valid:"BGXHIJEFCSPMDZNUVT";
+  bad:distinct types where not types in valid;
+  if[count bad;
+    .z.m.logerr[`pqx;err:"di.pqx: unsupported partition level type(s) ",.Q.s1[bad],", expected one of ",.Q.s1 valid];
+    'err
+  ];
+  :(::)
+
+ };
```
DI-Software-Engineering, line 288: Missing semicolon after `:(::)` — the line `:(::)` has no trailing `;` before the closing `};`, violating the repo style guide requirement that every line ends with a semicolon. While this may not cause a runtime error in itself, it is inconsistent and the blank line between `:(::)` and `};` is also irregular. More critically, `:(::)` as an explicit return is only needed here because the `if` block above no longer unconditionally throws; if execution reaches `:(::)` normally that is fine, but the missing `;` may cause a parse issue depending on what follows — add `;` at the end.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **1 warning(s)** | **1 suggestion(s)**

### Suggestions
- **di/pqx/pqx.q:305** — The `depth` and subsequent `if` checks are now reachable even when the earlier `if[1<count distinct depths;...]` branch throws — that is correct — but `depth:(first[depths]-1)-lv;` is computed AFTER the early-exit `if` block closes. However, when `count splits` is 0 (no files found), `depths` is an empty list, `first[depths]` returns a null (0N), and the subtraction proceeds silently producing a null rather than an error, then `if[depth<>count levels;...]` compares null to an int. This null-check guard was not present before and is not added here. Add a guard: `if[0=count splits; .z.m.logerr[...]; 'err]` before accessing `first depths`.

di/pqx/pqx.q
ConorSwainDI: Fixed

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
...
+  / return stats and (new) compression ratio
+  :(update estbyt:rowcnt*bytesperrow*o[`compressionratio] from cnts;o[`compressionratio])
+ };
+
+calibrateratio:{[t;o;writeopt]
+  / writes a sample of data to disk and reads its size on disk
+  / calculates the compression ratio and returns if a new ratio was successfully calculated, otherwise old ratio is maintained
+
+  / remove leading : from outdir
+  testloc:$[":" ~ first string[o`outdir];
```
jprgrant: Instead of the conditional and the duplicated code, you can just use the boolean result as the number of chars to drop - i.e. drop zero chars if not true:
testloc:(":" ~ first string[o`outdir])_ string[o`outdir],"/testWrite.parquet";

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
...
+    .z.m.loginfo[`pqx;"Enforcing one file per virtualcols combination"];
+    grp:0!?[t;();(o`virtualcols)!o`virtualcols;`syms`estbytes!((o`symcol);(sum;`estbyt))];
+    :update seqno:enlist each 1+til count grp, estbytes:enlist each "j"$estbytes from grp
+  ];
+
+  plans:();
+
+  / if one sym per file, enlist each sym to assign to individual buckets
+  / else move to oversized and packing logic
+  $[o`onesymperfile;
```
jprgrant: I find the structure of this conditional block a bit hard to read - this is partly just personal preference but I don't like having large multiline code blocks in $[;;] conditionals - I prefer just using 'if' blocks even if you have to make the else check explicit, e.g.
if[o onesymperfile;
  / code
  ];
if[not o onesymperfile;
  / else code
  ];

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
...
+  / above) and encoded in its path already, so t is stripped of them in place afterwards - once here, not
+  / duplicated into every row of the on-disk data, and not dropped later from an already row-indexed slice
+  / (which breaks the arrow write whenever a file holds more than one instrument). guarded on count: a
+  / functional delete of an EMPTY column list against a `p#-attributed symcol (set by presort, on by
+  / default) silently returns a zero-row table instead of being the no-op it should be
+  idx:.z.m.datalookup[t;o`symcol;syms;count map;combomask];
+  t:$[count o`virtualcols;![t;();0b;o`virtualcols];t];
+
+  / build paths for each seqno
+  paths:writedir,/:o[`filestub],/:"-",/:("0"^-5$string[map`seqno]),\:".parquet";
```
jprgrant: Can you explain the -5 here? I haven't worked with parquet file much before so maybe it is obvious, but why do we need to pad the seq to 5 chars, and do we know for sure we can never overflow 99999?

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
...
+      opts[`virtualcols]:enlist opts`virtualcols
+    ]
+  ];
+
+  / if presort, sort by sym then time
+  / apply parted attribute on symcol
+  if[opts`presort;
+    .z.m.loginfo[`pqx;"Sorting data"];
+    t:opts[`symcol`timecol] xasc t;
+    t:![t;();0b;(enlist opts[`symcol])!enlist(#;enlist `p;opts[`symcol])]
```
jprgrant: Is there a reason to do this as a functional update rather than an amend, i.e.
t:@[t;opts[\`symcol];\`p#]  
The only reason I can think is if the table might be keyed

di/pqx/pqx.q
```diff
@@ -0,0 +1,442 @@
+/ define default config
+default:(
```
jprgrant, line 2: For readability it would be better to have the keys and values on the same line. maybe like this:
```
default:enlist[\`]!enlist(::); / initialise with sentinel key to avoid type errors
default[\`targetsize]:  512*1024*1024; / ~512 MB target file size
default[\`maxfactor]: 1.5;  / hard cap = target * maxfactor
...
default:default  _ `; / drop empty key
```

jprgrant (commented):
One general comment - this is unix (linux/mac) only due to the system command usage (find & mkdir). I think that's probably fine but the docs should state it explicitly

di/pqx/pqx.q
ConorSwainDI: I believe there's no functional need for this, I can remove it.

di/pqx/pqx.q
ConorSwainDI: No reason, I can refactor to amend

di/pqx/pqx.q
```diff
@@ -0,0 +1,465 @@
...
+  / default) silently returns a zero-row table instead of being the no-op it should be
+  idx:.z.m.datalookup[t;o`symcol;syms;count map;combomask];
+  t:$[count o`virtualcols;![t;();0b;o`virtualcols];t];
+
+  / build paths for each seqno
+  paths:writedir,/:o[`filestub],/:"-",/:(string[map`seqno]),\:".parquet";
+
+  / data to write, iterated by file
+  res:raze {[t;o;writeopt;split;syms;map;path;i]
+    .z.m.loginfo[`pqx;"Writing seqNo ",string[map`seqno],", path - ",path];
```
DI-Software-Engineering, line 217: Removing the zero-padding (`"0"^-5$string[map`seqno]`) from sequence numbers changes the file naming format: files were previously named `foo-00001.parquet`, `foo-00002.parquet`, etc. (zero-padded to 5 digits), and are now named `foo-1.parquet`, `foo-2.parquet`, etc. This breaks any downstream tooling, manifest queries, or consumers that rely on the padded format. The test in test.csv is updated to match, confirming this is intentional, but it is a breaking change to the on-disk file naming convention that could silently corrupt or lose data if existing parquet stores are read by code expecting the old naming scheme.

di/pqx/pqx.q
```diff
@@ -0,0 +1,465 @@
...
+  if[o`onesymperfile;
+    .z.m.loginfo[`pqx;"Enforcing one sym per file"];
+    plans,:enlist each t[o[`symcol]]
+  ];
+    / if split oversized is required, check against maxsize and return a plan entry for each required file
+  if[not o`onesymperfile;
+    if[o`splitoversized;
+      .z.m.loginfo[`pqx;"Splitting large instruments"];
+      t:update islargerthantargetsize:estbyt>maxsize from t;
+      oversized:select from t where islargerthantargetsize;
```
DI-Software-Engineering, line 142: The refactoring from `$[onesymperfile; [...]; [...]]` to two separate `if` blocks breaks the mutual-exclusion logic. In the original code, the two branches were exclusive (either `onesymperfile` ran OR the `splitoversized`/packing logic ran). Now both `if[o`onesymperfile;...]` and `if[not o`onesymperfile;...]` can execute independently, but critically the `onesymperfile` block modifies `plans` via `plans,:enlist each t[o[`symcol]]` and then execution falls through to the `if[not o`onesymperfile;...]` block which is skipped correctly when `onesymperfile` is true. That part is fine. However, if `onesymperfile` is false, the second block runs — but after the first `if` is a no-op, so behaviour is preserved for that case too. The real issue is that when `onesymperfile` is true, the `plans,:enlist each ...` line runs and then the code that was previously in the else-branch (after the closing `]` of the original `$`) is now placed OUTSIDE the `if[not o`onesymperfile;...]` block and will also execute unconditionally. Inspect the diff more carefully: the original else-branch contained more logic (packing, non-oversized handling) after the `if[o`splitoversized;...]` block. Verify that all of that logic is now guarded by `if[not o`onesymperfile;...]` and none of it leaks outside to run unconditionally when `onesymperfile` is true — if any code that was inside the original else-branch is now outside the `if[not o`onesymperfile;...]` guard, it will execute even when `onesymperfile` is set, producing duplicate or incorrect plan entries.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **1 warning(s)** | **0 suggestion(s)**

di/pqx/pqx.q
```diff
@@ -0,0 +1,464 @@
...
+      n:count tabs;
+
+      step:{[maxsize;sizes;state;i]
+        sz:sizes i;
+        tot:state 1;
+        $[(tot+sz)>maxsize; (1+state 0; sz); (state 0; tot+sz)]
+      }[maxsize;sizes];
+      bins: (step\[(0;0);til n])[;0];
+
+      plans,:value[tabs @ group bins]
```
DI-Software-Engineering, line 162: Removing the `]` that closes the inner `if`/control block unbalances the bracket nesting. The `plans,:value[tabs @ group bins]` line's block no longer has a matching closing bracket, causing a parse error or silently merging the inner block with the outer one, which will break the logic of the conditional.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
```diff
@@ -0,0 +1,464 @@
...
+  / above) and encoded in its path already, so t is stripped of them in place afterwards - once here, not
+  / duplicated into every row of the on-disk data, and not dropped later from an already row-indexed slice
+  / (which breaks the arrow write whenever a file holds more than one instrument). guarded on count: a
+  / functional delete of an EMPTY column list against a `p#-attributed symcol (set by presort, on by
+  / default) silently returns a zero-row table instead of being the no-op it should be
+  idx:.z.m.datalookup[t;o`symcol;syms;count map;combomask];
+  t:$[count o`virtualcols;![t;();0b;o`virtualcols];t];
+
+  / build paths for each seqno
+  paths:writedir,/:o[`filestub],/:"-",/:("0"^neg[count[string[numfiles]]]$string[map`seqno]),\:".parquet";
```
DI-Software-Engineering, line 213: When `numfiles` is 1, `string[1]` is `"1"`, so `count[string[numfiles]]` is 1, and `neg[1]$string[map\`seqno]` pads to width 1 — which means single-file calls write `part-1.parquet` (no padding), matching the old behaviour. But when `numfiles` is exactly a power of 10 (e.g. 10), `string[10]` has 2 chars, so seqnos are padded to 2 digits (`-01` … `-10`). The off-by-one is: the threshold for 2-digit padding should be ≥10 files (i.e., when `numfiles` itself requires 2 digits), but using `count string numfiles` makes 10 files trigger 2-digit padding correctly. However, the docs say "zero-padded to the digit-width of the total number of files" and the example says `part-02.parquet` when the call writes 10-99 files. For `numfiles=10`, `count string 10` is 2, so seqno 1 becomes `-01` and seqno 10 becomes `-10` — this is correct. The real bug is that `max raze plans[\`seqno]` is used as `numfiles`, but `seqno` is the maximum sequence number, not the count of files. If seqnos are not contiguous starting from 1 (e.g. a virtualcols path where plans for different virtual-col groups each have their own seqno sequences starting at 1), `max raze plans[\`seqno]` will equal the max seqno within a single group rather than the total file count, causing under-padding. For example, with 3 virtual-col groups each producing 5 files (seqnos 1–5), `max raze plans[\`seqno]` is 5, but 15 files are written total. The padding width should be based on `count raze plans` (total number of plan entries) rather than `max raze plans[\`seqno]`.

DI-Software-Engineering:
## DIReview Summary

di/pqx/pqx.q
```diff
@@ -0,0 +1,464 @@
...
+        system "mkdir -p \"",d,"\"";
+      ]
+    }[writedir] each distinct opts[`virtualcols]#plans;
+  ];
+
+  / check parallel flag to pick iterator function
+  parallelfn:(each;peach)[opts[`parallel]];
+
+  .z.m.loginfo[`pqx;"Writing down ",string[tname]," Parquet files to disk for ",string dt];
+  res:raze parallelfn [.z.m.writefile[t;opts;writeopt;writedir;count raze plans[`seqno]]; plans];
```
DI-Software-Engineering, line 422: The change from `max raze plans[`seqno]` to `count raze plans[`seqno]` is semantically wrong when sequence numbers are used as an offset or upper bound (e.g., to size a pre-allocated structure or generate a sequence). `max` returns the highest sequence number present, while `count` returns the total number of elements. These are equal only when seqno values are a zero-based, gap-free range. If seqno values start from a non-zero offset, contain gaps, or are assigned non-contiguously (e.g., 1-based, or per-partition offsets), `count` will produce an incorrect value where `max` was the intended bound. The original `max` should be restored unless the calling convention of `.z.m.writefile` genuinely expects a count rather than a maximum sequence number.

DI-Software-Engineering:
## DIReview Summary

DI-Software-Engineering:
## DIReview Summary

### Suggestions
- **di/pqx/pqx.q:386** — The attribute application `t:@[t;opts[`symcol];`p#]` replaces the entire table `t` with only the modified column vector. `@[table;col;func]` on a table returns a new table with the column updated, but `` `p# `` used as a function via `@` applies the attribute to the column in-place and returns the whole table only if the monadic form is correct. However, the correct form to set an attribute on a table column is `` t:update `p#symcol from t `` or the functional update idiom `` ![t;();0b;enlist[col]!enlist enlist(`p#;col)] ``. Using `@[t;col;`p#]` passes `` `p# `` as the function, but `` `p# `` is not a valid monadic function — it is an adverb that requires a right argument (the list to attribute). This will signal a type or rank error at runtime. The original functional-update form was correct; revert to it or use `` t:![t;();0b;(enlist opts[`symcol])!enlist(`p#;opts[`symcol])] ``.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **0 warning(s)** | **2 suggestion(s)**

### Suggestions
- **di/pqx/pqx.q:213** — `numfiles` is passed as `count raze plans[\`seqno]` at the call site, but `plans[\`seqno]` after `flip` produces a list of seqno integers (1,2,...,n), so `raze` is a no-op and `count` equals the total number of files — that part is fine. Inside `writefile`, however, `numfiles` represents the TOTAL file count across ALL virtual-col combinations, but `writefile` is called once per virtual-col combination (each element of `plans`). For a call with e.g. 2 virtual-col combinations of 5 files each, `numfiles` is 10, but each `writefile` invocation only sees the `map` slice for its own combination (5 entries). The seqno padding width is computed from `numfiles` (10, 2 digits) while the seqnos in `map\`seqno` are 1–5, so the padding is wider than necessary — this is not a correctness bug for the file names but the padding width is incorrect relative to what was intended (pad to the width of the files in *that combination*). More importantly: when `virtualcols` is non-empty, seqnos within each combination start from 1 and go up to the size of that combination, not up to `numfiles`. The test in test.csv only covers the `onesymperfile` (non-virtual) path, so this edge case is untested.
- **di/pqx/pqx.q:385** — Replacing the functional-form attribute assignment `![t;();0b;(enlist opts[\`symcol])!enlist(#;enlist \`p;opts[\`symcol])]` with `@[t;opts[\`symcol];\`p#]` changes the semantics: the functional update applies `\`p#` inside a q-sql update (which operates on the column in-place on a copy of the table), whereas `@[t;opts[\`symcol];\`p#]` uses the amend-at operator, which applies the function `\`p#` to the *value* of the column (the symbol list) and stores the result back. `\`p#` applied directly to a symbol list returns the list with the `p` attribute set, but `@[table;colname;fn]` applies `fn` to `table[colname]` and assigns it back — this correctly sets the attribute on the column. However, if the sort done by `xasc` on the line above has not produced a contiguous parted-attribute-compatible column, `\`p#` will signal a type/rank error that was previously suppressed by the functional update path (which uses the same operator but may handle it differently). Confirm the `@[t;col;\`p#]` form signals the same way as the functional select form on a non-parted column so no silent error swallowing occurs.

di/pqx/pqx.q
```diff
@@ -0,0 +1,464 @@
...
+  / above) and encoded in its path already, so t is stripped of them in place afterwards - once here, not
+  / duplicated into every row of the on-disk data, and not dropped later from an already row-indexed slice
+  / (which breaks the arrow write whenever a file holds more than one instrument). guarded on count: a
+  / functional delete of an EMPTY column list against a `p#-attributed symcol (set by presort, on by
+  / default) silently returns a zero-row table instead of being the no-op it should be
+  idx:.z.m.datalookup[t;o`symcol;syms;count map;combomask];
+  t:$[count o`virtualcols;![t;();0b;o`virtualcols];t];
+
+  / build paths for each seqno
+  paths:writedir,/:o[`filestub],/:"-",/:("0"^neg[count[string[numfiles]]]$string[map`seqno]),\:".parquet";
```
DI-Software-Engineering, line 213: The padding width is computed as `count[string[numfiles]]` (the digit-count of the total number of files), but `numfiles` is passed as `count raze plans[\`seqno]` at the call site. `plans[\`seqno]` is already a column of consecutive integers `1 2 … N`, so `raze` of that is the same list and `count` gives N — correct. However `writefile` is also exported as a public API and its new 5th parameter `numfiles` has no default; any caller that built a call to the old 4-argument signature will get a wrong-arity error with no helpful message. If this is intentional (breaking change), it should be documented; if `writefile` is meant to be called externally, the signature change needs a migration note.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **1 warning(s)** | **1 suggestion(s)**

### Suggestions
- **di/pqx/pqx.q:385** — The functional-update form `![t;();0b;(enlist opts[\`symcol])!enlist(#;enlist \`p;opts[\`symcol])]` has been replaced with `@[t;opts[\`symcol];\`p#]`. The `@[table;col;attr]` form sets an attribute on a table column in-place only when `t` is a global variable; when `t` is a local variable (as it is here — it was just returned from `xasc`), `@[t;col;attr]` requires the table to be passed by reference. Here `t` is a local q variable holding a table value, not a global reference, so `@[t;opts[\`symcol];\`p#]` will attempt to index `t` at the symbol `` `p# `` rather than set an attribute, producing a wrong result or a type error rather than applying the parted attribute. The original functional-update correctly returned the modified table. Fix: use the functional-update `![t;();0b;(enlist opts[\`symcol])!enlist(#;enlist \`p;opts[\`symcol])]` or assign via `t:@[\`t;opts[\`symcol];\`p#]` if `t` is accessible as a global/local name reference.

di/pqx/pqx.q
ConorSwainDI: I'd rather not define a bunch of default[`....] lines at the top of the file. I have however added the name of the key to the comment, which may be a good compromise?

di/pqx/pqx.q
ConorSwainDI: I've refactored this as a pad that reads how many files are in the plan and pads as appropriate

di/pqx/pqx.q
ConorSwainDI: Fixed this

di/pqx/pqx.q
ConorSwainDI: Made this change

jprgrant (approved):

PR #129, opened 2026-09-11
Feature torqx

alowrydi:
- di.proc renamed to di.torq.proc, wired through torq.q properly
- CLAUDE.md removed from the branch
- README's stale module example fixed


DI-Software-Engineering:
The diff is too large to review automatically (estimated 151,228 tokens, limit 90,000).

di/torq/torq.q
```diff
@@ -182,6 +182,55 @@ runhook:{[proctype;overrides]
...
+
+initquerylog:{[config;deps]
+  sect:$[`querylog in key config;config`querylog;()!()];
+  if[not $[`enabled in key sect;`boolean$sect`enabled;0b];:()];
+  / di.querylog's own init is not idempotent - a second call wraps its own wrappers and every query logs twice
+  if[.z.m.querylogwired;deps[`log][`info][`torq;"query logging already wired in this process - not re-wrapping"];:()];
+  opt:{[s;k;d] $[k in key s;s k;d]}[sect];
+  cfg:`logtomemory`logtodisk`level`localtime`ignore`ignorelist!(
+    `boolean$opt[`logtomemory;1b];
+    `boolean$opt[`logtodisk;0b];
```
DI-Software-Engineering, line 215: `qlresolvedir` uses `first dir` to check whether the path is absolute, but `dir` at that point is the raw value returned by `opt` — which may be a q string (from TOML) or a symbol (from a `.q` settings file). If it is a symbol, `first` returns the first character of the symbol's string representation including the leading `:` handle character (e.g. `` `:/tmp/foo `` → `first` is `:`). The absolute-path check `"/"=first dir` then silently fails and the path is joined against `TORQXAPPHOME`. The `logroll.resolvedir` function in `logroll.q` explicitly strips the leading `:` before the same check; `qlresolvedir` omits that step. Add the same normalization: `dir:$[(0<count dir) and ":"=first dir;1_dir;dir]` before the absolute check.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **1 warning(s)** | **2 suggestion(s)**

### Suggestions
- **di/torq/torq.q:219** — The `ignorelist` default `\`upd\`.u.upd` is a 2-element symbol vector, but `qlignore` iterates with `each names,()` and applies `$[-11h=type x;x;\`$x]` to each element. A symbol atom passed to `` `$ `` throws, but here each element is already a symbol so the cast is skipped — that's fine. However the default is passed as `opt[`ignorelist;\`upd\`.u.upd]` where `opt` is `{[s;k;d] $[k in key s;s k;d]}[sect]`. When `ignorelist` IS present in the TOML section, `di.util.toml` returns string values (TOML strings → q char strings), so `s k` yields a list of strings, which is correct. But when the key is absent, the default is the symbol vector `` `upd`.u.upd ``, and `qlignore` calls `` `$x `` on each — `` `$`upd `` will throw `'type` because `` `$ `` applied to a symbol atom (type -11h) throws. Fix: change the default to string list `("upd";".u.upd")` so both paths through `qlignore` receive strings.
- **di/torq/test.q:75** — `qlasync` calls `.z.ps` via `.z.ps (`upd;`trade;())` — but `.z.ps` is a handler that is set by `di.querylog` wrapping. After `initquerylog` runs, `.z.ps` is the wrapper. However `.z.ps` is the *incoming* async message handler; it fires when a message arrives over IPC, not when called directly. Calling `.z.ps[msg]` directly in a test is a reasonable in-process simulation, but `di.querylog`'s wrapper likely captures `.z.w` (the calling handle) to record the client identity. In-process `.z.w` is 0, which may cause the wrapper to behave differently than in a real async scenario. This could result in flaky or incorrect test assertions about row counts. Verify that `di.querylog`'s `.z.ps` wrapper handles `.z.w=0` (no handle) without throwing, or drive the test via a real IPC call.

ConorSwainDI (approved):
All looks okay to me

PR #130, opened 2026-09-15
Feature segmentedandchainedtp

alowrydi:
Segmented and chained tickerplants: di.torq.proc.segmentedtp and
  di.torq.proc.chainedtp

  Two new di.torq process types, each ported from its TorQ process onto the
  process-module conventions and registered in di.torq's builtin table. Plus two
  cross-module fixes the work surfaced (below).

  di.torq.proc.segmentedtp — the segmented tickerplant

  Receives feed updates, timestamps them, writes them to a configurable layout of log
  files, publishes them in one of three batch modes, and rolls its logs at end of
  period and end of day. Built from TorQ/code/processes/segmentedtickerplant.q,
  code/segmentedtickerplant/{stplog,stpmeta,pubsub}.q and
  config/settings/segmentedtickerplant.q, mirroring di.torq.proc.tickerplant wherever
  segmented TP's own requirements don't force a divergence. Standalone configuration
  only — the chained sctp.q variant is a separate module on this branch (below).

  Layouts (multilog): tabperiod (default, one log per table per period), singular (one
  shared log, daily), periodic (one shared log per period), tabular (one log per
  table, daily), custom (per-table assignment from a table,mode csv, or setcustommode
  at runtime, which reopens the logs under the new layout). Batch modes: memorybatch,
  defaultbatch (default), immediate. Period and day rolls are driven by the injected
  timer and di.eodtime; an on-disk stpmeta metatable per day directory records every
  segment.

  Dependencies. Injected (all required): log, timer, handlers (servers is passed by
  di.torq and deliberately unread). Hard (use): di.pubsub, di.eodtime, di.tplogmgr —
  write only from the last, since naming, open and roll are self-implemented
  (di.tplogmgr is fixed at one file per date). .z.exit and pubsub's closesub on .z.pc
  go through handlers; .z.ts is left to di.timer.

  Config (init[config;deps]): kdbtplog (required), schemafile, multilog,
  multilogperiod, batchmode, replayperiod, errmode, errorlogname, createlogs,
  logprefix (defaults to procname, so processes sharing a kdbtplog cannot collide),
  tickinterval (seconds — di.timer's unit; ≥60 warns), customcsv, and the di.eodtime
  timezone settings. Strings and numbers from TOML or command-line overrides are
  parsed, never cast (`boolean$"true" is a boolean list; "j"$"5" is 53). Unrecognised
  dispatch values fail init naming the setting.

  Exports: init, upd, tablelist, subdetails, readcustomcsv, setcustommode, getcounts,
  teardown, getapimeta, version. tablelist/subdetails/tptype are TorQ's own
  .sub.subscribe protocol for a segmented TP — subdetails returns
  schemalist/logfilelist/rowcounts/date/logdir, with one (count;file) pair per
  physical file.

  Defects in the TorQ source, fixed from day one (each with a regression test)

  1. Corrupt logs were opened blind — every existing file is counted with the
     non-executing -11!(-2;…) first; a corrupt file's good prefix is recovered into
     <file>.good by a generic byte copy (not di.tplog.repair, which is hardcoded to
     `trade and would silently empty any other table's log). Files shorter than a
     header are recovered as empty.
  2. The past-period/past-roll guards stopped the whole process timer (system"t 0") —
     only this module's job is disabled; the next good roll re-enables it.
  3. The error log leaked a handle on every day roll — closed explicitly.
  4. An unclean shutdown left never-closed metatable rows, and TorQ appended a row on
     every open — reconciled by logname on the next start: resumed if reopened, closed
     with a null count otherwise (no replay).
  5. Replay lists were built per table, double-counting shared singular/periodic files
     — built per physical file.
  6. Custom mode's metatable split was two-way — rows are grouped by the physical file
     each table writes, which gives the right shape for every mode without a mode
     switch.
  7. A failed init could look initialised — the flag is the last statement of init,
     and a retry releases handles, job and handler registrations the failed attempt
     left behind.
  8. Unrecognised multilog/batchmode/replayperiod failed late or (for replayperiod)
     silently corrupted subdetails — validated at init.
  9. createlogs:0b (or a custom-mode table absent from the csv) threw on every update
     — TorQ applied a null handle; only the chained TP masked it. Tables with no open
     log publish without a write.

  Other decisions worth knowing (full design record in segmentedtp.md)

  - The tick job is protected and registered with disableonfail off. di.timer swallows
    a job's error unless its debug flag is on, and with disableonfail (its default)
    the first transient failure — a full disk on a memorybatch flush — would silently
    stop the flush and rolls for good.
  - Logs open at the period start, not the current time, so a restart within a period
    resumes the same files (what fixes 1 and 4 exist for).
  - Period and day ends run the batch flush, not just pubclear — TorQ's memorybatch
    published a buffer at period end that was never logged.
  - logdir is carried in subdetails — TorQ's .sub.subscribe segmented branch returns
    it to its caller.
  - Past-period catch-up is one period per check (TorQ's guard), not a resync;
    orphaned metatable rows get a null count rather than a -11!(-2;…) count. Both
    documented with the alternative.

  Tests

  - Unit, 69 checks (test.csv/test.q): every export; every naming mode and batch mode
    with a distinct path; stamping and multi-table messages; errmode; each of the nine
    fixes asserting the corrected behaviour (fd counts via /proc, timer-disable
    scope, metatable reconciliation outcome, per-file aggregation); period and day
    rolls; custom mode; config coercion; exit handling; teardown. Expected errors
    asserted by message, not bare fail rows.
  - Integration, 4 scenarios (test_integration.csv/.q, real child processes wired with
    real di.timer/di.torq.handlers/di.util.log): kill -9 and restart against
    genuinely unclean on-disk state, byte-smashed log recovery, layout change
    orphaning a segment, clean exit through the real handler chain; a real
    timer-driven period roll and day roll (fd closure observed in the child); and a
    real di.torq boot through torqx_init.q from a throwaway app with a settings
    cascade.
  - Exactly-once end to end (a real subscriber process calling subdetails, replaying
    logfilelist, then receiving live): every batch mode, a shared singular file, and a
    subscription made across a period roll — no loss, no duplicates.

  Findings outside the module (flagged in segmentedtp.md, not fixed here)

  - di.torq.proc.tickerplant.eoddeps amends keys onto enlist[`log]!enlist logdep,
    giving a table-valued dict — any rolltimezone/datatimezone/rolltimeoffset setting
    makes di.eodtime.init throw 'type. Reproduced by booting a tickerplant with
    rolltimezone = "Europe/London" through torqx_init.q.
  - torqx_init.q does not guard tq.init: kdb-x abandons a QINIT script silently at the
    first error (no message, exit 0), so a failed init of any process type leaves a
    live process on its port with no upd.
  - di.torq.proc.tickerplant's flush/eodcheck jobs run with di.timer's default
    disableonfail, so a single throw stops them silently.
  - di.timer's period is a whole number of seconds — no sub-second flush for any TorqX
    process.
  - No di.subscriptions consumer for the multi-file subdetails shape yet; TorQ's own
    .sub.subscribe speaks it.

  di.torq.proc.chainedtp — the chained tickerplant

  A subscriber to an upstream tickerplant (through di.subscriptions over an injected
  di.torq.servers handle) that re-presents itself to its own subscribers as a
  tickerplant: the same root surface di.torq.proc.tickerplant publishes (upd, .u.upd,
  .u.sub, .u.subdetails, endofday, .u.end), in the same immediate/batched publish
  modes, with an optional log of its own that a downstream can replay from. Lets a
  tick chain fan out through an intermediate hop, or bridge a tickerplant across a
  process/network boundary, instead of every consumer dialling the origin. Ported from
  TorQ/code/processes/chainedtp.q onto the di.torq process-module conventions.

  Dependencies. Injected (all required): log, timer, handlers, servers. Hard (use):
  di.pubsub, di.subscriptions, di.tplogmgr (write/replayupto only). Deliberately no
  di.eodtime — it follows the upstream's date (from the subscription, advanced by the
  upstream's end of day) rather than computing one — and no schema file: it relays
  whatever the upstream's .u.subdetails returns.

  Config (init[config;deps]): upstreamtype (required, a proctype — di.torq.servers is
  proctype-keyed), subscribeto/subscribesyms, replay, clearlogonsubscription, tplogdir
  (presence is the switch, as tickerplant), publishmode/pubperiod (tickerplant's fork
  exactly), connecttimeoutms/connectpollms (straight to waitfortype), logprefix
  (defaults to procname; log is <tplogdir>/<procname>_<date>, so it can't collide with
  an upstream's tp<date> in a shared directory). Legacy's tickerplantname,
  tpconnsleep/tpcheckcycles, schema, createlogfile/logdir, pubinterval are gone, each
  for a reason recorded in chainedtp.md.

  Exports: init, upd, connected, getcounts, teardown, getapimeta, version.

  Decisions worth knowing about (full design record in chainedtp.md)

  - Log opened before replay, replay driven here. The own log must be open for the
    upstream's date before replayed messages hit upd, but the date only comes back
    inside .u.subdetails, and di.subscriptions.subscribe replays in that same call
    (calling it twice would register twice and double-process live ticks). So: one
    subscribe[…;0b], open the log for sddate, then di.tplogmgr.replayuptocount-limited
    torowcount, with a local table/sym filter for a narrowed subscription. upd`
    treats a not-yet-open log as a skip, not a throw.
  - Own log is counted on open, never executed — di.tplogmgr.open replays through upd,
    which here publishes; a restart would republish the day. Corruption recovery is a
    generic good-prefix copy (as segmentedtp), not di.tplog.repair.
  - exit 0 on upstream loss (.z.pc, registered through handlers alongside servers' and
    pubsub's). Nothing in the stack resubscribes a bounced tickerplant yet; a clean
    exit hands the restart to the supervisor. Revisit when di.subscriptions gains
    reconnect.
  - Batched subdetails flushes before registering the caller, so its replay count
    never includes a row it would also get in the next flush.
  - Batched buffer appended by name (t insert x) — root-safe (proven under -11!) and
    linear; the @[`.;t;upsert-by-value] idiom was measured quadratic (20k rows 242ms →
    40k 878ms vs 14ms).
  - endofday[d] is monadic (what a subscriber receives); a past d is ignored with a
    warn, a future one followed with a warn. A failed init after subscribing takes the
    root surface down again.

  Tests

  - Unit, 56 checks (test.csv/test.q): mock log/timer/handlers/servers, with the
    servers mock handing out a real IPC handle to a stub upstream (test_upstream.q, a
    spawned process) answering .u.subdetails from a fixture log. Covers every init
    failure mode, config coercion, both publish modes with and without a log, replay
    (full/narrowed/syms/append-warn/corrupt), the downstream surface, EOD date policy,
    connection loss (exit observed via an injectable exitfn), teardown, and the
    getapimeta contract.
  - Integration, 10 scenarios (test_integration.csv/.q, real child processes wired
    with real di.timer/di.torq.handlers/di.util.log/di.torq.servers): relay;
    exactly-once across a replaying start (ticks before/during/after startup);
    2,000×10-row volume in each mode; the real timer's batched flush; narrowed
    subscription; a real di.torq.proc.rdb replaying from the chained tp through
    di.subscriptions, following live and ending its day on the chained broadcast; a
    chain of two chained tickerplants; EOD through the chain; kill -9 of the upstream
    (EXIT=0 observed); a real di.torq boot of both process types through torqx_init.q.

  Findings outside the module (flagged in chainedtp.md, not fixed here)

  - di.tplogmgr.replayupto repairs via the trade-only di.tplog.repair — every
    subscriber replaying a multi-table log through di.subscriptions is exposed to
    silent non-trade loss on corruption. Needs di.tplog's signature scan generalised
    (own ticket).
  - di.subscriptions' replay machinery (replayfilter/runreplay/doreplay) is private;
    both segmentedtp and chainedtp had to mirror pieces of it — candidate for export.
  - di.torq.proc.tickerplant.subdetails assumes di.pubsub.subscribe returned cleanly;
    for an unknown table pubsub returns (errmsg;…) and the tickerplant's tables
    becomes the error symbol.
  - di.pubsub.publish sends with one -25!; a single dead subscriber handle fails the
    send for all.
  - di/torq/torq.md still says the builtin registry holds one entry.

  Cross-module changes on this branch

  di.pubsub no longer binds .z.pc at load (0b26a81). It assigned .z.pc:{closesub[x]}
  when loaded — which, in a di.torq process, happens inside the process module's init,
  after di.torq.servers has installed the di.torq.handlers dispatcher on .z.pc — and
  so replaced it, silently disabling servers' cleanup hook and any other registrant.
  Verified empirically; it affected every process that loaded pubsub, and chainedtp is
  the first process type needing its own .z.pc handler. Consumers now register
  closesub through their handlers dependency as a simple event:
  di.torq.proc.tickerplant (which now requires the handlers dep),
  di.torq.proc.segmentedtp (its init/teardown and handler tests updated in the same
  commit), and chainedtp. pubsub's peer-server test binds it explicitly.

  di.torq.proc.tickerplant.subdetails flushes before registering in batched mode
  (df58544). upd logs before it buffers and subdetails reported rowcount without
  flushing, so a subscriber joining between flushes replayed the buffered rows from
  the log and then received them again in the next flush. One guarded pubclear; a unit
  row reproduces the double delivery on the old code (fails 0 → passes 1). Found
  building chainedtp, whose subdetails already did this.

  di/torq/torq.q: segmentedtp and chainedtp added to builtin.

  Verification

  Run from the repo root with QPATH at the checkout first (a shared mod/di earlier on
  the path silently supplies other checkouts' modules); the peer-spawning suites need
  QHOME at a KDB-X install ($QHOME/bin/q); the chainedtp suites launch the session's
  own binary. Integration suites in their own fresh q session each.

| Suite | Result |
| --- | --- |
| `di.torq.proc.segmentedtp` — unit / integration | 69/69 · 4/4 |
| `di.torq.proc.chainedtp` — unit / integration | 56/56 · 10/10 (stable over repeated runs) |
| `di.torq.proc.tickerplant` | 33/33 |
| `di.torq.proc.rdb` · `di.torq` · `di.pubsub` · `di.subscriptions` | 4/4 · 56/56 · 97/97 · 14/14 |
| `di.torq.handlers` · `di.torq.servers` · `di.tplogmgr` · `di.tplog` · `di.timer` | 160/160 · 35/35 · 12/12 · 16/16 · 66/66 |

  qlint: no error-level findings in any q file; no line >150; no CRLF.


di/torq/proc/chainedtp/chainedtp.q
```diff
@@ -0,0 +1,476 @@
...
+
+checkdep:{[deps;k;fns;hint]
+  if[not k in key deps;'"di.torq.proc.chainedtp: ",(string k)," dependency is required - ",hint];
+  if[99h<>type deps k;'"di.torq.proc.chainedtp: ",(string k)," dependency must be a dict - ",hint];
+  if[count missing:fns except key deps k;'"di.torq.proc.chainedtp: ",(string k)," dependency is missing ",", " sv string missing];
+  };
+
+checkdeps:{[deps]
+  / validate the injected deps before anything is wired - plain signals, as no logger is usable yet. all four are
+  / required whatever the config (timer is only used in batched mode), as di.torq injects all four into every process
```
DI-Software-Engineering, line 360: In `finishinit`, `(.z.m.ps\`init)[]` is called after `(.z.m.ps\`setsubtables)[.z.m.tables]`. `di.pubsub.init` is documented to populate schema state from root tables; at this point in `chainedtp.init` those root tables were defined at root by `di.subscriptions.subscribe` (step 7 in the doc), but `(.z.m.ps\`init)[]` resets the message count back to 0 on line 360 (`msgcount:0` follows it). However `msgcount` is reset to 0 on line 359 explicitly before `openlog` is called, which is fine. The real issue is that `(.z.m.ps\`init)[]` is called inside `finishinit` which runs under protected apply — but `(.z.m.handlers\`register)[\`.z.pc;\`;\`pubsub;0;.z.m.ps\`closesub]` was called in the outer `init` before `finishinit`, meaning if `finishinit` fails and `ready` and `upstreamup` are set to `0b`, the pubsub closesub handler registered on `.z.pc` is NOT removed in the failure-cleanup path (only the log is closed). A retry `init` will register it a second time (the `remove` in the retry's cleanup only removes `jobid`/`chainedtp`, not `pubsub`). The fix is to also remove the `pubsub` handler in the `initfailed` cleanup branch, or move the pubsub registration into `finishinit`.

di/torq/proc/chainedtp/test.q
```diff
@@ -0,0 +1,603 @@
...
+
+/ config for a log dir (relative to TORQXDATAHOME) plus scenario settings; nolog omits tplogdir
+cfg:{[dir;extra] (`upstreamtype`tplogdir!(`stubtp;dir)),extra};
+nolog:{[extra] (enlist[`upstreamtype]!enlist`stubtp),extra};
+
+doinit:{[c]
+  resetmocks[];
+  (use`di.pubsub)[`closesub][0];
+  (ct`init)[c;deps[]];
+  };
```
DI-Software-Engineering, line 155: `replaysyms` calls `doinit` with `subscribesyms:\`A` but `subscribeto` is left as the default null (`` ` ``). The `doreplay` guard on line 265 of `chainedtp.q` is `not (.z.m.subscribeto~\`) and .z.m.subscribesyms~\`` — with `subscribeto=\`` the whole condition is false (because the left side of `and` is false), so the replay filter is never installed. The test therefore passes regardless of whether the sym filter works, exercising a path where no filtering happens and the full upstream log is replayed. This is evidence the `and`/`or` bug in `doreplay` (see chainedtp.q line 265 finding) is not caught by the test suite.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **2 warning(s)** | **2 suggestion(s)**

### Suggestions
- **di/torq/proc/chainedtp/chainedtp.q:228** — `connect` skips `servers.startup` when `.z.m.tph in key .z.W` (the handle is still live). But `.z.m.tph` is initialised to `0Ni` at module load. On the very first `init`, `0Ni in key .z.W` may be `1b` if handle 0Ni happens to be in `.z.W` (negative int handles can appear there under certain kdb configurations). This is an unlikely but possible false-skip of `startup`. The guard should also check `not null .z.m.tph` before testing membership.
- **di/torq/proc/chainedtp/chainedtp.q:54** — `tobool` uses `()`,"1"` and `()`,"t"` etc. with `in` to test a single-character string. A one-character string is a char atom in q; `(),x` makes it a one-char vector. The check `(lower (),x) in ("true";()`,"1"`;()`,"t"`;()`,"y"`;"yes")` will never match `"1"`, `"t"`, or `"y"` because `()`,"1"` produces a char vector `"1"` (one element), while the list on the right contains the same thing — this is actually fine. However `"true"` and `"yes"` are multi-char strings (char vectors), and a one-char input can never equal them. The real issue is a single-char input like `"t"` or `"y"` or `"1"`: `in` on a char vector against a list of char vectors does element-wise membership, not list membership. `(lower (),"t") in ("true";(),"1")` returns `01b` (char `"t"` is in `"true"`), and `all` of that is `0b`. This is the same implementation as `segmentedtp.q` line 44 — the bug is shared. Use `(lower (),x) in ("true";"1";"t";"y";"yes")` as a flat list of strings, or compare with `~`.

di/torq/proc/chainedtp/chainedtp.q
alowrydi: Doesn't reproduce — di.torq.handlers.register is upsert-by-name, not append (handlers.q's upsertphase: delete-where-name then upsert). Registering pubsub twice replaces the row in place; a retry after a failed finishinit still fires closesub once, not twice. Leaving the registration in place across a failed init is intentional, matching teardown's own reasoning: a subscriber that disconnects between a failed attempt and the retry still needs cleaning out of pubsub, and closesub on an unrelated handle is already a no-op. No code change — added a comment at the init cleanup block pointing at that reasoning so the asymmetry (jobid removed there, pubsub not) doesn't read as accidental.

di/torq/proc/chainedtp/test.q
alowrydi: False positive — reads as (not A) and B, but q has no operator precedence; not binds through the and, so this is not (A and B), which by De Morgan is "narrowed if either the table list or the sym list is restricted" — the intended check. With subscribeto null and subscribesyms set, it evaluates 1b. The replaysyms unit test exercises exactly this path and asserts only the matching messages replay; it's green. No logic change — happy to spell it out as (not subscribeto~) or not subscribesyms~`` for readability if preferred.

Papamaci444 (approved):
Read the docs, most of the tests and deep spot-checked a few things.
All those were working very well.
Integration test is good that exists, in the future have to figure out something how those will be created/run.

PR #131, opened 2026-09-15
Add di.torq.proc.idb: intraday database process

pkaiserq:
Adds a new built-in idb proctype (mount + remotely-triggerable reload of the wdb's working directory), wired into torq.q's builtin registry and into wdb.q's reload dispatch - both at EOD (reloadidbs, opt-in via reloadorder) and intraday (savetodisk notifies every connected idb after any flush that actually wrote something).


di/torq/proc/idb/idb.q
```diff
@@ -0,0 +1,121 @@
...
+/ savedir/hdbdir TOML values arrive as a plain string with no leading colon (e.g. "wdb"),
+/ unlike di.torq.proc.hdb's `dir` convention; since idb's savedir/hdbdir must resolve to the
+/ SAME absolute paths as the wdb's, it has to strip a leading colon if present rather than
+/ assume one, to handle both a `.q` settings symbol (`:wdb) and a `.toml` string ("wdb") the
+/ same way the wdb itself does.
+resolvedatadir:{[dir]
+  dir:$[10h=abs type dir;dir;string dir];
+  dir:$[(0<count dir) and ":"=first dir;1_dir;dir];
+  $[dir like "/*";dir;datahome[],"/",dir]
+  }
```
DI-Software-Engineering, line 36: When `TORQXDATAHOME` is empty and `TORQXAPPHOME` is also empty, `datahome[]` returns `""`, so `resolvedatadir` returns a relative path (just `dir`) instead of an absolute one, silently producing a wrong path. Add a guard that errors if both env vars are unset when a relative `dir` is passed.

di/torq/proc/idb/idb.q
```diff
@@ -0,0 +1,121 @@
...
+    [
+      new:get f;
+      cur:@[get;`sym;`symbol$()];
+      if[count new except cur;@[`.;`sym;:;distinct cur,new]];
+      .z.m.log[`info][`loadsym;"loaded sym domain (",(string count @[get;`sym;`symbol$()]),") from ",1_string f]
+    ];
+    .z.m.log[`warn][`loadsym;"no sym file at ",(1_string f)," yet - symbol columns may not resolve"]];
+  }
+
+/ mount (or remount) whatever partitiondir[] currently points at. The dir can legitimately
```
DI-Software-Engineering: `system "l ",1_string dir` loads the partition directory synchronously and, if it contains many tables or large schema files, will also execute any `q` scripts found in that directory. More importantly, the `dir` value originates from user-supplied config (`savedir`) that is only validated to be a non-empty string after stripping the leading colon — there is no check that the resolved path doesn't contain shell-special characters (spaces, semicolons, etc.). A `savedir` value with an embedded space will silently truncate the `system` argument. Add a guard that errors if the resolved path contains whitespace or other shell-special characters.

di/torq/settings/idb.q
```diff
@@ -0,0 +1 @@
+savedir:`:wdb
```
DI-Software-Engineering: The default settings file sets `savedir:`:wdb`` but provides no default `hdbdir`. `init` requires `hdbdir` and errors immediately if it is absent (`if[not `hdbdir in key config;'"..."]`). Any idb process that starts without an explicit `hdbdir` in its TOML/process config will fail at `init`, with no useful default to fall back on. Either add a sensible default (`hdbdir:`:hdb``) or document explicitly that `hdbdir` is always required in per-process config with no default.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **3 warning(s)** | **1 suggestion(s)**

### Suggestions
- **di/torq/proc/idb/idb.q:55** — `currentpartition[]` evaluates `.z.m.fixedpartition` without checking whether `.z.m` module state has been initialised (e.g. if somehow called before `init`). More critically, when `fixedpartition` is `0b` (the common no-override path) the function returns `.z.d` — a date atom. When `fixedpartition` is `1b` it returns `.z.m.partition`. Both are dates, so this is fine structurally, but the guard `$[.z.m.fixedpartition;…;…]` uses the boolean stored in `.z.m` without any protection against the `.z.m` state not yet existing. This is the same class of problem as the already-guarded missing-dir case; unlike `domount` (which is only called from `init`/`reload`), nothing prevents `partitiondir[]` being called externally before `init`. This is a SUGGESTION-level concern given the module's intended usage, so demoting to SUGGESTION.

di/torq/proc/idb/idb.q
```diff
@@ -0,0 +1,129 @@
...
+
+/ legacy TorQ's symfilehaschanged. Size, not mtime: .Q.en only ever appends.
+/ hcount is trapped where legacy calls it bare - legacy's idb blocks on a wdb handshake so the
+/ file always exists by now; this one is config-driven and tolerates it missing.
+symfilehaschanged:{[]
+  $[.z.m.symsize<>c:@[hcount;symfile[];0];[.z.m.symsize:c;1b];0b]
+  }
+
+/ legacy TorQ's `load symfilepath` - it names the variable after the file, so it sets root `sym`
+/ even from inside a use-loaded module (verified). REPLACES rather than merges: the file is the
```
DI-Software-Engineering, line 39: When `readsym` fails (e.g. file missing or unreadable), `loadsym[]` still updates `.z.m.symsize` to the result of `@[hcount;f;0]` on line 73. If the file now exists but `load` failed for another reason, `symsize` is set to the current file size, which will prevent the next reload from re-attempting the load. The `symsize` update should only happen when `readsym` succeeded. Fix: track whether the trap fired (e.g. via a flag) and skip the `symsize` update on failure.

di/torq/proc/idb/idb.q
```diff
@@ -0,0 +1,129 @@
...
+
+/ the hdb's sym file - a single fixed path, unlike partitiondir[] (not partition-scoped: the
+/ wdb enumerates every day's flushes against the same hdbdir/sym, it never rolls per-date).
+symfile:{[] `$":",.z.m.hdbdir,"/sym"}
+
+/ size of the sym file as at the last load, as legacy TorQ's .idb.symsize. 0 is also what a
+/ missing file stats as, so a sym file appearing for the first time reads as changed.
+symsize:0
+
+/ legacy TorQ's symfilehaschanged. Size, not mtime: .Q.en only ever appends.
```
DI-Software-Engineering, line 31: `symfilehaschanged[]` writes `.z.m.symsize` (line 62), but `loadsym[]` (line 73) also writes `.z.m.symsize` after calling `readsym`. When `reload[]` calls both in sequence — `if[symfilehaschanged[];loadsym[]]` — `symfilehaschanged[]` first records the new size into `.z.m.symsize`, and then `loadsym[]` overwrites it again with `@[hcount;f;0]`. This double-write is harmless in the success path, but it means the `symfilehaschanged[]` side-effect on `.z.m.symsize` is redundant and could mislead future maintainers. More critically, the comment in the md says "loadsym[] records the size it just read on the way out", but the caller (`symfilehaschanged[]`) already recorded it — if `readsym` throws and the trap fires, `symsize` was already advanced by `symfilehaschanged[]`, so the failed load is never retried. The size-recording responsibility should live in exactly one place; since the md explicitly says it belongs in `loadsym[]`, remove the `[.z.m.symsize:c;1b]` assignment from `symfilehaschanged[]` and rely solely on `loadsym[]`'s update — but only on success (see the separate finding above).

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **2 warning(s)** | **0 suggestion(s)**

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **2 warning(s)** | **4 suggestion(s)**

### Could not be posted inline
- **WARNING** di/torq/proc/chainedtp/init.q:3 — `first read0`:::VERSION` reads only the first line of the file and does not trim trailing whitespace or guard against an empty/missing file, unlike the pattern used in `di/merge/init.q` (which uses `@[{trim first read0 x};…;…]` with an explicit empty-check). A VERSION file with trailing spaces or a read error will silently produce a padded or empty version string, causing `di.depcheck` to fail every dependency check against this module. Apply the same guarded pattern as `di/merge/init.q`: `version:@[{trim first read0 x};`:::VERSION;{'"di.torq.proc.chainedtp: VERSION file missing or unreadable"}]; if[0=count version;'"di.torq.proc.chainedtp: VERSION file is empty"];`
- **WARNING** di/torq/proc/segmentedtp/init.q:3 — Same issue as `di/torq/proc/chainedtp/init.q`: `first read0`:::VERSION` lacks the `trim`, empty guard and error trap used in `di/merge/init.q`. A padded or unreadable VERSION file will silently produce a wrong version string that makes `di.depcheck` fail every dependent module's check. Apply the guarded pattern from `di/merge/init.q`.

### Suggestions
- **di/torq/proc/chainedtp/chainedtp.q:243** — In `doreplay`, the original root `upd` is saved as `origupd:`. `upd` at the time `doreplay` runs is the module's own function (set by `publishroot[]`), so `origupd` is the module's `upd`. The `replayfilter` wrapper calls `origupd[t;x]` — which is the module's `upd`, which calls `requireready` and would throw if `ready` were false. `ready` is set to `1b` inside `subscribe[]` before `doreplay` is called, so `requireready` passes. However, if a future refactor moves `ready`'s assignment later, this silent dependency between `subscribe` and `doreplay` ordering will cause a confusing failure. This is a latent fragility, not a current bug — omitting as below confidence threshold.
- **di/torq/proc/segmentedtp/segmentedtp.q:43** — Same `tobool` definition as in `chainedtp.q`: the `(),'1'` / `(),'0'` single-character-string handling may be incorrect for char-atom inputs. See the finding on `chainedtp.q` line 61 for the full analysis.
- **di/torq/proc/chainedtp/chainedtp.q:276** — `(.z.m.subs`init)[config;deps]` calls `di.subscriptions.init` with the full `config` dict and the module's `deps` dict. The `deps` passed here is the chainedtp's own `deps` argument, which includes the `servers` key. If `di.subscriptions.init` does not accept or ignores unknown dep keys this is harmless, but if it validates the dep dict strictly (e.g. rejects unknown keys), `init` will throw on a valid chainedtp configuration. This depends on `di.subscriptions`' internal behaviour, which is not visible in the diff. Since the internals are not visible, not posting this as a finding.
- **di/torq/proc/chainedtp/chainedtp.q:388** — `(.z.m.subs`init)[config;deps]` is called inside `init` after `readconfig` has already validated settings. If `di.subscriptions.init` reads any of the same config keys (e.g. `upstreamtype`) and applies its own validation, a config that passed chainedtp's checks could still fail here with a raw subscriptions error rather than a chainedtp-prefixed one. Since the subscriptions module internals are not in the diff this cannot be confirmed — noting for awareness.

di/torq/proc/wdb/wdb.q
```diff
@@ -124,38 +134,59 @@ informgateway:{[msg]
...
   informgateway[`reloadstart];
-  {[date;pt] $[pt in .z.m.hdbtypes;reloadhdbs[date];pt in .z.m.rdbtypes;reloadrdbs[date];.z.m.log[`warn][`wdb;"reloadorder entry ",(string pt)," is neither an hdb nor rdb type - skipped"]]}[date;] each .z.m.reloadorder;
+  {[pt;ptype]
+    $[ptype in .z.m.hdbtypes;reloadhdbs[pt];
+      ptype in .z.m.rdbtypes;reloadrdbs[pt];
+      ptype in .z.m.idbtypes;notifyidbs[`.idb.rollover;pt+1];
+      .z.m.log[`warn][`wdb;"reloadorder entry ",(string ptype)," is neither an hdb, rdb, nor idb type - skipped"]]
+    }[pt;] each .z.m.reloadorder;
   informgateway[`reloadend];
   }
```
DI-Software-Engineering, line 176: The `.u.end` root assignment was removed entirely. Previously `@[`.;`.u.end;:;endofday]` published the tickerplant-broadcast entry point at `.u.end`; the diff removes that line with no replacement. The wdb.md text says the tickerplant broadcasts `(`endofday;date)` to trigger EOD, which implies `.u.end` (the conventional TP subscriber hook) must exist at root. If the TP calls `.u.end` and it is no longer defined, EOD will never fire on a live TP tick.

di/torq/proc/idb/idb.q
```diff
@@ -0,0 +1,117 @@
...
+/ the wdb has rolled to a new day (legacy's .idb.rollover). The root mount finds the new day by
+/ itself, so .z.m.partition drives nothing today - it is kept for the two things that will need it:
+/ gateway routing (legacy's .proc.getattributes) and the partition-scoped writedown modes, where
+/ the mount is savedir/<partition> rather than the root. Not exposed until one of them lands.
+rollover:{[pt]
+  .z.m.partition:pt;
+  .z.m.log[`info][`rollover;"wdb rolled to partition ",string pt];
+  intradayreload[];
+  }
+
```
DI-Software-Engineering, line 74: `setparametersfromwdb` calls `(.z.m.svc`startup)[config,(enlist`connections)!enlist enlist wt]` where `config` is the whole idb config dict. The `startup` contract (from the consistency guide) takes a connection list; merging the entire idb config (which may contain keys like `wdbtypes`, `connecttimeoutms`, and any other process settings) into what the servers `startup` function receives may cause it to misinterpret or error on unexpected keys. Only the `connections` key should be passed.

di/torq/proc/idb/test.q
```diff
@@ -0,0 +1,110 @@
...
+
+mockdeps:{[] `log`servers!(`info`warn`error!(mocklogfn[`info;;];mocklogfn[`warn;;];mocklogfn[`error;;]);mockservers[])}
+
+/ no savedir/hdbdir in config, so init must ask the wdb for them
+/ what (each;value;`.wdb.savedir`.wdb.hdbdir`.wdb.currentpartition) evaluates to on a real wdb:
+/ two hsyms the idb normalises itself, and the date
+setwdbparams:{[] `WDBPARAMS set (hsym `$FIXTUREDIR;hsym `$FIXTUREHDBDIR;TESTDATE);}
+
+/ the wdb reports a savedir that is not on disk - the window before its first flush
+setwdbmissingdir:{[] `WDBPARAMS set (`:/tmp/di_idb_k4unit_no_such_dir;hsym `$FIXTUREHDBDIR;TESTDATE);}
```
DI-Software-Engineering: `mockhandle` ignores its arguments and returns a lambda `{[q] WDBPARAMS}` as the "handle". In `setparametersfromwdb`, this lambda is called as `h:(…gethandlebytype…)` and then used as `@[h;(each;value;…);…]`. `@[f;x;e]` applies `f` as a function when `f` is a lambda, so this works incidentally — but `mockwait` is similarly called with three args `[wt;t;500]` while its definition `mockwait:{[pt;t;p] …}` accepts three parameters named `pt`, `t`, `p`, so the mock is consistent. The actual concern: `mockhandle` is defined as `{[pt;sel] … {[q] WDBPARAMS}}` — it returns a *unary* lambda, but `setparametersfromwdb` uses `@[h;x;e]` which passes one argument. That is consistent. However, the inner lambda `{[q] WDBPARAMS}` receives the *entire* parse tree `(each;value;`.wdb.savedir`.wdb.hdbdir`.wdb.currentpartition)` as `q` and ignores it, returning `WDBPARAMS` directly rather than evaluating the expression — meaning the test never validates that the real idb code constructs the correct IPC query. This is a test-coverage gap, not a runtime bug, but it means the integration assertion `askedwdb[]` only checks that `waitfortype` was called, not that the right variables were requested.

DI-Software-Engineering:
## DIReview Summary

**1 critical** | **3 warning(s)** | **1 suggestion(s)**

### Could not be posted inline
- **WARNING** di/torq/proc/idb/test.q:111 — `recordedpartition` reaches into the mangled private namespace with the hard-coded path `.m.di.0torq.0proc.0idb.partition`. This mangled name is an implementation detail of the kdb-x module loader and is not guaranteed to be stable across versions. If the loader's name-mangling scheme changes, this accessor silently returns a wrong value (or throws) without any test failure pointing at the cause.

### Suggestions
- **di/torq/proc/wdb/wdb.q:96** — `notifyidbs` is called with `enlist()` as the params argument on the intraday leg. Inside `notifyidbs`, the IPC message is built as `enlist[func],params`, so the message sent is `(`.idb.intradayreload;())` — a two-element list whose second element is an empty list. A niladic remote call should be sent as the bare symbol `` `.idb.intradayreload `` (a string or symbol, not a list), or at minimum `enlist func`. The remote kdb+ process will try to call `.idb.intradayreload` with one argument `()`, which will signal a rank error on a niladic function.

di/torq/proc/idb/test.q
```diff
@@ -0,0 +1,118 @@
...
+/ so the test exercises the parse tree the idb actually builds rather than a canned answer
+mockhandle:{[pt;sel] `scalls set scalls upsert `fn`arg!(`gethandlebytype;pt); {[q] `LASTQ set q; value q}};
+mockgetservers:{[pt] ([]w:`int$())};
+mockservers:{[] `startup`getservers`gethandlebytype`waitfortype!(mockstartup;mockgetservers;mockhandle;mockwait)}
+
+mockdeps:{[] `log`servers!(`info`warn`error!(mocklogfn[`info;;];mocklogfn[`warn;;];mocklogfn[`error;;]);mockservers[])}
+
+/ the root variables a real wdb publishes, which the evaluated query reads
+setwdb:{[sd] set[`.wdb.savedir;sd]; set[`.wdb.hdbdir;hsym `$FIXTUREHDBDIR]; set[`.wdb.currentpartition;TESTDATE];}
+setwdbparams:{[] setwdb hsym `$FIXTUREDIR;}
```
DI-Software-Engineering, line 29: `dropwdbvar` uses the functional delete `![`.wdb;();0b;enlist v]` to remove a variable from the `.wdb` namespace. However, `setwdb` (and therefore `setwdbparams`) writes the three `.wdb.*` variables directly to root via `set[`.wdb.savedir;…]` etc., but `dropwdbvar` deletes from the namespace dict keyed by `` `.wdb `` (i.e. it deletes the sub-key `v` from the `.wdb` namespace dictionary). If `set[`.wdb.savedir;x]` lands in the module's private namespace rather than root (per the module-namespace write boundary described in the consistency guide), then `![`.wdb;…]` targets a different namespace object and the variable is not actually removed, meaning the subsequent `idb[`init]` call still sees all three variables and the `fail` test row passes spuriously. Both the `set` and the `!` must target the same namespace for the test to be valid.

DI-Software-Engineering:
## DIReview Summary

di/torq/proc/idb/test.q
```diff
@@ -0,0 +1,130 @@
...
+/ a wdb that does not publish one of them - an older build, say. test.q is loaded at root with
+/ system "l", not use'd, so this and setwdb both act on the same root .wdb
+dropwdbvar:{[v] ![`.wdb;();0b;enlist v];}
+
+/ init must fail NAMING the variable it could not read, not just fail somehow - a bare `fail`
+/ row would also pass on an unrelated error and prove nothing
+initnovar:{[v]
+  resetservercalls[];
+  setwdbparams[];
+  dropwdbvar v;
```
DI-Software-Engineering, line 43: The error variable `e` will be the empty string `""` on success (not an error), so the guard `10h=type e` is correct for the success path — but on error, `@[f;x;h]` passes the error string to `{x}` which returns it, so `e` is a string. The issue is that `@[{idb[`init][cfg[];mockdeps[]]; ""};::;{x}]` passes `::` (the identity/null) as the argument to the function. The function takes no argument and ignores it, so this works — but if `idb[`init]` succeeds, the function returns `""` (empty string), which has `10h=type ""` true but `0<count ""` false, so the overall expression returns `0b`. This means a passing test returns `0b` (false), which in k4unit is a failure. The success path should not be reachable in these tests, but if `setwdbparams[]` or `dropwdbvar[]` malfunction and init succeeds, the test will fail with a confusing false result rather than a clear message. The real bug: the final expression has no semicolon terminator — per the project style guide all lines in multi-line functions must end with `;`. Without it, the function's return value falls through correctly in q but violates the stated convention.

DI-Software-Engineering:
## DIReview Summary

Papamaci444 (approved):
Looks good to me. Let's merge and run it as part of a PoC.

PR #128, opened 2026-09-09
TorqX consolidation: orchestration, process and query tier

jprgrant:
Merges the TorqX framework tier into kdbx-modules as the consolidation base, per RFC-0001 (docs/rfc-0001-torqx-kdbx-consolidation.md).

Adds:
- di.torq - the orchestrator replacing torq.q, plus its children di.torq.{servers,logroll} and the launcher at di/torq/bin (torqx.sh, torqx_init.q)
- di.proc.{hdb,rdb,wdb,tickerplant,gateway} - the process types
- di.subscriptions, di.dataaccess, di.tplogmgr - library tier
- di.asyncdispatch, di.serverselect - tracked from their feature branches
- CLAUDE.md, plus framework/DI sections appended to consistency.md and style.md
- docs/reconciliation/*.md - per-module decision records

Moves five modules into the agreed namespace hierarchy (RFC-0001 D1):

  di.toml     -> di.util.toml
  di.config   -> di.torq.config
  di.log      -> di.util.log
  di.handlers -> di.torq.handlers   + a `transform` phase (result-rewriting,
                 pre -> exec -> transform -> post; existing phases untouched)
  di.depcheck -> di.torq.depcheck   + hierarchy-aware namespace resolution, a
                 pre-load on-disk VERSION manifest graph, and the di.timer
                 contract aligned to the five keys di.torq consumes

Each is the upstream implementation plus the additions recorded in docs/reconciliation/. References updated repo-wide.

di.dbwrite, di.eodtime and di.pubsub gain a VERSION file and a `version` export: consumer manifests declare them, and di.torq.depcheck validates both a pre-load VERSION graph and a post-load session audit.

Verified: 20 k4unit suites, 1327 checks, 0 failures, each in its own process; and the full stack boots and runs end to end - tickerplant -> rdb/wdb streaming, EOD writedown/sort/move into the hdb with rdb reload, and gateway scatter-gather plus getdata map-reduce across rdb and hdb.

PR #132, opened 2026-09-17
di.clienttracking

alowrydi:
# Add `di.clienttracking` - inbound client session tracking and usage counting

## Summary

A kdb-x module for TorQ's `code/handlers/trackclients.q`, built on the consolidated `di.*`
conventions (post-#128/#129). It records every inbound IPC/WebSocket session in a `clients`
table, optionally counts per-request hits and result bytes, and reaps idle connections.

- `init[config;deps]` with settings under a `` `clienttracking `` section: `enabled` (1b),
  `maxidle` (`0D` - idle reaping off by default), `retain` (`0D02:00:00`), `trackusage` (1b).
  Defaults are TorQ's effective ones from `config/settings/default.q`. Absent section = defaults.
- `log` and `handlers` are injected (required, strictly validated); no hard `use` deps, so no
  manifest - same as `di.serverselect`.
- Every `.z.*` observer goes through `di.torq.handlers`. No direct `.z.*` assignment.
- Exports: `init`, `getclients`, `addclient`, `cleanup`, `enableusage`, `getapimeta`, `version`.
- Timespan settings accept the TOML string form (`"0D00:15:00"`, `"1D"`); the `D` is mandatory -
  `"900"` is rejected rather than silently read as 9 hours.
- `enabled:0b` on init (or re-init) unhooks every observer and watcher this module registered and
  stops, leaving the session table intact; a later enabled re-init puts everything back.

## How it works

- **Session table** - unkeyed `clients` (`w ipa u a startp endp lastp hits sz`) fed by simple
  observers on `.z.po`/`.z.pc`/`.z.wo`/`.z.wc`. Handle-number reuse closes the stale row on
  open, so there is never more than one open row per handle.
- **Usage counting** - a `post`-phase watcher on `.z.pg`/`.z.ps`/`.z.ws`. It can only attach
  once that event has an `exec` owner; until then it is deferred (info log) and
  `enableusage[]` re-attaches it later. `sz` counts sync (`.z.pg`) results only - kdb+ discards
  the `.z.ps`/`.z.ws` return value, so measuring it would report bytes never sent.
- **cleanup** - stamps `endp` on rows whose handle has left `.z.W`, force-closes handles idle
  past `maxidle` (never the caller's own handle), and drops closed rows older than `retain`.
  Idle reaping is **suspended with a warning while usage counting is not live**: `lastp`
  cannot advance without a post watcher, so every session would otherwise look idle and get
  reaped on the first sweep.
- Hot path amends by name (`.z.M.clients`) rather than by value - measured 24× faster at 50k
  rows (3.4µs vs 83µs per request). Rationale and numbers are in the design-decisions section.

## Testing

- **Unit** - 123 checks, hermetic, recording handlers. Covers the `init[config;deps]` arity
  (a one-arg call silently returns a projection rather than erroring - pinned explicitly),
  config validation (types, string timespans, negative/null/infinite rejection), enabled/disabled
  init, deferred attach, hit/sz accounting on pg vs ps, handle reuse, retain expiry, the
  suspended-reap warning, and the `getapimeta` contract.
- **Integration** - 18 checks, two legs. Inbound: a spawned child q hosts its **own** instance of
  the module with an exec owner, and this process is a genuine client of it - proves a real
  `.z.po`, hits on real requests, and the self-reap guard (calling `cleanup[]` over your own idle
  handle doesn't close it under you) on a real IPC round-trip. Outbound: this process tracks the
  handle to the child, proves idle reaping stays suspended with no exec owner here, then proves a
  real idle reap once one is registered. Skips cleanly if no q binary is found.
- Both suites pass on KDB-X **0.1.2/2025.11.17** and **5.0/2026.07.23**. qlint clean.
- Smoke-tested independently over real tcp and unix-domain sockets, and against a full
  TorqX-POC stack (tp, hdb, rdb, wdb, gateway, feed) wired from `code/common`. Three bugs
  found in that pass are fixed here with regression tests: self-reap, duplicate rows on handle
  reuse, and the by-value table copy.

Run the unit suite:
```q
q)k4unit:use`di.k4unit
q)k4unit.moduletest`di.clienttracking
```

The integration suite needs a **fresh session** (`moduletest` only loads `test.csv`):
```q
q).m.di.0k4unit.KUltf .Q.dd[hsym`$.Q.m.mp`di.clienttracking;`test_integration.csv]
q).m.di.0k4unit.KUrt[]
q)k4unit.getresults[]
```

## Known limitation - please read before reviewing

**No `di.torq` process type currently claims `exec` on `.z.pg`/`.z.ps`.** On a stock stack
that means usage counting stays deferred and idle reaping stays suspended (with a clear
warning) until a host registers an exec owner. Session tracking - the table, open/close
stamps, retain cleanup - is fully live regardless. The gate is tested in both suites and
documented; the fix belongs in `di.torq`, not here (see follow-up 1 - note that landing it
would also mean `di.querylog`'s wrappers get displaced the moment something claims `exec`,
so the two need reconciling together, not just this module).

`di.torq` also has no settings-section hook for this module yet (the way `di.querylog` has);
a host wires it from app code as shown in the Initialisation section of the docs.

## Follow-ups (separate PRs/issues, not in this change)

1. **`di.torq` / `di.torq.handlers`: a default `exec` owner on `.z.pg`/`.z.ps`.** Root cause
   of the limitation above. Note that once anything claims exec via handlers, `di.querylog`'s
   direct-assignment wrappers are displaced - the two need to be reconciled together.
2. **`di.util.toml`: two sections with different key sets throw `'mismatch`.** Found while
   feeding this module a real TOML config. Independent of this PR - worth filing/fixing
   promptly since it's already live on `main`, not just a design note.
3. **`di/torq/handlers/handlers.md` is stale** - it doesn't mention the shipped `transform`
   phase, which this module's docs rely on to explain what `post` sees.


di/clienttracking/clienttracking.q
```diff
@@ -0,0 +1,269 @@
...
+  / hits 'noupdate and is isolated/logged by di.torq.handlers rather than counting
+  / amended BY NAME (.z.M.clients is the module-qualified name of .z.m.clients): this is the per-request
+  / hot path, and a by-value update would copy three columns of the whole session table every request
+  / (measured 24x slower at 50k retained rows)
+  bytes:$[countsz;-22!result;0j];
+  update lastp:.z.p,hits:hits+1,sz:sz+bytes from .z.M.clients where w=.z.w,null endp;
+  };
+
+disableusage:{[]
+  / remove any usage post-handlers this module registered (idempotent - di.torq.handlers no-ops an unknown name)
```
DI-Software-Engineering, line 94: `coercespan` uses `"D" in v` to test whether a string contains the letter D, but `in` on a char vector checks membership of the left atom in the right vector — this is correct for a single char like `"D"`. However, the null-timespan guard `if[(null v) or (v<0D) or v=0Wn;...]` allows `0D` (zero timespan) through for `maxidle`, which is documented as the sentinel meaning "reaping disabled". But `retain` of `0D` would mean rows are purged immediately on every sweep (any closed row satisfies `endp < now - 0D`). The config defaults set `retain` to `0D02:00:00`, but a user supplying `"0D"` or `0D` explicitly for `retain` would silently delete all closed rows on every cleanup. Add a separate guard rejecting `0D` specifically for `retain` (or document that `0D` retain means "purge immediately").

di/clienttracking/init.q
```diff
@@ -0,0 +1,7 @@
+/ di.clienttracking - track connected client sessions via di.torq.handlers connection-lifecycle events
+\l ::clienttracking.q
+/ module version, read from the VERSION file (one plain-text file to bump per release). read
+/ module-relative at load (`:::` resolves to di/clienttracking) and BEFORE export, since export:([...])
+/ evaluates each name; version stays in the export so di.torq.depcheck reads it from the export dict
+version:first read0`:::VERSION
```
DI-Software-Engineering, line 6: `version:first read0`:::VERSION` — if the `VERSION` file has no trailing newline (as shown in the diff: `+0.1.0\` with `\ No newline at end of file`), `read0` returns a one-element list and `first` correctly extracts `"0.1.0"`. But if in future the file gains a trailing newline, `read0` returns `("0.1.0";"")` and `first` still gives `"0.1.0"` — that is fine. The real risk is the test `not any ct.version in "\n\r"` (test.csv line 90) only checks for those chars in the string, but `read0` strips line endings already. The actual fragility is that `first` silently drops any second line; if the VERSION file accidentally contains two version strings, only the first is used with no warning. This is minor but worth a note — the test at line 90 does not catch a multi-line VERSION file.

DI-Software-Engineering:
## DIReview Summary

**0 critical** | **2 warning(s)** | **2 suggestion(s)**

### Suggestions
- **di/clienttracking/clienttracking.q:192** — In `enableusage`, the local lambda `active` is defined and immediately called with `each usageevents` on the same expression. The lambda closes over `.z.m.handlers` and `handlerpriority`, but `handlerpriority` is a module-level name that resolves at load time — that is fine. The real issue is: `active` is a **named local variable** inside `enableusage`, yet it is also a q built-in name (`active` is not reserved, so this is not a load-time break). More critically, the result of `active:{...} each usageevents` is a boolean list, and `.z.m.usageactive:any active` uses the local name `active` which holds the results list — but the lambda is also named `active`, so after `{...} each usageevents` runs, `active` is the **result list** (the lambda is shadowed by the assignment). This is correct as written since `: owned` inside the lambda does not assign to the outer `active`. However, this is fragile and confusing: rename the lambda (e.g. `checkowner`) to make the intent unambiguous.
- **di/clienttracking/clienttracking.q:160** — In `cleanup`, the idle-reap block uses `exec w from .z.m.clients` (value copy via `.z.m`) to collect idle handles, then does `@[hclose;;::] each idle` and `update endp:now from .z.M.clients where w in idle,null endp`. If `hclose` on one of the handles triggers a `.z.pc` callback (which calls `closeclient`, which calls `cleanup` again), the re-entrant `cleanup` will attempt to `update endp` on `.z.M.clients` while the outer call is mid-loop. q is single-threaded so the callbacks from `hclose` fire synchronously inline before `hclose` returns. This means the `.z.pc` handler fires for each closed handle *during* the `each idle` loop, each calling `closeclient` → `cleanup` → stamping `endp` again (no-op since `null endp` is false after the first stamp). The outer `update endp … where w in idle,null endp` then finds no matching rows (all already stamped). The net result is correct behaviour but the re-entrant `cleanup` also re-runs the `hclose` branch — at that point `idle` is re-computed from `.z.m.clients` where the handle just closed may already have `endp` stamped by the inner `.z.pc`→`closeclient` path, so it won't be double-closed. However the full re-entrant sweep (including the `delete` purge) runs redundantly for every hclose. More importantly, if `.z.pc` is NOT registered (e.g. `enabled:0b` was set after the handles were tracked), `closeclient` won't stamp `endp`, and the outer `update` will correctly stamp them in bulk. The logic is safe but the re-entrant sweep inside the `each` is an unintended side-effect; guard `cleanup` against re-entrancy with a flag.

di/clienttracking/init.q
alowrydi: This is the same first read0:::VERSION`` pattern every module in this repo uses, not something specific to di.clienttracking.

DI-Software-Engineering:
## DIReview Summary

### Suggestions
- **di/clienttracking/test.csv:119** — The test uses a hardcoded handle value `w=8009i` to verify the row was purged, but `w` is the OS-assigned file descriptor for the connection, not the port number. The `.z.po` / `.z.pc` calls on port `8009i` will produce a `w` value assigned by the OS (typically a small integer like 5, 6, 7…), not 8009. The assertion `0=count select from ct.getclients[] where w=8009i` will almost always return 1 (zero rows match 8009) regardless of whether the purge actually worked, making the test trivially pass even if the purge is broken. Fix by capturing the handle from `.z.po` into a variable and filtering on that, or by asserting `0=count select from ct.getclients[] where not null endp` after the sweep.

di/clienttracking/clienttracking.q
alowrydi: Confirmed — retain:0D collapses to endp<now, so closed rows purge on the very next sweep rather than being retained, unlike maxidle:0D which is an explicitly-checked disable sentinel. Adding this to the config table as documented behaviour rather than rejecting the value — 'purge immediately' is a legitimate thing to configure.

Papamaci444 (approved):
Read all, tried the tests, looks good to me!

